PR #18, opened 2017-04-21
WithI18n.getWrappedInstance

brunesto:
Ahoj, Thanks for this nice library!

This is a request for adding getWrappedInstance() to WithI18n, i.e. same as in  redux provider

### DIFF
--- a/packages/lingui-react/src/WithI18n.js
+++ b/packages/lingui-react/src/WithI18n.js
@@ -3,24 +3,26 @@ import React from 'react'
 import type { I18n } from 'lingui-i18n'
 
 type WithI18nOptions = {
-  update?: boolean
+  update?: boolean,
+  withRef?: boolean
 }
 
 type WithI18nProps = {
   i18n: I18n
 }
 
-export default ({ update = true }: WithI18nOptions = {}) => function<P, C: React$Component<*, P, *>> (WrappedComponent: Class<C>): Class<React.Component<void, $Diff<P, WithI18nProps>, void>> {
+export default ({ update = true ,withRef = false }: WithI18nOptions = {}) => function<P, C: React$Component<*, P, *>> (WrappedComponent: Class<C>): Class<React.Component<void, $Diff<P, WithI18nProps>, void>> {
   return class WithI18n extends React.Component {
     static contextTypes = {
       i18nManager: React.PropTypes.object
     }
 
-	ctrls={wrappedInstance:null}
+    wrappedInstance=null
+    
+    setWrappedInstance = withRef?(ref => this.wrappedInstance = ref):null
    
-    getWrappedInstance(){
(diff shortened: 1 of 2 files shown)


packages/lingui-react/src/WithI18n.js
```diff
@@ -41,7 +47,7 @@ export default ({ update = true }: WithI18nOptions = {}) => function<P, C: React
     render () {
       const { i18n } = this.getI18n()
       // $FlowIgnore: https://github.com/facebook/flow/issues/3241
-      return <WrappedComponent {...this.props} i18n={i18n} />
+      return <WrappedComponent ref={(x)=>this.ctrls.wrappedInstance=x} {...this.props} i18n={i18n} />
```
tricoder42: Using fat-arrow functions inside `render()` forces re-rendering or `WrappedComponent` because it creates a new function for every render.

`react-redux` (https://github.com/reactjs/react-redux/blob/master/src/components/connectAdvanced.js) binds function in constructor to avoid this, we can use class properties:

```
setWrapperComponent = ref => this.wrappedComponent = ref

render () {
   ...
   return <WrappedComponent ref={this.setWrappedComponent} {...this.props} i18n={i18n} />
}
```

packages/lingui-react/src/WithI18n.js
```diff
@@ -16,6 +16,12 @@ export default ({ update = true }: WithI18nOptions = {}) => function<P, C: React
       i18nManager: React.PropTypes.object
     }
 
+	ctrls={wrappedInstance:null}
+   
+    getWrappedInstance(){
+		return this.ctrls.wrappedInstance
```
tricoder42: Feel free to add `wrappedInstance` property directly to instance as `react-redux` does it.

packages/lingui-react/src/WithI18n.js
brunesto: good point

brunesto:
Hi, second try :)  , I added the option withRef and removed the fat arrow function from render(). I  also added 2 tests but I was not able to get one of them running ... when I put some logging inside WithI18n it seems that the argument to ref in 

      return <WrappedComponent ref={this.setWrappedInstance} ...

is null during the tests, so getWrappedInstance() will always return null. I am not familiar with jest so I leave it as is. It work when used normally outside of tests though

brunesto:
Hi tricoder42, do you think you will have time to look at it in the next days ? Thanks

tricoder42:
I've fixed the test by using stateful component (stateless components can't be referenced using refs) and I've added some validation to getWrappedInstance as `react-redux` does. I'm gonna close this one and merge #20 as soon as CI checks it. Release coming soon.

Thank you for working on this!

tricoder42:
New version of `lingui-react` (0.7.0) publish to NPM just now.

PR #28, opened 2017-06-30
Update README.md

Peping:
Fix references to WithI18n (docs said WithReact beforehand)


tricoder42:
Good catch, thank you! 👍

PR #63, opened 2017-08-17
Fix yarn/npm add lingui-cli on react tutorial

felipeko:


tricoder42:
Good catch! Thank you!

Issue #75, opened 2017-09-04
lingui doesn't work in Node.js


benbender:
Hey,

if I try to use lingui in a SSR environment and I'll get an exception, because lingui-formats directly accesses the window-object, which doesn't exist in a node-environment.

**Trace:**
>ReferenceError: window is not defined
    at Object.<anonymous> (/Users/bb/Development/xxx/src/node_modules/lingui-formats/dist/index.js:5:24)
    at Module._compile (module.js:573:30)
    at Module._extensions..js (module.js:584:10)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/bb/Development/xxxx/src/node_modules/babel-register/lib/node.js:152:7) 

Is SSR simply not supported at the moment, is it a bug or am I missing something?

tricoder42:
You're right, SSR isn't tested at the moment. However, I would consider it as a bug, because `Intl` object is present in Node.js. Interested in sending PR? This looks like a quick fix.

benbender:
Sure - You'll get the pull in a few minutes.

benbender:
Done, see #76

tricoder42:
Thank you!

tricoder42:
Fixed in latest release:
 - lingui-cli@1.2.2
 - lingui-formats@1.0.2
 - lingui-i18n@1.2.3
 - lingui-react@1.1.1

PR #76, opened 2017-09-04
Fix compatibility with Intl in Node-environment

benbender:
See #75


tricoder42:
Looks good, thank you! I'll fix flow type err later.

PR #81, opened 2017-09-11
Allow react 16-alpha as peer-dependency

benbender:


tricoder42:
Cool, thank you! Published in **lingui-react@1.2.1**

Issue #85, opened 2017-09-29
Documentation proofreading


tricoder42:
# Happy Hacktoberfest! 

**Working on your first Pull Request?** You can learn how from this *free* series [How to Contribute to an Open Source Project on GitHub](https://egghead.io/series/how-to-contribute-to-an-open-source-project-on-github)

## Problem

There's a new [documentation](https://lingui.github.io/js-lingui/). However, I'm not a native english speaker and the documentation really needs some external proofreading.

List of pages:

- **Tutorials**
  - [x] React (@BitYog)
  - [x] CLI (@BitYog)
- **Guides**
  - [x] How plurals work (@ayushgangrade)
  - [x] Dynamic loading of translations (@ayushgangrade)
- **API references**
  - [x] React (@Andrew565)
  - [x] CLI (@Andrew565)
  - [x] Configuration (@Andrew565)
- **Releases**
  - [x] Migration from 0.x to 1.x (@Andrew565)

## How-to

All documentation is in `docs` folder. It's build using Sphinx Document Generator and uses [reStructuredText](http://www.sphinx-doc.org/en/stable/rest.html).

If you want to build documentation locally, you need to have Python installed.

1. Install [`pipenv`](https://docs.pipenv.org/) tool:

```sh
pip install pipenv
```

2. Install project dependencies

```sh
pipenv install
```

After making any changes, simply run `make html` inside `docs` folder.

```sh
cd docs
make html
```

## Support

Need a help? Ping me on [gitter](https://gitter.im/lingui/js-lingui). I live in **Europe/Prague UTC+2** timezone and most of the time I answer questions on gitter instantly.

Andrew565:
Getting started on this now. :)

ayushgangrade:
@tricoder42 I'm also helping !

tricoder42:
@Andrew565 @ayushgangrade Wow, that was fast! You're awesome, thanks!

I added list of pages to issue description if you would like to split the work.

Andrew565:
Thanks @tricoder42, I'll get started with the Migration, work bottom-up.

ayushgangrade:
Cool! I'll start working on the guides! @Andrew565

ayushgangrade:
@Andrew565 I did some edits for the guides. I'll try to find fixes and additions for tutorials now. Not just grammar edits, more content too!

Andrew565:
That's it for the API docs, and the last I'm going to be working on this for now. Let me know if any tweaks are needed.

[...]

PR #87, opened 2017-09-29
Grammar fixes for Migration guide.

Andrew565:
For #85


tricoder42:
Wow this is awesome, thank you very much!

I should probably wait until October 1st to merge it, right? :)

Andrew565:
I believe that is correct, yes. :)

PR #88, opened 2017-09-29
React16

acenasir:


tricoder42:
Thank you for your interest, but branch `react16` isn't ready to merge yet.

If you're interested, take a look at #90, which describes how to add React 16 support to `jsLingui`. Feel free to send a valid PR!

PR #108, opened 2017-10-23
fix: Fix typo in react tutorial docs

kirillku:


tricoder42:
Awesome! Thank you very much!

PR #112, opened 2017-10-25
Remove .isRequired flag on linguiDefaultRender context type

NSLS:
Now with react 16 it is possible to return a string from a component and lingui supports it by accepting `null` as a value of the `defaultRender` param on `I18nProvider`.

However, if you pass `null`, you will receive a warning, saying "Failed child context type: The child context `linguiDefaultRender` is marked as required in `I18nProvider`, but its value is `null`."

This PR removes the .isRequired param, which would prevent this warning when `defaultRender` is `null`


tricoder42:
Makes sense! Thank you 👍

tricoder42:
Released in **lingui-react@1.3.4**

PR #119, opened 2017-11-20
Fix doc typos

thibautRe:
- `newProps` -> `nextProps`
- remove `data` from folder path.

Also there's another thing that I had to change to get the example working:

``` js
// Doc version
loadCatalog = async (language) => {/* ... */}

// My version
async loadCatalog(language) {/* ... */}
```

The reason for that change is that in the doc version, the arrow function doesn't have any `this` bound to it. I can also change that in the PR if you also experienced some troubles with it.

### DIFF
--- a/docs/guides/dynamic-loading-catalogs.rst
+++ b/docs/guides/dynamic-loading-catalogs.rst
@@ -80,7 +80,7 @@ our component until the message catalog is loaded. Let's add a
 
 .. code-block:: js
 
-   shouldComponentUpdate(newProps, nextState) {
+   shouldComponentUpdate(nextProps, nextState) {
       const { language } = nextProps
       const { catalogs } = nextState
 
@@ -115,7 +115,7 @@ Here we use the dynamic import syntax to load the message catalog:
    loadCatalog = async (language) => {
      const catalog = await import(
        /* webpackMode: "lazy", webpackChunkName: "i18n-[index]" */
-       `locale/data/${language}/messages.js`)
+       `locale/${language}/messages.js`)
 
      this.setState(state => ({
        catalogs: {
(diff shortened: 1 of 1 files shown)


tricoder42:
Cool! Thanks a lot!

Do you use [`transform-class-properties`](https://babeljs.io/docs/plugins/transform-class-properties/)? It's required to get arrow function work for class properties as well. I should probably mention it in docs, if it's the case.

thibautRe:
No indeed, I don't use this plugin and I didn't know about it! My personal habit is to use the `env` preset or the `stage-0` plugin out of the box without worrying too much about what's inside.

I can add a mention for this plugin at the top of the file

tricoder42:
That would be great!

I always use `env` preset, but `transform-class-properties` is so useful, that I include it as well. It's also part of [CRA preset](https://github.com/facebookincubator/create-react-app/blob/master/packages/babel-preset-react-app/index.js)

docs/guides/dynamic-loading-catalogs.rst
```diff
@@ -39,12 +39,14 @@ to ECMAScript. We need to install ``babel-plugin-syntax-dynamic-import`` and
...
-       "syntax-dynamic-import"
+       "syntax-dynamic-import",
+       "transform-class-properties"
      ],
      "env": {
        "test": {
          "plugins": [
-           "dynamic-import-node"
+           "dynamic-import-node",
+           "transform-class-properties"
```
tricoder42: The plugin isn't required here. It's already on line 43, which adds it in all environments.

thibautRe: Oh of course, my bad. I will fix!

tricoder42:
Thank you again, @thibautRe!

Updated docs are deployed and I added you to [contributors section](https://github.com/lingui/js-lingui#contributors) in README.

PR #123, opened 2017-11-22
Fix dead links

thibautRe:
The folder is named `ref`, not `refs`

Dead links are [here](https://lingui.github.io/js-lingui/tutorials/cli.html#further-reading)


tricoder42:
Wow, good catch! Thank you!

PR #153, opened 2018-01-30
Correcting documentation on the lingui-loader module

ktunkiewicz:
Documentation still mentions the old `lingui-loader` which uses <2.0.0 version of lingui-cli.
It fails to work completely if installed like `npm install --save lingui-loader` and that's very likely what many people will do when reading about "The recommended way is compile messages on-the-fly using lingui-loader".


tricoder42:
Good catch! Thank you

PR #156, opened 2018-01-31
set localTransComponentName only if importDeclarations[Trans] is defi…

lorand-arvay:
…ned to avoid accidental override with undefined

When importing the following from lingui-react:

```
import { withI18n, Trans as T } from 'lingui-react';
import type { I18n } from 'lingui-react';
```

the text wrapped in <T><\T> is no longer extracted and deleted from messages.json. This happens because `localTransComponentName` in 'packages/babel-plugin-extract-messages/src/index.js' is set after each import parse which in our case leads to it having `undefined` value.

This short fix solves the above issue.


tricoder42:
Awesome, thanks! Fixed in 2.0.1!

PR #111, opened 2017-10-25
Fix catalogs paths on Windows

tricoder42:
Closes #110

PR #160, opened 2018-02-03
fix: windows path

hiddenboox:
Hi Tomáš, btw. thanks for great project!

I had some issues when using @lingui/loader on windows and I think I was fixed it.
After some investigation I see it was already fixed in previous [PR](https://github.com/lingui/js-lingui/pull/111/files#diff-7edbf3c7bf996f1eae2df53f266cce71R11) but [here ](https://github.com/lingui/js-lingui/pull/120/files)we lost it?


tricoder42:
Good catch! I'm not sure how it happened that the fix was lost... Released in v2.0.2

Thank you! 👍

PR #168, opened 2018-02-21
Update plural documentation to use value instead of count

mlabrum:
Using `count` will throw an error from the babel plugin `Value argument is missing.`


tricoder42:
Good catch! Thank you!

PR #172, opened 2018-02-27
Typescript extractor

quentez:
This is a copy of the `babel` extractor with the addition of the TypeScript plugin to support `.ts` and `.tsx` files. I've tested it locally but wasn't able to fully build the CLI to make sure it works in every situation.

(Some tests will probably be needed before moving forward with this).


tricoder42:
This is great! Let's release it and we'll see if there any problems.

Thank you @quentez!

Issue #86, opened 2017-09-29
Add babel-macros for React


tricoder42:
# Happy Hacktoberfest!

## Problem

babel-macros defines a standard interface for libraries that want to use compile-time code transformation. Once there're macros for most common tasks (css-in-js, preval, i18n), [create-react-app](https://github.com/facebookincubator/create-react-app/issues/2730) will adopt this plugin, which makes additional extensions of CRA apps much easier.

## How-to

Issue is marked as challenging, because we need to figure out how to convert `babel-plugin-lingui-transform-*` plugins to `babel-macros`. Also, `babel-macros` are still a new concept.

1. Take a look at [`babel-plugin-lingui-transform-react`](https://github.com/lingui/js-lingui/tree/master/packages/babel-plugin-lingui-transform-react/test/fixtures) test cases to see what transformations are supported. Ignore optimization of imports, as it's not needed with macros.

2. For each i18n component (`Trans`, `Plural`, ...) write a macro which transforms the code in the same way as babel plugin. For example, `Trans` component is transformed in this way:

```js
// before
<Trans>Hello world</Trans>
// after
<Trans id="Hello World" />
```

```js
// before
<Trans>Hello {name}</Trans>
// after
<Trans id="Hello {name}" params={{ name: name }} />
```

```js
// before
<Trans>Read <a href="">the docs</a>.</Trans>
// after
<Trans id="Read <0>the docs</0>." components=[<a href="" />] />
```

All transformations are covered in [tests](https://github.com/lingui/js-lingui/tree/master/packages/babel-plugin-lingui-transform-react/test/fixtures)

3. Create example using babel-macros. There's an [example-usecase](https://github.com/lingui/js-lingui/tree/master/packages/example-usecase) which uses plugin. Feel free to copy it and simply replace plugins with macros.

## Support

Need a help? Ping me on [gitter](https://gitter.im/lingui/js-lingui). I live in **Europe/Prague UTC+2** timezone and most of the time I answer questions on gitter instantly.

kentcdodds:
Awesome issue @tricoder42!

If anyone wants to take this up but you're new to writing babel plugins/working with ASTs then [read this](https://babeljs.io/blog/2017/09/11/zero-config-with-babel-macros) and look at the resources at the bottom :)

Good luck. This looks like a lot of fun!

mlabrum:
Hey,

I've got a working solution for creating a macro form of `babel-plugin-transform-react` 
[...]

PR #184, opened 2018-03-16
Bugfix/windows test support

mlabrum:
Hey,
I'm trying to tackle https://github.com/lingui/js-lingui/issues/86 and I hit an issue with running the tests on windows... seems there are some line ending issues, so I've added a replace statement to normalize the line endings.

I also needed to run lerna bootstrap, so I've added this to the contributing file.

Thanks,

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -75,6 +75,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "mlabrum",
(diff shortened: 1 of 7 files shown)


tricoder42:
Hey @mlabrum, this is awesome!

I'm gonna merge it right away, but is `lerna:bootstrap` really necessary? I migrated the project to use yarn worspaces. Could you please confirm if the dev setup works without it?

PR #187, opened 2018-03-16
docs(configuration): remove `defaults` entry

machour:
Seems to have been deleted from the output


tricoder42:
I think this is bug fixed in #206. It's related to new `i18n._` API introduced in 2.x (see #202).

Please reopen if it's unrelated! Cheers 🎉

PR #206, opened 2018-04-17
Fixes #185 - Make extract work with new i18n._ signature

tricoder42:
Fixes #185


tricoder42:
Thank you @snegostup!

Issue #207, opened 2018-04-17
CLI: add-locale doesn't support hyphens


adamgruber:
I want to use `en-US` as my locale but the cli says it is unknown. Looking at the code, it seems like it would be a minor change to add support.

This RegExp
```js
// packages/cli/src/api/formats/utils/locales.js
export const localeRe = new RegExp(/(([a-z]+)(_([a-zA-Z]+))?)(\\|\/)?/)
```
could be changed to
```js
export const localeRe = new RegExp(/(([a-z]+)([_-]([a-zA-Z]+))?)(\\|\/)?/)
```

I can open a PR for the change but wasn't sure if there was a specific reason not to support hyphens?

tricoder42:
The only reason is that I missed this :/ Feel free to send a PR! Very appreciated!

tricoder42:
Thanks @adamgruber, fixed in **2.0.8**!

PR #208, opened 2018-04-17
Update locales.isValid method to support hyphens

adamgruber:
Support hyphens when using CLI `add-locale` command.

Addresses #207


spencermefford:
+1. Thanks @adamgruber for taking the time to do this!

tricoder42:
Thanks @adamgruber, this PR is released in **2.0.8**!

PR #209, opened 2018-04-17
Configure Lingui using a separate file

ghostd:


ghostd:
You'll maybe need to regenerate the yarn.lock

tricoder42:

Issue #171, opened 2018-02-26
Babel throws "Plugin/Preset files are not allowed to export objects, only functions."


ktunkiewicz:
lingui 2.0.4, babel 7.0.0-beta.40

```
> lingui extract --clean

Extracting messages from source files…
/code/App/node_modules/@babel/core/lib/config/config-descriptors.js:158
    throw new Error("Plugin/Preset files are not allowed to export objects, only functions.");
    ^

Error: Plugin/Preset files are not allowed to export objects, only functions.
    at createDescriptor (/code/App/node_modules/@babel/core/lib/config/config-descriptors.js:158:11)
    at /code/App/node_modules/@babel/core/lib/config/config-descriptors.js:101:12
    at Array.map (<anonymous>)
    at createDescriptors (/code/App/node_modules/@babel/core/lib/config/config-descriptors.js:100:27)
    at createPresetDescriptors (/code/App/node_modules/@babel/core/lib/config/config-descriptors.js:92:10)
    at /code/App/node_modules/@babel/core/lib/config/config-descriptors.js:64:19
    at presets (/code/App/node_modules/@babel/core/lib/config/config-descriptors.js:54:25)
    at mergeChainOpts (/code/App/node_modules/@babel/core/lib/config/config-chain.js:294:68)
    at /code/App/node_modules/@babel/core/lib/config/config-chain.js:247:7
    at buildRootChain (/code/App/node_modules/@babel/core/lib/config/config-chain.js:83:20)
```

There's a closed issue mentioning something like this: https://github.com/lingui/js-lingui/issues/163

The difference is that I actually *updated to 2.0.4 from 2.0.0 and got this error* (while the author of issue#163 had an opposite situation - he was getting this on 2.0.2 and it was "fixed" and updated to 2.0.4)

Any ideas how to fix that?

tricoder42:
What's the output of `npm list | grep lingui` command?

As a workaround, could you replace the preset with plugins and try that?

ktunkiewicz:
```
❯❯❯ npm list | grep lingui
├─┬ @lingui/babel-preset-react@2.0.4
│ ├── @lingui/babel-plugin-transform-js@2.0.4
│ └── @lingui/babel-plugin-transform-react@2.0.4
├─┬ @lingui/cli@2.0.4
│ ├─┬ @lingui/babel-plugin-extract-messages@2.0.4
│ │ └── @lingui/conf@2.0.4 deduped
[...]

PR #220, opened 2018-05-16
Switch back to Babel 6

tricoder42:
Upgrading to Babel 7 is backward-incompatible change and requires project to update to Babel 7 as well. This causes several bugs when using Babel 6, eg. #171.

Since Babel 7 is still in beta, js-lingui 2.x will use Babel 6 and new version 3.x will use Babel 7. These two releases will coexist in parallel for some time.


tricoder42:
@adamgruber Sorry for delay. Yeah, I'm having trouble to configure babel/rollup. For some reason, `transform-runtime` imports aren't wrapped in `_interopRequireDefault` which causes errors. Do you have any experience with this?

tricoder42:
@adamgruber Problem seems to be solved with a simple rollup config. I'm waiting for the tests (green locally) and then I'll proceed with release.

tricoder42:
I know... I really don't want to screw this up, because it's potentially backward incompatible release for all those using Babel 7 beta.  I would like to release 3.0-beta right after, but I don't have release process ready. I guess I'm gonna ship this pig tonight and 3.0-beta soon.

Thank you for patience!

tricoder42:
Released in **2.1.0**, thanks everyone for patience!

PR #227, opened 2018-06-16
fix: make date related tests work in different timezones

lecstor:
Date related tests may fail depending on where and when they are run. This PR fixes that for me.

```
Summary of all failing tests
 FAIL  packages/react/src/DateFormat.test.js
  ● DateFormat › should render

    expect(received).toEqual(expected)

    Expected value to equal:
      "6/17/2017"
    Received:
      "6/18/2017"

      15 |       "Render"
      16 |     )
    > 17 |     expect(node.text()).toEqual("6/17/2017")
      18 |   })
      19 |
      20 |   it("should render translation inside custom component", function() {

      at Object.<anonymous> (packages/react/src/DateFormat.test.js:17:25)

  ● DateFormat › should render translation inside custom component

    expect(received).toEqual(expected)

    Expected value to equal:
      "<p class=\"lead\">6/17/2017</p>"
    Received:
      "<p class=\"lead\">6/18/2017</p>"

      36 |       .html()
      37 |
    > 38 |     expect(html1).toEqual('<p class="lead">6/17/2017</p>')
      39 |     expect(html2).toEqual(html1)
      40 |   })
      41 | })

      at Object.<anonymous> (packages/react/src/DateFormat.test.js:38:19)
```


tricoder42:
Interesting! May I ask what timezone do you live in? I can't figure out in what timezone is `2017-06-17 14:00 UTC` considered as `2017-06-18`.

lecstor:
I'm in Australia/Brisbane (+1000) which makes it 12am on the 18th I guess.

tricoder42:
Ah, right! Obviously…

Merging, thank you for the fix!

PR #228, opened 2018-06-16
Typescript extractor

### LINKED ISSUE #151: Typescript support
Finalize plugin for extracting messages from typescript files. This need somebody with Typescript experience.

Status
=====

Base of plugin is created in [typescript-extract-messages](https://github.com/lingui/js-lingui/tree/9f60e7ff931d2cd1a86d3731b036c766b46740b0/packages/typescript-extract-messages) package (original version created by @johansigfrids). It uses the same test suite as `babel-plugin-extract-messages`. Extracting of messages works, but there're few problems:

- babel plugin also transforms file and removes `i18nMark` functions, because they are used only to mark strings for extraction. Typescript plugin so far doesn't modify original file and I don't know if Typescript works in the same way as Babel.
- babel transform plugins take `<Trans>` tags and `i18n.t` calls and transform them into low-level API calls. Since there're no transform plugins for typescript which would do the same, typescript users have to use low-level API directly.
- plugin doesn't check imports - Trans component and i18n object might be aliased
 
Once the plugin is finish, it can be easily hooked to [CLI](https://github.com/lingui/js-lingui/blob/next-2/packages/cli/src/api/extractors/typescript.js):

```jsx
import extract from "@lingui/typescript-extract-messages"

const extractor: ExtractorType = {
    ...

    extract(filename, targetDir) {
        extract(filename, { localeDir: targetDir })
    }
}
```

Goals
====

1. All tests must pass (should be the same test suit as for `babel-extract-messages` plugin or at least very similar). Remove typescript plugin from ignored modules in `scripts/jest/config.unit.js` to run the test suit.
2. Check that plugin works on [Test project](https://github.com/papoola/lingui-test-react-ts) (provided by @papoola)

huan086:
Resolves #151 by using `typescript` to output ES2016 to preserve tagged template literal, and to preserve JSX. The result is then passed to existing lingui transforms and extract.

Known issues: `typescript` currently supports new features like fragment syntax `<>children</>`, but these are only supported in Babel 7. Thus the extractor cannot parse files using the new features. I'm not sure if we should upgrade to Babel 7 (still in beta) to solve this.


huan086:
One more thing, no idea where to write unit test for this

tricoder42:
@huan086 Thank you for this PR! Great work!

Babel 7 will be used in next major version (3.x). Right now it's used in `master` branch, but I haven't merge `stabel-2.x` for a while. I'll take a look. Would be great to keep these two branches as close as possible so we can release `3.x` once Babel 7 becomes officially available.

For the unittests, I would put then in `src/api/extractors/typescript.test.js`, but atm I have no idea how to write them.

huan086:
Currently writing the Typescript definitions and submitting them to DefinitelyTyped. Facing difficulties for @lingui/react because the react types are quite hard to understand.

After the typings are done, I'll be able to write the unit tests without typescript complaining about xxx is not a module.

huan086:
Unit tests are failing 😱. Don't merge

tricoder42:
This looks solid! Would you be interested in updating [Typescript guide](https://github.com/lingui/js-lingui/blob/master/docs/guides/typescript.rst)? Is there any special setup required to make typescript work or is it enough to install `@lingul/cli` and run `lingui extract`?

huan086:
No special setup needed. The original typescript implementation already did all the groundwork 😊

tricoder42:
Awesome! 👍

tricoder42:

Issue #229, opened 2018-06-16
Predictable message catalog output when using lingui extract


queicherius:
I would love if `lingui extract` output the JSON file in a predictable way (e.g. ordered alphabetically), since we check in the resulting translation file and the shuffling of keys due to source code changing positions generates a lot of noise in the diffs.

I'm up to create a PR for this if you'd like.

nicolas-cherel:
Yes, I was bothered by that too. I'm not sure that an alphabetical order on key is the best choice though, since order would change with keys change.
I'd suggest an order on first origin:line with origins sorted alphabetically.

tricoder42:
Hey @queicherius, this is reasonable request. I'm wondering what would be the best sort.

- by message ID
- by first origin and message ID
- by first origin and line

I guess all of them have pros and cons and all of them are better than random sort. I still think sorting messages alphabetically by message ID is the best choice, because changed message ID is usually different message. What do you think, @nicolas-cherel?

queicherius:
Personally, I would prefer message ID.

When using origin the messages will move around when components get moved around or messages get re-used somewhere else (working with React, in our fairly large codebase that happens quite regularly).

When using message ID they (should) only move around when the message changed.

nicolas-cherel:
Yes sure, it's a tradeoff between the twos, so I'll just expose my rationale: 

I like the message position in code because a react refactor that causes components moving around either it should not change messages, so changeset is a not brainer (those messages just moves), or for a big refactor, changed messages are likely that have their context change, so semantically in the app they are different, and can't be treated as rewording.

If you order on message Id a rewording might move a message or not, making difficult to see in the diff that it's just a rewording.

That said, order by Id also has the convenient property that it'll be directly searchable in the `messages.json` file 😋, plus people that handcraft ids for each message might not find any value in having them sorted by position inside codebase.

tricoder42:
Thank you @queicherius and @nicolas-cherel for your inputs!

Let's sort it alphabetically for now. We might make it configurable in the future, but I like the simplicity of alphabetic sort.

@queicherius, still interested in PR?

tricoder42:
Fixed and released in [v2.2.0](https://github.com/lingui/js-lingui/blob/stable-2.x/CHANGELOG.md#220-2018-07-04). Thanks to all involved!

PR #230, opened 2018-06-20
Order extracted messages alphabetically by message ID

queicherius:
This is my try to implement #229.

- I did not work with ramda before, so I am sure there is a better way to implement `orderByMessageId`.
- Only a part of the tests ran locally for me, so I'm going to see what CI says
- I also did not test the actual output of the built command, since the build failed for me locally
- Feel free to completely discard this if there is a better way to implement it 😄

### DIFF
--- a/packages/cli/src/api/__snapshots__/extract.test.js.snap
+++ b/packages/cli/src/api/__snapshots__/extract.test.js.snap
@@ -50,3 +50,53 @@ Object {
   },
 }
 `;
+
+exports[`order should order messages alphabetically 1`] = `
+Object {
+  en: Object {
+    LabelA: Object {
+      translation: A,
+    },
+    LabelB: Object {
+      translation: B,
+    },
+    LabelC: Object {
+      translation: C,
+    },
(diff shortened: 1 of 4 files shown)


tricoder42:
> I did not work with ramda before, so I am sure there is a better way to implement orderByMessageId.

Me neither, looks good!

> I also did not test the actual output of the built command, since the build failed for me locally.

How the build failed? Would you like to share the output?

> Feel free to completely discard this if there is a better way to implement it 

No, it's awesome! Thanks, going to merge and release it soon.

queicherius:
Thanks for merging! :) Looking forward to this.

> How the build failed? Would you like to share the output?

I use Linux for windows, and sometimes builds just fail for me on that setup. I attached the output of `node ./scripts/test.js` here: [test-output.txt](https://github.com/lingui/js-lingui/files/2130660/test-output.txt)

tricoder42:

tricoder42:
@queicherius hmm, weird. This looks like something in lodash or probably lodash/windows combination :/

PR #219, opened 2018-05-15
Support locales with hyphens in Lingui Loader

spencermefford:
This adds further support for locales that have a hyphen rather than an underscore (e.g. `pt-BR`). I ran into an error when trying to do the following in version `2.0.8`:
```
catalog = await import( // eslint-disable-line function-paren-newline
      /* webpackMode: "lazy", webpackChunkName: "i18n-[index]" */
        '@lingui/loader!../locale/pt-BR/messages.json');
```
This was the error:
```
Error in @lingui/loader!../locale/pt-BR/messages.json
Module build failed: TypeError: Cannot read property 'toString' of undefined
    at createCompiledCatalog (/Users/spencermefford/dev/project/node_modules/@lingui/cli/api/compile.js:101:110)
    at Object._default (/Users/spencermefford/dev/project/node_modules/@lingui/loader/index.js:43:41)
```

PR #231, opened 2018-06-23
support locales with hyphens in cli compile

leonardodino:
my locale files are written with an hyphen, thanks to #219.

unfortunately `lingui compile` doesn't support them yet,
this fixed it on my project. 😄 

---

this was the log emitted before the changes.

```
Error: Invalid locale en-US (missing plural rules)!
Error: Invalid locale pt-BR (missing plural rules)!
Error: Invalid locale zh-CN (missing plural rules)!
```

they are all supported by `make-plural` package.


tricoder42:
Good catch, thanks!

tricoder42:

PR #267, opened 2018-08-14
add item in projects using js-lingui document

zenyr:
Hi, I'm Jinhyeok Lee working at MyMusicTaste.

Since we have already mentioned that we're using js-lingui in our [Frontend job description document](https://github.com/MyMusicTaste/recruit/blob/master/frontend.md) that had been [available to public](https://github.com/MyMusicTaste/recruit/blob/master/README_EN.md) long ago, I went ahead and added the service.

We've been using js-lingui on typescript before any TS support was available so I had to make up some workarounds to extract and transform ts/tsx files as expected. (by putting files through a set of tsc & babel plugins, basically `yarn i18n:scrap` to build js catalogs and `yarn i18n:merge` to update translations from transifex dump)

Time well spent, here we are, Thanks to your project.

commit messages may be a bit different than usual. Feel free to amend/squash/rebase them if necessary.

### DIFF
--- a/docs/misc/showroom.rst
+++ b/docs/misc/showroom.rst
@@ -4,3 +4,4 @@ Projects using jsLingui
 Feel free to `send a PR <https://github.com/lingui/js-lingui/issues/new>`_ to list your project here.
 
 - `ledgy.com <https://www.ledgy.com/>`_ (`source <https://github.com/morloy/ledgy.com>`_)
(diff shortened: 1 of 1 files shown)


tricoder42:
Hey @zenyr, that's really great to hear! Anything specific to TS support you would like to improve in jsLingui?

Issue #258, opened 2018-08-08
Feature: i18nMark should accept default value


HamidAghdaee:
I'm wondering why` i18nMark` doesn't accept a default value? why just an Id? I think it's use is similar to `defineMessages` of react-intl. 

Our use case is to create a json translations resource file from an object of key value pairs in source code.

Thanks.

tricoder42:
Sounds reasonable. I'll think about the api, because I'm also planning to add optional `description` to `<Trans>` and `i18n.t`

Just to clarify: the message and default value must be string literal. It can't be a variable, for example this won't work:

```
const msgs = {
  msg: "Default message",
  msg2: "Another default message"
}

msgs.forEach(key => i18nMark(key, msgs[key]))
```

HamidAghdaee:
Oh I see. That makes sense. So there's no way to do what you mentioned? turning something like this to translations:

```const msgs = {
  msg: "Default message",
  msg2: "Another default message"
}

msgs.forEach(key => i18nMark(key, msgs[key]))```

If so, I still think having `default value` added to i18nMark is useful, as is `description` by the way. But I don't think we would use it for our use case. We would probably just convert the translations JS object into json without using the extract command.

Thanks.

tricoder42:
Not really. `lingui extract` is parsing static AST. It would have to execute the code to make this work.

What's your use case? You have a key value pairs, where key is message ID and value is default translation? From where this object comes from?

tricoder42:
It seems like you have a "message catalog", which you would like to merge with other messages. If we could extract this object to separate file (JSON would be the best), then we might find a solution to merge it in message catalog... Still it would be very helpful to understand what are you trying to achieve. Thanks a lot!

HamidAghdaee:
@tricoder42 you are correct.

The catalog you mentioned, for us is a catalog of 'common' translations. We want to maintain this as an object in javascript which can be type checked using TS. So that a consumer can be sure a key value pair exists for the common translation they're looking for. 

We of course also want to generate the translations json from this and feed it to jslingui context. 

which brings up a related issue:

As you know in my other issue I've opened, we ALSO want to have translations that are local to packages. What's the best way to _extend_ an instance of i18n? I would like the the root i18n instance to load the common catalog and then to have `<I18nProvider>`s under it in the react component tree which extend what's above them by adding local translations to it.

tricoder42:
[...]

Issue #197, opened 2018-03-31
Feature: add metadata to translations


HriBB:
I want to add some metadata to translations, which would also be extracted during `lingui extract` process.

Use case: I want to specify the type of the translation, and later, in the CMS, present the user with a different editor based on the translation type.

```jsx
<h1 className="title">
    <Trans id="home.title" type="text">Page title</Trans>
</h1>
<p className="lead">
    <Trans id="home.lead"" type="textarea">Some long text, bla bla. Users will expect textarea or maybe even rich text editor</Trans>
</p>
```

I can write a custom extractor to do this, but I was thinking that maybe we can add support for this directly to the Lingui library. What do you think? Does it make sense to put metadata on translations?

To be honest, I haven't really thought things through :) I mean, I could easily write my message ids like `${ns}.${key}.${type}` :)

```jsx
<Trans id="home.title.text" />
or
<Trans id="home.lead.textarea" />
```

tricoder42:
Yeah, it should be possible to add metadata when using `lingui` format.

One metadata I have thought about was help text for translators:

```jsx
{/* i18n: Month */}
<Trans>May</Trans>
```

This would be extracted as:

```json
{
  "May": {
    "translation": "",
    "help": "Month",
    "origin": [...]
  }
}
```

We might use similar syntax to add other `meta tags` depending on usecase.

I'm not 100% convinced about `type` as it could be guessed from content (e.g: type = text, markdown, icu).

HriBB:
I'm not sure if I understand how you can guess the `type` of the content if you only have message `id`.

I would prefer to use props, but I can work around it by using comments and parsing content type from help string.

HriBB:
I can maybe give a PR a try. Just to see how it goes.

sedubois:
How about `instructions` or `hint`? Wordpress (which I know almost nothing about) uses a combo of structured type and instructions.

<img width="970" alt="screen shot 2018-04-03 at 16 57 10" src="https://user-images.githubusercontent.com/4217871/38253755-29de259c-3760-11e8-9dc3-aa1879158f66.png">

HriBB:
I think it would make much more sense to allow keys to be dynamic. If we add `i18n` or `instructions` or `hint`, people will use that, but sooner or later someone will need something else, for whatever reason.

But if we allow dynamic keys, people can do whatever they want (even misuse them).
```jsx
[...]

PR #274, opened 2018-08-21
RFC: js.macros

tricoder42:
Babel plugins doesn't handle all edge cases well. I would like to fix these problem as long as add other related features: namely #197 and #258.

[RFC Preview](https://github.com/lingui/js-lingui/blob/09094d52dc322f6bd304ca01657bde905044f83e/docs/rfc/001_message_definitions.rst)


tricoder42:
Closing, because RFCs are edited in master

PR #282, opened 2018-08-24
Add I18n render prop component

tricoder42:


tricoder42:
Released in [v2.5.0](https://github.com/lingui/js-lingui/releases/tag/v2.5.0)

PR #287, opened 2018-08-26
feat: Add extractBabelOptions to configuration

FredyC:
Fixes use case where .babelrc is not used (typically for Typescript) configuration and the source code uses some extra features with Babel plugins. Current `--babelOptions` for CLI has never worked, so it's deprecated now and `extractBabelOptions` in the config file can be used instead.

@tricoder42 Can you please compile and validate docs? I tried the guide, but it's some Unix stuff and I am unable to run that on Windows. Tried in [WSL](https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux) but failed.


JSteunou:
Very sweet!! So needed :+1:

HamidAghdaee:
When will this be published to the `lingui/cli` package? The latest (2.5.0) doesn't seem to have it.

FredyC:
@tricoder42 I've discovered a slight bug in this regarding validation of the config. Can you please have a look at it? I am not sure how to set up validation when there is an actual value. The `jest-validate` module seems undocumented regarding that.

```
  "extractBabelOptions": {
    "plugins": [
      "@babel/plugin-syntax-dynamic-import"
    ]
  }
```

This gives warning: `Unknown option "extractBabelOptions" with value {"plugins": ["@babel/plugin-syntax-dynamic-import"]} was found.`

FredyC:
Yes, of course, I double checked. It's weird validation mechanism for these cases it seems.

PR #294, opened 2018-08-27
feat: Refactor PO format with pofile module

### LINKED ISSUE #288: Keep PO headers unchanged for existing catalogs
**Describe the bug**
When I load extracted PO catalogs into some sort of editor and do some changes, it does modify headers in its own way. Later when I rerun `extract` again, headers are again modified, mostly lower casing, but nothing is lost.

This is mostly an issue about file versioning as when headers will be changing between commits, it means a bunch of unrelated changes all the time.

**To Reproduce**
You can use for example known POEdit to see this happening or simply edit PO file manually to change the casing of some header.

**Expected behavior**
Headers should be kept completely unchanged for existing catalogs.

**Additional context**
This is upstream issue https://github.com/smhg/gettext-parser/issues/12 which does not seem to be getting fixed any time soon. So it's either about forking it or copying out relevant parts.

FredyC:
Most of the snapshots are still in place and passing, there are just some small adjustments which shouldn't have an impact on anything. I've added two more tests for edge cases causing input PO file to be fixed of some invalid state.

Fixes #288


packages/cli/src/api/formats/po.js
```diff
@@ -1,104 +1,93 @@
...
-function parseOrigin(origins) {
-  if (!origins) return
+const getMessageKey = R.prop("msgid")
+const getTranslations = R.prop("msgstr")
+const getExtractedComments = R.prop("extractedComments")
+const getTranslatorComments = R.prop("comments")
+const getOrigins = R.prop("references")
+const isObsolete = R.either(R.path(["flags", "obsolete"]), R.prop("obsolete"))
+
+const deserialize = R.map(item => ({
```
FredyC: I am wondering if there is some better way with Ramda so the `item` does not need to be passed in every line. Closest I could find is `R.evolve`, but that cannot do a transformation of keys, just values.

tricoder42: I think you could use [`R.applySpec`](https://ramdajs.com/docs/#applySpec) instead of `R.evolve`.

Kudos for thinking functionally! 👍

FredyC: Alright, I missed that one. It actually works, the only problem is with comments which I am not sure how to write to extract both types of comments and concat it together. I am not sure how to tell it to create a function that accepts `item` and passes it only to those two functions within `concat`.

```
  R.applySpec({
    comments: R.concat(
      getTranslatorComments,
      R.compose(
        R.tail,
        getExtractedComments
      )
    ),
```

Ramda is something I want to tackle eventually, but it's somewhat hard :)

tricoder42:
```js
R.compose(
  R.apply(R.concat),
  R.juxt([
    getTranslatorComments,
    R.compose(R.tail, getExtractedComments)
  ])
)
```

See [repl](https://ramdajs.com/repl/?v=0.25.0#?function%20getTranslatorComments%28item%29%20%7B%0A%20%20return%20item.comments%0A%7D%0A%0Afunction%20getExtractedComments%28item%29%20%7B%0A%20%20return%20item.description%0A%7D%0A%0Avar%20deserialize%20%3D%20R.compose%28%0A%20%20R.apply%28R.concat%29%2C%0A%20%20R.juxt%28%5B%0A%20%20%20%20getTranslatorComments%2C%0A%20%20%20%20R.compose%28R.tail%2C%20getExtractedComments%29%0A%20%20%5D%29%0A%29%0A%0Adeserialize%28%7B%0A%20%20comments%3A%20%27comments%27%2C%0A%20%20description%3A%20%27description%27%0A%7D%29%0A)

Although I'm not sure if like it over:

```js
item => R.concat(
    getTranslatorComments(item),
    R.tail(getExtractedComments(item))
)
```

JavaScript is missing functional operators to make this readable...

FredyC: Ah lol, I am so much thinking in Ramda terms that I did not realize I can make such simple function there :D Yea, in this case, it will be definitely more readable. Thanks!

FredyC: Alright, it's more functional now, but I am not that much convinced it's prettier :) Could be interesting to measure if it's actually faster to create such a big curried function opposed to creating it on fly.

packages/cli/src/api/formats/po.js
```diff
@@ -1,104 +1,113 @@
 // @flow
 import fs from "fs"
 import * as R from "ramda"
+import { format as formatDate } from "date-fns"
```
tricoder42: Thank you for this 🙏

packages/cli/src/api/formats/po.js
```diff
@@ -1,104 +1,113 @@
...
   read(filename) {
     const raw = fs.readFileSync(filename).toString()
+    return this.parse(raw)
+  },
 
-    const pofile = gettextParser.po.parse(raw)
-    const translations = pofile.translations[""]
-    delete translations[""] // remove headers
-    return deserialize(translations)
+  parse(raw) {
```
tricoder42: What's the reason for having separate `this.parse` method, which is used just once?

packages/cli/src/api/formats/po.test.js
```diff
@@ -71,7 +89,59 @@ describe("pofile format", function() {
...
+      }
+    })
+
+    const filename = path.join("locale", "en", "messages.po")
+    const actual = format.read(filename)
+    mockFs.restore()
+    expect(actual).toMatchSnapshot()
+  })
+
+  it("should throw away additional msgstr if present", function() {
```
tricoder42: We could use `mockConsole` here, at least for hiding that error test output.

https://github.com/lingui/js-lingui/blob/072afd8d3664dce514c0628f7e77eef207682ce6/packages/cli/src/mocks.js#L12

packages/cli/src/api/formats/po.js
FredyC: I am trying with Loco to automate the translation process by uploading/downloading translation files. That means I need to be able to parse received PO file to catalog and I don't want to save to file system because of that. Then I can call catalog.merge and be happy hopefully :)

packages/cli/src/api/formats/po.test.js
FredyC: Yea, that's probably good idea, I was running only tests for PO and haven't really considered it might be annoying when running everything :) Will fix that tomorrow.

packages/cli/src/api/formats/po.js
FredyC: And I was worried you would object against including whole new package because of a single format call :) I am not sure if it will find some other uses, there isn't probably much of date time operations needed.

packages/cli/src/api/formats/po.js
tricoder42: Fair enough!

I used to have `sync` command to upload/download messages from 3rd party services, but I never released it. I guess everything could be handled in `extract` command. Maybe we could provide hooks, to intercept extracting/collecting messages... Please keep me posted what you manage to hack together :)

packages/cli/src/api/formats/po.js
tricoder42: It's a backend package, I wouldn't be worried about extra dependency. It's definitely better than format date manually...

packages/cli/src/api/formats/po.test.js
tricoder42: There're few hanging console warnings in output, but I'm trying to limit it as much as possible. As a bonus, you can also test the error message ;) But as I said, that's just a nitpick. Hiding console output is more important.

FredyC:
I've tried this in my project and run into some issue. I will look into those later today, so don't merge yet if you have itchy finger :)

tricoder42:
Whenever you're ready 👍

FredyC:
@tricoder42 Alright, it's ready to be merged. I have fixed console and also figured out that warning with `extractBabelOptions`. It shouldn't be in this PR, but it's such a tiny change, that it feels awkward doing extra PR :)

packages/cli/src/api/formats/po.test.js
```diff
@@ -71,7 +90,61 @@ describe("pofile format", function() {
...
+
+    mockFs({
+      locale: {
+        en: {
+          "messages.po": po.toString()
+        }
+      }
+    })
+
+    const restoreConsole = mockConsole()
```
tricoder42: How is this module better than [`mockConsole`](https://github.com/lingui/js-lingui/blob/072afd8d3664dce514c0628f7e77eef207682ce6/packages/cli/src/mocks.js#L12) in `src/mocks.js`? I don't mind extra dependency, it's dev code after all, but I would rather stick to one way of mocking console, so it's consistent across the repo.

I'm not sure how it handles exceptions before `restoreConsole` is called. What happens if format.read raises an error, does it restore console somehow automatically?

packages/conf/src/index.js
```diff
@@ -34,8 +34,8 @@ export const defaultConfig = {
   format: "lingui",
   rootDir: ".",
   extractBabelOptions: {
-    plugins: [],
-    presets: []
+    plugins: ["plugin"],
```
tricoder42: Does it work actually? I thought it's a default config. Does babel actually receive a config to use plugin called `plugin` and preset called `preset`?

FredyC: It's `exampleConfig`, not default one :) Yes, I've verified it has accepted different value than `plugin`. I guess it just needs to find out what type fits in there. It's an interesting idea, but it won't work well in a moment someone would want to pass some other properties that are part of Babel configuration. It's definitely nonsense to write everything here.

tricoder42: omg, I see it: `exampleConfig: defaultConfig,`. Who wrote this?! 😆 

Cool, make sense!

packages/cli/src/api/formats/po.test.js
FredyC: Oh, I haven't noticed there is already that function in place, sorry about that. I tried throwing an error in `format.read` and it works just fine. I suppose that Jest is not using console, but writing to `stdout/err` directly so it doesn't matter if you mock it. Either way, I've used your approach, I kinda like it more since you don't need to call restore manually :)

tricoder42:
Cool, thanks a lot! Great job 👍

PR #303, opened 2018-08-30
Transform React components without explicit imports

### LINKED ISSUE #284: Allow components to be global
Since we'll be using `Trans` in nearly every file in our app, it'd be nice to not need to import it in every file. We currently do this with vars like `React` and `_` because they're so ubiquitous.

However, Lingui's babel transformers don't try to convert `Trans` if it isn't explicitly imported in a file, so we see the error:

`@lingui/babel-preset-react is probably missing in babel config, but you are using <Trans> component in a way which requires it. Either don't use children in <Trans> component or configure babel to load @lingui/babel-preset-react preset. See tutorial for more info: https://l.lingui.io/tutorial-i18n-react`

I'm not sure exactly how React handles it, but if I add `React` to the global object, babel can still transform JSX, even though I don't `import React` at the top of a JSX file. If I don't hang `React` off the global object, I'll get some expected errors.

It'd be ideal if Lingui's babel transformer could implement the same logic, so that if its `Trans` is attached to the global object, and a var `Trans` isn't defined in a file, it will be treated as Lingui's `Trans` and transformed.

Right now, it looks like it's hardcoded to ignore any file that doesn't explicitly import it:

https://github.com/lingui/js-lingui/blob/222e62e071c424809ce6963af55fe1824f2344d9/packages/babel-plugin-transform-react/src/index.js#L331-L336

Simply importing the necessary functions in all files. This is what I'm doing currently.

**Additional context**
Here's my preferred setup:

```js
// .babelrc
{ "presets": ["env", "react", "@lingui/babel-preset-react", "stage-0"] }
```

```js
// test-helper.js
import React from 'react'
import { Trans } from '@lingui/react'
global.React = React
global.Trans = Trans
```

```js
// webpack-config.js
export default {
  plugins: new webpack.ProvidePlugin({
    React: 'react',
    Trans: ['@lingui/react', 'Trans']
  })
}
```

With the above setup, this component will throw errors:
```jsx
// lingui-broken.jsx
export default () => <Trans>Hello, world!</Trans>
```
It compiles to:
```js
exports.default = function () {
  return React.createElement(
    Trans,   // <--- problematic :sob:
    null,
    "Hello, world!"
  );
};
```

Whereas, this component throws no errors and works correctly:
```jsx
// lingui-works.jsx
import { Trans } from '@lingui/react'
export default () => <Trans>Hello, world!</Trans>
```

It compiles to:
```js
var _react = require('@lingui/react');

exports.default = function () {
  return React.createElement(_react.Trans, {   // <--- beautiful :raised_hands:
    id: 'Hello, world!'
  });
};
```

_____

It's not the end of the world to import this everywhere, but it's a nice-to-have for me. So please weigh the pros and cons with that in mind. 🙂

Thank you for your consideration and, most of all, your phenomenal library. I've converted our app from react-intl to Lingui and it's a **world** of difference.

🙇🙇🙇

tricoder42:
Closes #284


FredyC:
So does this actually mean I can have my own wrapped Trans component without losing the ability to extract or is that different use case? I mean when I set `Trans` in the config to be _global_, would it mind that it got imported from somewhere else?

tricoder42:
Basically yes. By configuring [plugin](https://github.com/lingui/js-lingui/blob/master/packages/babel-plugin-transform-react/README.md) you can bypass import check. However, it also means that all `Trans` components in all files are transformed, no matter if they come from lingui, you package or something else. It needs to be used with caution and you need to explicitly set which components you would like to have transformed without explicit imports.

PR #308, opened 2018-08-31
fix #271 - give feedback for unrecognized commands

### LINKED ISSUE #271: CLI is silent when passed an unrecognized option
**Describe the bug**
The `lingui` CLI is silent when it is passed an unrecognized option.

**To Reproduce**
`$ lingui foo`

**Expected behavior**
The CLI should display an error.

`lingui --version`: 2.4.1

dephiros:
fix #271

### DIFF
--- a/packages/cli/src/lingui.js
+++ b/packages/cli/src/lingui.js
@@ -1,5 +1,7 @@
 #!/usr/bin/env node
 
+import { fuzzValidateCommand } from "./api/utils"
+
 const program = require("commander")
 
 let version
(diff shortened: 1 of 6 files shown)


packages/cli/src/lingui.js
```diff
@@ -19,3 +21,5 @@ program
   .command("extract [files...]", "Extracts messages from source files")
   .command("compile", "Compile message catalogs")
   .parse(process.argv)
+
+console.log(fuzzValidateCommand(program.commands, program.args))
```
FredyC: Wouldn't it print an empty line for valid cases? Might be better to check for the result before printing it.

dephiros: Thank you. I will add that check

dephiros:
Realized that I forgot to push up the test file. Will push it up tonight

tricoder42:
@dephiros Yeah, don't worry about it. I honestly don't know why the test fails, but I rerun the test suit and everything is fine.

I'm gonna merge it in the morning. Great job, thank you very much! ⚡️ ⚡️ ⚡️

PR #318, opened 2018-09-09
RFC-001 - Macros

### LINKED ISSUE #197: Feature: add metadata to translations

### LINKED ISSUE #258: Feature: i18nMark should accept default value

### LINKED ISSUE #298: Support of context
Coming from [this discussion](https://github.com/messageformat/messageformat/issues/207) and from my experience with gettext

Some times a same message can be translated differently depending on context, so catalog should contains as many as entries as needed

A way to handle hierarchical structure of message id or abstraction around context option / attribute

I tried `id` with prefixed message but it could lead to translation error.

tricoder42:
Implementation of [RFC-001](https://lingui.js.org/rfc/001_macros_message_descriptors.html).

Resolves #298 
Resolves #258
Resolves #197 

Work in progress:

- [x] fix extracting messages from JSX macros
- [x] update webpack/react example
- [x] documentation: API reference for macros
- [x] documentation: React tutorial

In next batch:

- [ ] core: `i18n(...)` API as a shortcut for `i18n._(...)`
- [ ] documentation: JavaScript tutorial
- [ ] macros: allow used them as standalone plugins with implicit imports
- [ ] docs: Migration guide from plugins to macros (plugins still supported until v3)
- [ ] plugins/presets: deprecate packages


examples/webpack-react-macros/src/Usecases/ElementAttributes.js
```diff
@@ -1,38 +1,38 @@
...
-          X
-        </button>
-      </div>
+      <I18n>
+        {({ i18n }) => (
+          <div>
+            <a
+              className="expression"
+              href="/article"
+              title={i18n._(t`Full content of ${articleName}`)}
```
FredyC: Is it really an improvement considering how it was looking before? :) Was the previous approach working? I am not sure, haven't used that. Also, I think that in gitter we were talking about a variant like `i18n(t`Full content of ${articleName}`)` which would much nicer imo.

tricoder42: You mean `i18n.t` over `i18n._(t`? It's necessary to move to macros... I don't think there's a way to have a object property to be a macro.

Other way would be convert it to `i18n._` directly, but for that we would need `i18n` in scope, which wouldn't be used otherwise and linter would complain.

This approach is very flexible, but the only drawback is how it looks :/ But you really get used to it. I wrote `vanilla-js` example and it was so great.

`i18n(t` is planned, I just have to figure out how to create callable object. I already started working on it but then figure out this is not the core of the problem and can be done later. Both `i18n._(..)` and `i18n(...)` will be supported anyway until v3.

tricoder42: And yeah, the previous approach had several problems, especially when plugins were misconfigured. It doesn't allow you to use `this.props.i18n.t`, because only `i18n.t` worked. You couldn't use it with `i18n.use` method, which switches language locally (`i18n.use('cs')._(t...)`). And also for *lazy messages* you had to use different API, `i18nMark`. Now there's the same API for all messages, lazy/immediate translations, with/without IDs.

As I said, I like everything about it except how it looks... but I believe it won't be such a problem once we have `i18n(t...)` API.

tricoder42:
@FredyC I've added checklist to PR description, so it's clear what's missing. I would like to release an early version asap, so I can play with it and get some feedback.

examples/webpack-react-macros/src/Usecases/ElementAttributes.js
FredyC: Well, a "callable object" is just a function with additional properties/methods attached directly to it. There is nothing wrong about that in my opinion and it will make it much easier to use and look at :)

FredyC:
@tricoder42 Right ... I am a bit preoccupied for the next couple of days, but I will try to have a look occasionally.

examples/webpack-react-macros/src/Usecases/ElementAttributes.js
tricoder42: Cool! I believe there's nothing wrong with it as well, but it felt a bit hackish :) I'm halfway through with implementation, which I saved in different branch. I'll finalize it soon.

tricoder42:
@FredyC Making `i18n` callable is a bit tricky. I've found few references on SO:

https://stackoverflow.com/questions/12656079/constructor-for-callable-object-in-javascript
https://stackoverflow.com/questions/19335983/can-you-make-an-object-callable

After fiddling a bit, I come up with this solution:

```js
import { setupI18n as legacySetupI18n, I18n as legacyI18n } from "@lingui/core"

export function setupI18n(params = {}) {
  const i18n = legacySetupI18n(params)

  function _() {
    return i18n._.apply(i18n, arguments)
  }

  // this is unsupported by flow
  _.__proto__ = i18n

  return _
}
```

The only problem is that the line `_.__proto__` is unsupported by flow, so I can't add type annotations to this file, otherwise it would break type checking.

I would appreciate any help here :)

FredyC:
I don't think you need to care about prototype at all. Instead of having `class I18n` you would have something like this. I haven't tested it, but I think it won't be far from the truth. Regarding Flow, [this should help](https://flow.org/en/docs/types/functions/#toc-callable-objects).

```js
function setupI18n(params = {}) {
  let _language: string
  let _locales: ?Locales
  // all other private properties

  // this is instead of a current `constructor`
  function i18n() {
    if (process.env.NODE_ENV !== "production") {
      this.t = t
      this.select = select
      this.plural = plural(this)
      this.selectOrdinal = selectOrdinal(this)
    }
  }

  Object.assign(i18n, {
    get availableLanguages(): Array<string> {
      return Object.keys(this._catalogs)
    }

    get language(): string {
      return this._language
    }

    get locales(): ?Locales {
      return this._locales
    }
    // etc...
  })

  return i18n
}
```

tricoder42:
Yeah, but you're not creating a callable object. What you have is a pattern to create an object with some method on contructor (you should assign properties to i18n.prototype, btw).

What we need is a function and somehow assign a i18n object to function's prototype. That way the function is callable, but "inherits" properties/methods from i18n object.

I've just moved __proto__ assignment to external file to make Flow happy and added flow types in recent commit: https://github.com/lingui/js-lingui/pull/318/commits/b8691d360a1cc2998738bbcfcd27745c4a00184d. It works. I tried few variation, but this is the only one which works. Not sure if there's a better way, but suggestions welcome :)

tricoder42:
Let's ship the pig and see how it work 🤞

FredyC:
> Yeah, but you're not creating a callable object. What you have is a pattern to create an object with some method on contructor (you should assign properties to i18n.prototype, btw).
> 
> What we need is a function and somehow assign a i18n object to function's prototype. That way the function is callable, but "inherits" properties/methods from i18n object.

I don't think you are right about this. Even if you look at the Flow doc I've linked, they are creating *callable object* like this. It's the exact same approach I've drafted above. 

```js
function add(x, y) {
  return x + y; 
}
add.bar = "hello world";
```

tricoder42:
I have to review it again. These things just give me a headache.

And common, it's not that bad :) `i18n(t'Hello')` vs `i18n._(t'Hello')` is just a small difference :) To be honest, I would love to remove the parenthesis completely, but I don't know how to combine macro and i18n object... This is still suboptimal, the previous version looked better (`i18n.t'Hello'`), it was just buggy :/

FredyC:
I guess this is a limitation of macros, they are nice and all, but with reliability on actual imports, it makes it much verbose. But would it be too bad to have a separate **optional** babel plugin that could convert current `i18n.t'Hello'` to an actual `i18n._(t'Hello'`

PR #321, opened 2018-09-11
Fix react-intl examples

edorivai:


tricoder42:
Good catch! Thank you 👍

PR #326, opened 2018-09-12
RFC-002 - Catalogs

### LINKED ISSUE #257: Feature: lingui extract should collect messages from multiple sources
Thank you so much for this wonderful library. 

One issue we're facing at the moment::

Our application is large enough that we want to extract translations from various src folders in our application into different locale folders. Currently there is no way to pass certain config options to `lingui extract` command such as `srcPathDirs` and `<rootDir>/locale`

It would be nice to either be able to pass these in as parameters or to force the command to read these from a custom config file.

Thanks again.

tricoder42:
Implementation of [RFC-002](https://lingui.js.org/rfc/002_catalogs.html)

Resolves #257 

Work in progress:

- [x] Deprecate configuration, add `catalogs`
- [x] Add `locales` to configuration
- [x] Extract messages -> create message files (POT)
- [x] Create translation files (PO)
- [x] Support exlude pattern
- [x] Documentation: update `@lingui/conf` reference
- [ ] Documentation: add guide about catalogs, deprecate guide about ignoring files 

In next batch:

- [ ] Load messages from script


yaitskov:
and?

zagoa:
Any news on this ?

tricoder42:
Hey @abhashmi, @ahmed-belal, @alexarnautu, @Photonios and @cipix2000! 👋 

I saw your great work in [lingui-multi](https://github.com/SectorLabs/lingui-multi). New configuration for catalogs seems to solve exactly the same problem. I wonder if you could give me your feedback on [this RFC](https://lingui.js.org/rfc/002_catalogs.html)? I'm planning to merge it into pre-release this week.

The PR is almost ready, I'm just thinking about the final format of configuration:

**mapping {[path: string]: CatalogConfig}**

```js
{
   "lingui": {
      "catalogs": {
         "./locale/{locale}/shared": [
            ".",
            "!./node_modules/"
         ],
         "./locale/{locale}/graphql": "run!scripts/fetch_graphql_enums.js",
         "./components/{name}/locale/{locale}": "./components/{name}/",
      }
   }
}
```

**list of catalog configs**

```
{
   "lingui": [
      {
         "path": "./locale/{locale}/shared",
         "include": ["."],
         "exclude": ["node_modules"]
      },
      {
         "path": "./locale/{locale}/graphql",
         "include": [{
            "type": "script",
            "path": "scripts/fetch_graphql_enums.js"
         }],
      },
      {
         "path": "./components/{name}/locale/{locale},
         "include": [./components/{name}/"],
         "exclude": ["node_modules"]
      }
   }
}
```

I think I tried both, because most of codebase uses the first format, while one part uses the second format. It's probably unimportant, but the first one seems to be more readable, while the second one is more explicit. 

What do you think?

PR #339, opened 2018-09-24
Fix Markdown link

edorivai:
I'm not sure how these doc links are supposed to work, but I'm guessing the bracket types should be flipped :sweat_smile:.


tricoder42:
Hey @edorivai, it's [reStructuredText](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#hyperlinks), but thank you for spotting the bug! It's gonna be live soon.

Cheers 👍

PR #340, opened 2018-09-25
Update react-intl.rst

edorivai:
The code block was invalid:

- forgot to close `<NumberFormat>`
- `one` and `other` params were invalid JSX, since they weren't wrapped in a tag. I wrapped them in a fragment now.

### DIFF
--- a/docs/misc/react-intl.rst
+++ b/docs/misc/react-intl.rst
@@ -216,6 +216,7 @@ components in both variables and the message itself.
 
       <Plural
          id="welcome"
+         value={number}
          one={<>Hello <b>{name}</b>, you have <NumberFormat value={number} /> message.</>}
          other={<>Hello <b>{name}</b>, you have <NumberFormat value={number} /> messages.</>}
       />
--- a/docs/misc/react-intl.rst
+++ b/docs/misc/react-intl.rst
@@ -216,8 +216,8 @@ components in both variables and the message itself.
 
       <Plural
          id="welcome"
-         one={Hello <b>{name}</b>, you have <NumberFormat value={number} message.}
-         other={Hello <b>{name}</b>, you have <NumberFormat value={number} messages.}
+         one={<>Hello <b>{name}</b>, you have <NumberFormat value={number} /> message.</>}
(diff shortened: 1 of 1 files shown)


FredyC:
@edorivai What makes you think that you need to wrap it into fragment there? As far I know it's definitely valid to have a string with <Element> in a middle without enclosing to anything.

https://codesandbox.io/s/k3ox964rw5

tricoder42:
@FredyC Yeah, but not as an attribute.

This doesn't work:

```
<div children={<>Start editing to see some<b> magic </b>happen!</>} />
```

But this does:

```
<div children={<>Start editing to see some<b> magic </b>happen!</>} />
```

tricoder42:
I usually use `Trans` tag for this, because it's omitted in final output, but I don't use Fragments shortcut at all.

Question is, which one should be preferred and documented? They both do the same:

```jsx
<Plural
   id="welcome"
   one={<>Hello <b>{name}</b>, you have <NumberFormat value={number} /> message.</>}
   other={<>Hello <b>{name}</b>, you have <NumberFormat value={number} /> messages.</>}
/>
```

```jsx
<Plural
   id="welcome"
   one={<Trans>Hello <b>{name}</b>, you have <NumberFormat value={number} /> message.</Trans>}
   other={<Trans>Hello <b>{name}</b>, you have <NumberFormat value={number} /> messages.</Trans>}
/>
```

tricoder42:
There's no logic at all for `Trans` component either. Both cases are (or at least should be) transformed into this anyway:

```jsx
<Trans 
  id="welcome"
  defaults="{number, plural, one {Hello <0>{name}</0>, you have {number, number}  message.} one {Hello <0>{name}</0>, you have {number, number}  messages.}}" 
  values={{ number }} 
  components={[<strong />]} />
```

I'll check if fragments are supported correctly in the transform plugin. I'm afraid they're treated as any other tag...

Actually, I've just realized that there's another bug in that example: missing `value` prop.

tricoder42:
@edorivai Awesome, thank you!

PR #344, opened 2018-09-27
LinguiJS v3.0

tricoder42:
See #334 for list of tasks. This PR is just to check the build status.


tricoder42:
Merged into `main`

PR #356, opened 2018-10-08
Correct Plural documentation in Macro reference

dsempel:


tricoder42:
Good catch! Thank you!

PR #358, opened 2018-10-09
Allow symlinks in srcPathDirs config for extract cli

brandonc:
Hello! Thanks so much for this project. It's been a pleasure to integrate so far. However, my project has an unconventional use case that would benefit from symlinks in srcPathDirs being followed so I'm wondering if we can make a minor change to the extract api.

### Notes
My understanding is that `lstatSync` only differs from `statSync` in that it does not follow symlinks. This change would allow symlinked directories to be read since isDirectory() would now return true and readdirSync already works for symlinks.

### DIFF
--- a/packages/cli/src/api/extract.js
+++ b/packages/cli/src/api/extract.js
@@ -51,7 +51,7 @@ export function extract(
     )
       return
 
-    if (fs.lstatSync(srcFilename).isDirectory()) {
+    if (fs.statSync(srcFilename).isDirectory()) {
       const subdirs = fs
         .readdirSync(srcFilename)
         .map(filename => path.join(srcFilename, filename))


tricoder42:
Hey @brandonc, this looks good! I'm not sure if I'll be able to release it this week, as I'm working towards v3. Maybe next Monday/Tuesday it could land in 2.8.

nLight:
Hey @tricoder42 any chance releasing 2.8? Thanks

tricoder42:
Oh, snap! I merged it with wrong commit tags (`docs:`) and then I missed it when writing release docs. This issue was already released in [v2.7.1](https://github.com/lingui/js-lingui/releases/tag/v2.7.1)

nLight:
Thank you! 🙇

PR #374, opened 2018-10-26
fix: disable Google Clojure Compiler rewritePolyfills behavior

### LINKED ISSUE #373: Lingui production version broke RN internal polyfill
**Describe the bug**
Production version of Lingui packages are minified(optimized) by Google Clojure Compiler and it by default adds build-in polyfill of Symbol.iterator and other (monkey patching global). This broke some of build-in polyfills in RN. As a result it lead to unexpected behavior of polyfilled Set/Map/etc. In my real use case Set used in RN Native Modules to store callbacks and after lingui applied own polyfill it becomes not iterable.
I know, it's a little bit hard to explain (i wasted 6 hours debugging it and now it's clear for me) 😄

**To Reproduce**
It's a little bit hard to provide you with actual demo of this issue, if you are similar with RN I can prepare a project showing issue.
Conditions: 
- It will only reproducible on env when there is no build-in Symbol in global (aka JSC on Android using RN).
- Looks like polyfill provided in lingui productions build is applied lazily (only when i'm using feature to include some jsx into Trans tag). I have investigate it, and this usage used ```Array.prototype.values``` method under the hood, which is polyfilled and also rewrite ```Symbol.iterator``` in global...

```

const sub = new Set();

sub.add(() => {});

const subscriptions = Array.from(sub.values()); // here  is Array.from tried to get iterator method from .values() iterator using Symbol.iterator
console.log(subscriptions.length) // 1 <- as expected

// but after rendering somewhere a Trans tag with inner jsx
   return (
     <Text>
       <Trans>
         Outer text <Text>inner text</Text>
       </Trans>
     <Text>
  );
}

// this code on same instance from above will return 0, because polyfill from Clojure Compiler monkey patched Symbol.iterator (was "@@iterator" and becomes "jscomp_symbol_iterator0")

const subscriptions = Array.from(sub.values());
console.log(subscriptions.length) // 1 <- as expected
```

**Expected behavior**
I don't know exactly how to fix it, but as for me lingui should not include polyfills (with monkey patching globals). I have tested is it possible and found a property in [Google Clojure Compiler](https://github.com/google/closure-compiler-js) to disable this behavior:

rewritePolyfills | true | Rewrite ES6 library calls to use polyfills provided by the compiler's runtime.
-- | -- | --

I have tried to compile lingui with ```rewritePolyfills: false``` enabled, and tested the result in my project, and it's worked! All tests in lingui repo worked too. So, maybe I can provide it as pull request. But anyway I need someone from lingui team to look on it, to be sure it's not broke something else 😄 

**Additional context**
- jsLingui version = 2.7.0
- Babel version - babel-core@7.0.0-bridge.0, @babel/core@7.1.0
- Your Babel config = using React Native
- I'm using @lingui/macro package.

**Possible workaround**
Looks like it can be temporary fixed adding aliases for @lingui packages:
```
"@lingui/react": "./node_modules/@lingui/react/cjs/react.development.js",
"@lingui/core": "./node_modules/@lingui/core/cjs/core.development.js"
```
So metro bundler can minify/optimize JS code itself

ikhilko:
Fixes #373

### DIFF
--- a/scripts/build/rollup.js
+++ b/scripts/build/rollup.js
@@ -50,7 +50,8 @@ const closureOptions = {
   warningLevel: "QUIET",
   applyInputSourceMaps: false,
   useTypesForOptimization: false,
-  processCommonJsModules: false
+  processCommonJsModules: false,
+  rewritePolyfills: false,
 }
 
 function getBabelConfig(updateBabelOptions, bundleType, filename) {


tricoder42:
Thank you!

I would like to try https://github.com/terser-js/terser in next release, do you know this minifier?

ikhilko:
Thanks! No, I don't use this minifier previously, but yes, you can give a try to it 😄

tricoder42:
Released in [v2.7.1](https://github.com/lingui/js-lingui/releases/tag/v2.7.1)

PR #376, opened 2018-10-30
Typescript docs: @types/lingui__macro installation

andys8:
`@types/lingui__macro` exists, but is not listed in the docs. Not sure, but I'd guess it's safe to add it. Review much appreciated.


tricoder42:
Thanks!

In the future I would like to ship typings with macro package, but for now it's good.

PR #401, opened 2018-11-20
Log original error with Babel compatibility info

sdgluck:
I encountered a problem where lingui thought I needed to install `babel-core` version 6 however the actual problem was being obscured because the error message which prompts lingui to log these instructions is not itself shared with the developer.

The actual problem I was having was that the lib `babel-preset-react-app` was not up-to-date, so the instructions to "please install missing Babel 6 core package" were misleading.

This PR makes the Babel compat. messages also print the original error in order to help people debugging cases like this.

Thanks for a great library!

Edit: Let me know what needs to be done re: tests and I will fix up the PR accordingly.

### DIFF
--- a/packages/cli/index.js
+++ b/packages/cli/index.js
@@ -1,1 +1,3 @@
+/* eslint-disable import/export */
+// This file is just used as an entry point for rollup.
 export * from "./src/lingui"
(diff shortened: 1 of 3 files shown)


tricoder42:
Great idea, thank you! I just made minor fixes with prettier to make linter happier.

tricoder42:

Issue #402, opened 2018-11-22
Marcos id generator


antonlyaptsev:
**Is your feature request related to a problem? Please describe.**
-
Consider existing implementation macros generates id from default translation if no key provided
```js
t`Attachment ${name} saved`

/*i18n*/{
  id: "Attachment {name} saved",
  values: { name }
}
```

It can be frustrating when the project uses file structure like this:
```
- common
  | - BuyButton
    | - index.js
- desktop
  | - BuyButton
    | - index.js
- touch
  | - BuyButton
    | - index.js
```
Developer should use different custom id in each BuyButton and keep in mind all of them
```js
// common/BuyButton/index.js
t('buy-button.common')`Buy ticket`

// desktop/BuyButton/index.js
t('buy-button.desktop')`Buy ticket`

// touch/BuyButton/index.js
t('buy-button.touch')`Buy ticket`
```

**Describe the solution you'd like**
-
I found experimental feature of `babel-plugin-macros` that can solve problem and improve developer experience (https://github.com/kentcdodds/babel-plugin-macros/blob/master/other/docs/author.md#config-experimental) by using custom id generator in config. It can looks like:
```js
// babel-plugin-macros.config.js
module.exports = {
  jsLingui: {
    id(expression, file) {
      // your code here
    }
  }
}
```
which produce something like this
```js
// desktop/BuyButton/index.js
t`Buy ticket`

/*i18n*/{
  id: "Buy ticket|desktop",
  defaults: 'Buy ticket'
}
```

**Describe alternatives you've considered**
-
Other alternatives was connected with writing own marcos with reusing some parts of `@lingui/macro`. But loosing support is not appropriate for my project

**Additional context**
-
no

I can make a PR. Are you interested in this feature?

tricoder42:
Hey @antonlyaptsev, although I don't sympathise with it, I saw similar requests in other issues.

# Implementation

## Config

All macros run at build time, which means we have access to filesystem and we could load [Lingui config](https://lingui.js.org/ref/conf.html) directly. Macros config is still experimental and I saw few comments where @kentcdodds mention, we would like to move away from them. On top of that, having a single config for all Lingui related features is definitely better. Also, in the future I would like to read more config in macro (e.g. whether messages should be replaced with translations at build time).

Using JSON config:

```json
{
  "extractId": "./config/i18n-id.js"
}
```

Using JS config:

```js
module.exports = {
  extractId: function (message, filename) { ... }
}
```

[...]

PR #406, opened 2018-11-30
Adds extractId option to conf

antonlyaptsev:
See problem in https://github.com/lingui/js-lingui/issues/402


antonlyaptsev:
I'm going to add documentation this weekend

antonlyaptsev:
Found bug with nested values. Fixing it

antonlyaptsev:
> Found bug with nested values. Fixing it

Order of imports affects the result. It doesn't look good 😕 

```js
import { plural, select } from '@lingui/macro'
```
processes all the plurals first. Next, selects processing will get already processed plurals and `transformChoiceMethod` will assign `undefined` as a value.

Making
```js
import { select, plural } from '@lingui/macro'
```
fixes the problem

antonlyaptsev:
@tricoder42 look at PR, please

tricoder42:
@antonlyaptsev PR looks good! I just re-ran tests in AppVeyor, because there were some strange failures.

The ordering problem is annoying. Thanks for noticing though! I'm gonna tackle it in v3 macros, because I'm refactoring them anyway.

antonlyaptsev:
@tricoder42 fixed these strange mocking failures 😈 
`macro` and `babel-plugin-transform-react` tests now pass

deecewan:
i'm pretty keen to see this PR go through. I'd like to have my message IDs be a hash of the contents - that will remove duplicates, for a start (assuming the context is the same - hash should include that), and the translation service I use doesn't support overly long IDs or IDs with spaces, so I have to explicitly add an ID to every translation currently.

tricoder42:
Hi @antonlyaptsev, sorry for delay. I'm not going to merge it to master, but rather to `next` branch which is the upcoming v3 release. There're many breaking changes, especially plugins are removed completely and replaced with macros.

I'm gonna keep this PR open until I cherry pick relevant changes to v3.

Thank you! 🙏

tricoder42:
Just a note: In the next version I would like to minify IDs for production build:

```js
// source code
const msg = t`This is not a love song`
```

**generated output - development/extracting**

```js
const msg = /*i18n*/{
  id: "This is not a love song"
}

// Message file
{
  "This is not a love song": "Translation of a love song"
}
```

**generated output - production**

```js
const msg = "a"

// Message file
{
  "a": "Translation of a love song"
}
```

For a while I was thinking about moving `extractId` call invocation to CLI, when messages are extracted, but with message minification we could actually use it for our advantage:

```js
function getIdProp(props, file, defaultValue) {
    let id
    const config = getConfig()

    if (
      config.extractId &&
      typeof config.extractId === "function" &&
      !props.defaults
    ) {
      id = config.extractId(props, file)
    }

    if (id == null) id = defaultValue

    if (process.env.NODE_ENV === "production" && config.minifyId) {
        const idMap = loadIdMap()  // load map of IDs, so it's persistent across builds.
        id = minifyId(id, idMap)
    }

    return id
  }
```

I don't have specific plan how to do the minification yet, any suggestions welcome. I'm just writing this down to not forget it.

stovmascript:
> especially plugins are removed completely and replaced with macros.

So changes added to `packages/babel-plugin-transform-react` should now be somewhere in `packages/macro`?

tricoder42:
@stovmascript Yes, indeed. You need to do it on several places:
- JS macro - [replacePathWithMessage](https://github.com/lingui/js-lingui/blob/next/packages/macro/src/macroJs.ts#L44)
- JS macro - [processDescriptor](https://github.com/lingui/js-lingui/blob/next/packages/macro/src/macroJs.ts#L184)
- JSX macro - [replacePath](https://github.com/lingui/js-lingui/blob/next/packages/macro/src/macroJsx.ts#L56) (any occurence of t.stringLiteral(id))

Config is already loaded in macro, but you need to pass `extractId` value to each [transformer](https://github.com/lingui/js-lingui/blob/next/packages/macro/src/macroJsx.ts#L56)

nathanforce:
I'd love to have this feature.  Is this PR still compatible after all this time?

PR #419, opened 2018-12-27
fix: made pseudolocales to take into account variable names

cstefanache:
Updated pseudolocale regex in order to take into account variable names. 
Previously the string: ```replace {count}``` would have been replaced with ```ŕēƥĺàćē  {ćōũńţ}```.
This would introduce an issue at runtime when the interpolator tries to find the variable ```ćōũńţ``` in the current variables context which does not exists.

With this change the variable names remain intact and the translation message is created as such: ```ŕēƥĺàćē {count}```

### DIFF
--- a/packages/cli/src/api/pseudoLocalize.test.js
+++ b/packages/cli/src/api/pseudoLocalize.test.js
@@ -36,7 +36,7 @@ describe("PseudoLocalization", () => {
     ).toEqual("{value, plural, one {# ƀōōķ} other {# ƀōōķś}}")
   })
 
-  it("should pseudolocalize variables", () => {
+  it("shouldn't pseudolocalize variables", () => {
     expect(pseudoLocalize("replace {count}")).toEqual("ŕēƥĺàćē {count}")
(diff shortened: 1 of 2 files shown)


packages/cli/src/api/pseudoLocalize.test.js
```diff
@@ -35,4 +35,9 @@ describe("PseudoLocalization", () => {
       pseudoLocalize("{value, plural, one {# book} other {# books}}")
     ).toEqual("{value, plural, one {# ƀōōķ} other {# ƀōōķś}}")
   })
+
+  it("should pseudolocalize variables", () => {
```
tricoder42: I guess this is a typo as it tests that *it shouldn't pseudolocalize variables*

```suggestion
  it("shouldn't pseudolocalize variables", () => {
```

cstefanache:
Sorry for the typo - I have checked in the suggested change

tricoder42:
Thanks again, @cstefanache!

tricoder42:

Issue #427, opened 2019-01-08
Plural strings with variable placeholders break pseudolocalization in specific cases


professorplumb:
**Describe the bug**
When a `plural` expression containing template strings with variable placeholders is marked and extracted, compilation fails due to pseudolocalization if at least one of the plural forms has the variable placeholder followed by just a single word with no punctuation.

**To Reproduce**
Run `lingui extract` on this file:
```
import { t, plural } from '@lingui/macro';

let messagesCount;
const messagesString = {
  1: 'one',
  2: 'two',
  3: 'three',
  4: 'four',
  // ...
};
const expr = plural({
  value: messagesCount,
  one: t`${messagesString} message`,
  other: t`${messagesString} messages`,
});
```
Extracted form:
```
{messagesCount, plural, one {{messagesString} message} other {{messagesString} messages}}
```
Running `lingui compile` on this with a `pseudoLocale` set results in this compilation for the pseudolocale:
```
{messagesCount, plural, one {{mēśśàĝēśŚţŕĩńĝ} message} ōţĥēŕ {{mēśśàĝēśŚţŕĩńĝ} messages}}
```
which results in an error from the messageformat parser [here](https://github.com/lingui/js-lingui/blob/master/packages/cli/src/api/compile.js#L18-L23)

**Expected behavior**
Compiling the above extracted form should result in
```
{messagesCount, plural, one {{messagesString} mēśśàĝē} other {{messagesString} mēśśàĝēś}}
```

**Additional context**
The issue appears to be that `PluralRegex` in `packages/cli/src/api/pseudoLocalize.js` expects at least two words (or one word plus any punctuation) following the final variable placeholder in an extracted plural expression.

[This](https://regex101.com/r/1ENkXn/1) regex101 example shows the problem: the regex and the first two test strings are the same as the `PluralRegex` example from `pseudoLocalize.js`.  The third shows that the regex operates correctly with variable placeholders instead of `#` as long as there are multiple words following the placeholder.  The fourth demonstrates that the issue explained above results from `PluralRegex` misidentifying the closing `}` of a variable placeholder for the closing `}` of a `plural` expression argument.

Note that if you enter a space plus any word after "message(s)" in both plural expressions of the fourth string, or if you add any punctuation before or after "message(s)", the regex detects the delimiters correctly.

I am playing with the `PluralRegex` to see if there's an easy way to fix this problem (which, to be fair, is an edge case) without breaking other expressions; if I find one I'll create a PR for this issue.

[...]

PR #428, opened 2019-01-09
Issue #427: fix edge case bug in plural pseudolocalization

professorplumb:
Fixes issue #427.
Depends on PR #419 - the test added here will fail on this branch until that PR is incorporated.

This fix will still fail in the case that a variable placeholder in a template string inside a `plural` macro is followed by a single word which matches one of the plural rules, e.g. `{count, plural, one {{countString} other} other {{countString} others}}`.

This is a much more extreme edge case, though - it must be
1. A template string
1. Inside a `plural` expression
1. With a variable placeholder
1. Followed by `zero|one|two|few|many|other`
1. With no other words or punctuation after that word
1. With pseudolocalization enabled

IMO this is an acceptable tradeoff for a more readable regular expression.


tricoder42:
Thank you, @professorplumb. I think this is the best what we can get with regular expressions. 👍 

Other method is use message-parser tokenizer, traverse through text tokens, pseudolocalize them and then join them back to message format, but that seems to be overengineered at the moment.

This patch looks good, thanks!

tricoder42:

PR #438, opened 2019-01-19
Fix typo on macro Babel plugin config

haruelrovix:
Otherwise, you'll get an error:

```
Metro Bundler ready.

Loading dependency graph, done.
error: bundling failed: ReferenceError: Unknown plugin "macro" specified in
{{ a bunch of line errors }}
...
```


tricoder42:
Ah, you're right. Thank you! 👍

PR #444, opened 2019-01-25
Update testing example

nuc:
Hey there,

The current testing example is a little bit problematic, especially when you want to mount and test stateful components.

Enzyme's mount gives you access only to the root component's state. That means that by doing: `<I18nProvider><YourComponent /></I18nProvider>` you can't access `YourComponent`'s state for testing.

In `react-intl`, we were using a higher order component [taken from this gist](https://gist.github.com/mirague/c05f4da0d781a9b339b501f1d5d33c37/), which extracts the context from the provider and passes it down to the component tree. That solves the issue, and makes testing great again (:yikes:).

We have adapted that solution for the I18nProvider and it works great for us.

I updated the documentation, and I hope you also find it useful.

Cheers 🍻

### DIFF
--- a/docs/guides/testing.rst
+++ b/docs/guides/testing.rst
@@ -3,22 +3,93 @@ Testing
 
 .. warning:: This is an incomplete draft of a guide.
 
-Components using :component:`Trans` or :js:func:`withI18n` must be wrapped in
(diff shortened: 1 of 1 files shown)


tricoder42:
Interesting, thank you very much!

PR #454, opened 2019-02-13
Don't prevent babel options to help setup lingui in monorepos

adamscybot:
Currently, lingui complains with option validation issues if you attempt to use the `extends` or `rootMode` babel options. In a monorepo, with Babel 7, you often want message extraction to happen in each module, but using a common babel configuration that exists in a top level "workspace"

We should allow these options without throwing warnings. I'm not sure we should actually have any validation on the babel options (since whats allowed depends on babel version, and theres not much need to limit he feature set) so potentially, this validation could be removed entirely.

### DIFF
--- a/packages/conf/src/index.js
+++ b/packages/conf/src/index.js
@@ -44,6 +44,8 @@ export const defaultConfig = {
 const exampleConfig = {
   ...defaultConfig,
   extractBabelOptions: {
+    extends: "babelconfig.js",
+    rootMode: "rootmode",
     plugins: ["plugin"],
     presets: ["preset"]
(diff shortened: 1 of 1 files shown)


tricoder42:
@adamscybot Ah, sorry, completely forgot about it.

yeah, I had another problem with config validation before, so I'm gonna probably move away from jest-validate in the future.

tricoder42:
And thanks for the PR! 👍

tricoder42:
Released in [2.7.4](https://github.com/lingui/js-lingui/releases/tag/v2.7.4)

PR #476, opened 2019-03-19
Remove illegal comma from json example

grgilad:


tricoder42:
Good catch, thank you! 🙏

PR #477, opened 2019-03-20
Fix escaped backticks in macro-transformed template strings

ackwell:
To use a literal backtick in a template string (such as for markdown code blocks), it needs to be escaped (```const str = `\``;```).

Unfortunately, this gets picked up as ``\\` `` due to `value.raw`, which ends up as a literal ``\` `` in the translation string.

In turn, this leads to final translation strings ending up with the literal ``\` ``, which is broken syntax, throwing a `Error: Can't parse message. Please check correct syntax"`.

This PR provides a quick fix for the issue, allowing backticks to be used.

### DIFF
--- a/packages/macro/src/js.js
+++ b/packages/macro/src/js.js
@@ -239,7 +239,7 @@ export default function({ types: t }) {
 
     parts.forEach(item => {
       if (t.isTemplateElement(item)) {
-        props.text += item.value.raw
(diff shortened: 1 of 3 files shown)


tricoder42:
It looks mostly harmless.

Good job! 👍

tricoder42:
Released in [v2.8.0](https://github.com/lingui/js-lingui/releases/tag/v2.8.0)

PR #479, opened 2019-03-23
[WIP] Macros refactoring

tricoder42:
## Refactoring

JS/JSX macros and ICU MessageFormat were separated. The structure is following:

**tokenizer -> message format -> transformer**

Tokenizer parses AST and produces tokens, which are fed into message format (ICU only at the moment). Message format produces `message`, `values`, `id`, `comment` which are used in transformed code.

Example:

```js
t`Hello ${firstName}`
```

is tokenized into:

```js
[
  {
    type: "text",
    value: "Hello "
  },
  {
    type: "arg",
    name: "firstName",
    value: {/* node from AST, identifier firstName */}
  }
]
```

ICU message format reads tokens and produces:

```js
{
  message: "Hello {firstName}",
  values: {
    firstName
  }
}
```

and macro is finally transformed into:

```js
i18n._("Hello {firstName}", { firstName })
```

The main reason for this refactor was deduplication of code and possible extension for other AST parsers (e.g. TS) and other message formats (e.g. Fluent).

- [macroJs.ts](https://github.com/lingui/js-lingui/blob/b0cfbdf9c3eb92fb37bf3c54d313297d3d7ed3e1/packages/macro/src/macroJs.ts) - transform JS macros
- [macroJsx.ts](https://github.com/lingui/js-lingui/blob/b0cfbdf9c3eb92fb37bf3c54d313297d3d7ed3e1/packages/macro/src/macroJsx.ts) - transforms JSX macros
- [icu.ts](https://github.com/lingui/js-lingui/blob/b0cfbdf9c3eb92fb37bf3c54d313297d3d7ed3e1/packages/macro/src/icu.ts) - ICU message format generator

## What's missing

I still have to figure out API of core library. My original intention to use global variable `i18n` doesn't seem to be a good idea after all. It'll be problematic in multi-threaded environments (e.g. Next.js), it's harder to extend when multiple i18n objects are used (e.g. when sub-package handles it's own translation).

I'm back to original idea that to use JS macros, first we need to get `i18n` object. That variable will mostly sit unused in the code.

For example, following code:

```
function App({ firstName }) {
  const i18n = useLingui()

  return t`Hello ${firstName}`
}
```

becomes

```
function App({ firstName }) {
  const i18n = useLingui()
  return i18n._("Hello {firstName}", { firstName })
}
```

I didn't want to go this way, because now linters will complain about unused variable, but that should be fixed in linter. When considering two options, either broke the code using global variable or broke the linter with valid code, I would rather go the other way.


FredyC:
Thank you for opening this. So it seems there are two major standing issues.

## Where to get the `i18n` instance

I am not following why you would need to enforce Context here. Sure, multi-threading environment might need it, but it could be opt-in and not the only option. The macros don't need to care where they get that instance, right? In environments like CRA it will work with global variable just fine. 

It can be primary recommendation to use Context, but there are surely some use cases of translations happening outside of the React components and you would have to be passing `i18n` through arguments and that's rather bad.

There is one benefit of using Context here. With autoimport feature in VSCode I do mistakes where I am importing `i18n` from `@lingui/core` instead of my exported one. It's so subtle bug and it makes translation not working because of that.

## Unused `i18n` variable

The ESLint is solvable as it has the configuration to ignore some patterns. But TypeScript is just "all in or out" variant, there is `noUnusedLocals` setting. You would have to do this all over the place.

```ts
function App({ firstName }) {
  // @ts-ignore I can write a comment here, not to disable a specific rule
  const i18n = useLingui()

  return t`Hello ${firstName}`
}
```

Besides this kinda bad if you later decide to move localization somewhere else and the `i18n` stays behind because it won't be marked as unused. It's not a huge deal, but it's weird for sure.

Perhaps as an alternative it should be possible to do this. Sure it adds some extra clutter, but at least it does not suffer from unused problem.

```js
function App({ firstName }) {
  const i18n = useLingui()

  return i18n(t`Hello ${firstName}`)
}
```

## Need to use `i18n` variable name

Frankly, I do typos with this all the time. I would love to opt for something like `IN` instead. With the version proposed above, it would be easy to accept any naming convention. Macro would grab any wrapper there is. Or that's my simplistic view, there might be some issues as well.

bopfer:
I can only speak from the perspective of client side React apps, but the "global" i18n is basically how may apps are currently using it.

I have an app-wide context with just a few things being tracked, one of them being the language to use.  The browser's language is detected, catalog loaded, and `i18n` exported for use throughout the app.  Example:

```javascript
/* In the app context */
const i18n = setupI18n();

/**
 * In the context's actual component logic... do all the stuff to detect the language,
 * lazy load the proper catalog, and render the I18nProvider (inside the context provider)
 */

export { i18n };  /* along with the other context exports */
```

Exporting `i18n` allows easy access anywhere in the app w/o needing to load the context itself.  It's just an import line from the app context.  This could arguably be it it's own context/component.  But, it seems to make sense to my team to just have it in the global app context.

This all seems to work well, but maybe there is a better way.  Especially when considering SSR or other multi-threaded envs.

tricoder42:
> The macros don't need to care where they get that instance, right? In environments like CRA it will work with global variable just fine.

Well, it depends. There're two approaches. 

## 1. Message Descriptor

```js
const hello = t`Hello ${world}`

// becomes
const hello = {
  id: 'Hello {world}',
  values: { world }
}
```

Then it's up to you to get the i18n instance:

```js
// from global config
import i18n from "./config/i18n"

// from local context
const i18n = useLingui()

// and then just pass message descriptor to it
i18n(t`Hello ${firstName}`)
```

I thought we're trying to avoid this variant. One problem is callable object. We talked about it before. You showed me an implementation, but it didn't work. I google around and callable objects are just problematic with IE compatibility issues.

We could make a workaround though:

```
i18n.translate(t`...`)

// or destructure it from hook
const { translate } = useLingui()
translate(t`...`)
```

## 2. Function call

```js
const hello = t`Hello ${world}`

// becomes
const hello = i18n._({
  id: 'Hello {world}',
  values: { world }
})
```

Then you need to get `i18n` from somewhere. Either manually (with TS burden) or automatically. That could be configurable (`import { i18n } from `@lingui/core` by default or anything like `import { i18n } form './config/i18n'). Question is how to locally disable it, so we can optionally use context. Comment on top of the file? `// @lingui/macro disable-i18n-import`? That could probably work, but it just feels complicated.

## Conclusion

Personally I would go with ```i18n(t`...`)``` syntax, because then it's clear that macros are optional. We just have to figure out the proper API. I know you dislike `i18n._` (I don't), but `i18n.translate` might be a better alternative.

FredyC:
Ugh, I kinda feel you have responded so something different than I was talking about :)

```js
function App({ firstName }) {
  const i18n = useLingui()

  return i18n(t`Hello ${firstName}`)
}
```

I meant by this to essentially take care of the problem of unused variables. It doesn't need to be a callable object, it can just transform to `i18n._` as usual. Or at least I hope that is possible. I honestly don't know if macros can touch other parts of AST. In case it's possible, you would not even need to rely on specific `i18n` name, it could just suffix `._` to anything that's around a macro call.

The `i18n.translate` is like a public API? In that case that's awfully verbose :) If you mean it as internal runtime, I don't care really :)

tricoder42:
Aha!

You mean this:

```js
function App({ firstName }) {
  const i18n = useLingui()

  return i18n(t`Hello ${firstName}`)
}
```

would be transformed into this?

```js
function App({ firstName }) {
  const i18n = useLingui()

  return i18n._("Hello {firstName}`, ...)
}
```

Well, it's definitely unusual. Macros are intended to transform any node _below_ in AST. But of course, they can also access any node _above_.

> The i18n.translate is like a public API? In that case that's awfully verbose

What would you suggest? Considering it mostly will be used with `t` macro and sometimes with `plural`, `select` or other macros.

```js
// all this is public API, what you see in your code
i18n._(t`...`)
i18n.t(t`...`)
// ?
```

FredyC:
> What would you suggest? Considering it mostly will be used with `t` macro and sometimes with `plural`, `select` or other macros.

I don't think it would hurt to have multiple aliases. Someone might prefer a less verbose version `i18n._` and someone likes to be explicit.

> Well, it's definitely unusual. Macros are intended to transform any node _below_ in AST. But of course, they can also access any node _above_.

Yes, that's what I meant :) Beauty is that could be completely optional. If someone is bothered by an unused variable, they can just wrap that call. Or they can directly call `i18n._` which would not be transformed in any way and still satisfy the linting.

tricoder42:
This is usually where I get stuck. There're decisions that needs to be made before moving on. Making too many entry points for theoretical usecases is maintenance nightmare.

> I don't think it would hurt to have multiple aliases. Someone might prefer a less verbose version i18n._ and someone likes to be explicit.

[There should be one-- and preferably only one --obvious way to do it.](https://www.python.org/dev/peps/pep-0020/)

I'm all in for explicit `i18n` which comes either from global import or local context. I'll schedule some time this week to think about it. Once we have this, the macros can be finalized and we can finally move on.

FredyC:
> [There should be one-- and preferably only one --obvious way to do it.](https://www.python.org/dev/peps/pep-0020/)

Well, I would say you can always deprecate and remove alias that people seem to dislike more. Personally, I would hate to see `i18n.translate` all over the code, it adds too much clutter. The message is what's important, not names the API. There is a static typing to check for correctness of the call after all.

renchap:
I am not expert in babel macros nor LinguiJS, but could this be something that work:

```ts
function App({ firstName }) {
  const i18n = useLingui()

  return i18n.t`Hello ${firstName}`
}

Transformed into:

```ts
function App({ firstName }) {
  const i18n = useLingui()

  return i18n._("Hello {firstName}`, ...)
}
```

This makes it look like `i18n.t` is a real method which solves some problems:
- it "makes sense" when reading the code, you understand where `t` comes from
-  you can create typings for it
- `i18n` is seen as used by the various linters

Or did I miss something obvious?

FredyC:
@renchap This is exactly how it works with current babel plugin :)

Macro is slightly more limited in this. Considering that `i18n` is a runtime object, it cannot be also the macro to do the transformation. The `t` in here is the actual macro which can be replaced with something else.

JSteunou:
I'm really not a big fan have having unused variable. We already have a webpack plugin to inject react everywhere (but that one is easy).

Why cannot we have both world, macros with global i18n instance or runtime for this in need of specific instance

```js
  return t`Hello ${firstName}`
```

and

```js
  const i18n = useLingui()
  return i18n.t`Hello ${firstName}`
```

JSteunou:
Or macro for everyone if the produced code can load the nearest i18n context. Let the provider provide. Developer own responsabilty

tricoder42:
As @FredyC said, `i18n.t` can't be macro, unless the `i18n` is a macro. It's not ideal. Plugins works that fine thought.

---

I tried to look at macros from different angle. Here's my train of thoughts:

Let's start with a code *without* internationalization library.

```js
value === 1 ? `{value} book` : `{value} books`
```

Soon we would realize that `value === 1` depends on language (i.e. plural function)
and there might be more than two plural forms. So we would probably come up with
a function, which maps `value` to `pluralForm` (i.e. `one`, `few`, `other`):

```js
plural(value, {
   one: "# book",
   other: "# books"
})
```

However, if we want this function to return translated message, we need to
transform it into a i18n message format. In other words, we need to serialize
`plural(...)` call into a string:

```js
i18n("{value, plural, one {# book} other {# books}}", { value })
```

We could rewrite it into an object with `id` and `values` props:

```js
i18n({
   id: "{value, plural, one {# book} other {# books}}",
   values: { value }
})
```

Which is especially handy if we want to add additional i18n metadata, like message ID
or comments for translators:

```js
i18n({
   id: "msg.plural",
   comment: "Verbose description of number of books",
   values: { value }
   message: "{value, plural, one {# book} other {# books}}",
})
```

One way to sort these props is that `id` and `comment` are metadata,
while `values` and `message` determines the final message:

```js
i18n({
   // metadata
   id,
   comment,

   // message
   values,
   message
})
```

But more useful is to group props, which are required at runtime and the rest, which
are optional and required only in message catalogs:

```js
i18n({
   // required/runtime
   id,
   values,

   // optional/definition
   comment,
   message
})
```

Based on this, we can construct the API for translation function:

```js
i18n._(id: string)
i18n._(id: string, values: Object)
```

This is fine until we want to add optional data as `options` parateter:

```js
i18n._(id: string, values: Object, options: Object)
i18n._(id: string, options: Object) // clash!
```

The last call has the same signature as `i18n._(id, values)`, which is bad.

Another option: Keep first two shortcuts, but if we want to add options,
we need to call function with full message descriptor:

```js
i18n._(id: string)
i18n._(id: string, values: Object)
i18n._({
   id: string,
   values: Object,

   comment: string,
   message: string,
})
```

Now the question is, how to write macro which maps onto this API.

---

The point of having macro (`plural` in this case) is to transform this:

```js
i18n(plural(value, {
   one: "# book",
   other: "# books"
}))

// ↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓ ↓

i18n({
   id: "{value, plural, one {# book} other {# books}}",
   values: { value }
})
```

Metadata doesn't require compile-time transformations, but we need to include them somehow, somewhere.

I tried several approaches to include `id` and `comment` into `t`, `plural` and other macros.
None of them was perfect, they all had some drawbacks. The most important problem: Once
you start adding `id` and `comments` to the code, it becomes more verbose and harder to
navigate, which is opposite of what macros are trying to achieve.

So, another approach - if you want to add `id` or `comment` to you message, you
need to use completely different macro and define the message:

```js
const hello = defineMessage({
   id: "msg.hello",
   comment: "Comment for translators",
   message: t`Hello ${"name"}`
})

const plural = defineMessage({
   id: "msg.plural",
   comment: "Comment for translators",
   message: plural('value', {
      one: "# book",
      other: "# books"
   })
})
```

Later, in your code, you can translate and format the message:

```js
i18n(hello, { name: "World" })
i18n(plural, { value: 42 })
```

Another approach. Since it's common to have several messages per component or file, it would be handy to define them all in one place:

```js
const messages = defineMessages({
   hello: {
      id: "msg.hello",
      comment: "Comment for translators",
      message: t`Hello ${"name"}`
   },
   plural: {
      id: "msg.plural",
      comment: "Comment for translators",
      message: plural('value', {
         one: "# book",
         other: "# books"
      })
   }
})
```

Later in the code:

```js
messages.use(i18n)
messages.hello({ name: "World" })
messages.plural({ value: 42 })
```

Boom. API ready for typechecking:

```js
messages.plural({ vale: 42 })
// Argument of type '{ vale: number; }' is not assignable to parameter of type '{ value: number; }'.
```

---

## Example with hooks

```js
function Component({ numMessages }) {
  // load i18n from context and bind messages to it
  const { messages } = useLingui(messages)
  
  console.log(messages.hello())
  console.log(messages.plural({ numMessages }))
}
```

# Simple example without IDs

function Component({ name, numMessages }) {
  const { i18n } = useLingui()
  
  console.log(i18n._(t`Hello ${name}`)))
  console.log(i18n._(plural(numMessages, { ... }))
}
```

# Wrap up

This API works well in simple cases, when you don't need `id` and `comment`. If you do, it allows more structured approach, when you extract messages into object (similar to StyleSheets in react-native). 

- It prevents from defining two messages with different defaults (common question).
- The link between message definition and code is typechecked, it's harder to make an error

```js
// this can have typo in msg ID or variable name
i18n._('msg.hello', { name })

// this can't
msg.hello({ name })
```

FredyC:
An interesting new approach. I especially love the idea of binding `i18n` internally. It's a pity that it's helpful for a custom ID world only. I mean it would work with the message as ID too, but you always have to specify some "key" to be able to reference that message. That kinda beats the purpose of it. Although in some cases (especially lazy scenarios) it might be worth to opt for that approach.

I am thinking if it's actually necessary to specify `id` prop in the `defineMessages` when you already have the prop key there. Being explicit is great until it isn't. Feels a bit too verbose and duplicate imo. In case you don't want to infer the id from the key automatically, there could be an option to opt-in/out of that.

tricoder42:
> Although it's a pity that it's helpful for a custom ID world only. I mean it would work with the message as ID too, but you always have to specify some "key" to be able to reference that message. That kinda beats the purpose of it. Although in some cases (especially lazy scenarios) it might be worth to opt for that approach.

I would still find it valuable:

```js
const messages = defineMessages({
   // if you want to use id, comment or other metadata, use message
   hello: {
      comment: "Comment for translators",
      message: t`Hello ${"name"}`
   },
   // otherwise just assign message to the key
   plural: plural('value', {
       one: "# book",
       other: "# books"
    })
})

function Component({ name, value }) {
  // load i18n from context and bind messages to it
  const { messages } = useLingui(messages)
  
  console.log(messages.hello({ name }))
  console.log(messages.plural({ value }))
}
```

> I am thinking if it's actually necessary to specify id prop in the defineMessages when you already have the prop key there. Being explicit is great until it isn't. Feels a bit too verbose and duplicate imo. In case you don't want to infer the id from the key automatically, there could be an option to opt-in/out of that.

I see your point and we might opt-in such behavior. But for several reasons I wouldn't assume that the key in `messages` is the same as message ID.

1. Message IDs might be much more complex:

[Message IDs](https://twitter.com/andrewingram/status/1085528395281248263)

```js
const messages = defineMessages({
   hello: {
      id: "User Profile / Welcome Message",
      comment: "Comment for translators",
      message: t`Hello ${"name"}`
   },
   plural: {
      id: "Bookcase / Number of books",
      message: plural('value', {
         one: "# book",
         other: "# books"
      })
   }
})
```

2. Refactoring might have unexpected side-effect of changing keys in all message catalogs:

```diff
const messages = defineMessages({
-   hello: { // implicit messageId: `hello`
+  aloha: { // implicit messageId: `aloha`   
      comment: "Comment for translators",
      message: t`Hello ${"name"}`
   },
})
```

After such change, all catalogs must be changed and previous translation matched. That might be unexpected and confusing.

3. You will limit your messageIDs to valid JS identifiers, so no `msg.hello` or `Message / Hello`.

But, yeah, we could make it opt-in. There's a room for improvement and it doesn't affect the explicit API.

tricoder42:
Let's merge it and move on.

PR #482, opened 2019-03-28
Add Turisto to the showroom

gothy:
Another thankful "client" of Lingui :)


tricoder42:
Cool! It's great to have you!

PR #500, opened 2019-05-16
Update showroom.rst

cezarneaga:
just finalised migration from react-intl for the react sections of our site


tricoder42:
Amazing! 🇨🇿

PR #501, opened 2019-05-16
Optional --config argument for advanced configuration scenarios

brandonc:
## Overview

I have an advanced configuration scenario that requires multiple locale targets and with multiple source directories. I want to end up with multiple catalogs that are loaded in different circumstances.

Presently, I work around the obvious config issue by swapping out my .linguirc file in between runs of `lingui extract` and `lingui compile`. This isn't ideal and I thought lingui might benefit from a CLI  option to specify a configuration file.

This change allows a `--config <path>` option on add-locale, compile, and extract commands. If the path exists, I use `cosmiconfig#loadSync` instead of the normal config file probe.

If you find this change welcome, I can work more to add details to the docs.

## Tradeoffs

- Adding a new option could make each command to be perceived as slightly more complex

### DIFF
--- a/docs/ref/cli.rst
+++ b/docs/ref/cli.rst
@@ -44,6 +44,14 @@ Install
          }
       }
 
+Global options
+==============
+
+.. lingui-cli-option:: --config <config>
+
+Path to LinguiJS configuration file. If not set, the default file
+is loaded as described in :doc:`LinguiJS configuration </ref/conf>` reference.
+
 Commands
 ========
 
--- a/packages/cli/src/api/pseudoLocalize.test.js
+++ b/packages/cli/src/api/pseudoLocalize.test.js
@@ -35,8 +35,12 @@ describe("PseudoLocalization", () => {
       pseudoLocalize("{value, plural, one {# book} other {# books}}")
     ).toEqual("{value, plural, one {# ƀōōķ} other {# ƀōōķś}}")
     expect(
-      pseudoLocalize("{count, plural, one {{countString} book} other {{countString} books}}")
-    ).toEqual("{count, plural, one {{countString} ƀōōķ} other {{countString} ƀōōķś}}")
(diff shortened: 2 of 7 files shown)


tricoder42:
Hey @brandonc, thank you! This is handy! I've added one paragraph to docs and it's ready for release.

Could you please take a look at [proposal](https://lingui.js.org/rfc/002_catalogs.html) for new configuration for catalog? I'm gonna finish the PR once I merge #499 and it would be great to get a feedback from you, because you use multiple source directories and multiple locale targets.

tricoder42:
Released in [v2.8.0](https://github.com/lingui/js-lingui/releases/tag/v2.8.0)

hamczu:
This is a breaking change for many webpack users who specify --config option for webpack.

brandonc:
@hamczu I'm having trouble picturing the use case. Are you using webpack-cli to invoke lingui? This should work even if you are chaining commands together (`webpack --config myconfig.js && lingui --config myconfig.yaml`) Can you provide more details?

tricoder42:
@brandonc I guess the problem might be regular webpack build - when you run `webpack --config ...` and you're using `@lingui/loader`, it'll load incorrect config. We don't check `--config` presence in command args, but rather in `process.args`.

hamczu:
yep, exactly this is happening - I'm just running a webpack dev server: `webpack-dev-server --config some-path`

and in my (react) application I use @lingui/loader to load the translated messages:
`require('@lingui/loader!../en/messages.json')`

brandonc:
@tricoder42 It seems to me that for this feature to be compatible with @lingui/loader, there needs to be options loading support added there. I'm working on that change, in addition to passing in command args to getConfig instead of checking process.argv

brandonc:
Fix: https://github.com/lingui/js-lingui/pull/509

PR #504, opened 2019-05-18
Add declaration files on root of react and core packages

lucasecdb:
we should also copy these files to the build folder, but I've never used rollup before and I'm kinda lost in the configurations. can anyone help?

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -1,0 +1,1 @@
+export * from "./cjs"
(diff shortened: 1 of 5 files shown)


tricoder42:
I thought that I added `types` entry to `package.json`, but it's missing in every package.

Could you please check if adding:

```json
{
  "types": "cjs/index.d.ts"
}
```

solves the problem? That would work for `core` and `react` package.

For `macro` package, we need to figure out how to copy the `index.d.ts` file anyway :)

tricoder42:
As for rollup, this should work: https://www.npmjs.com/package/rollup-plugin-copy

lucasecdb:
oh, nice! didn't know about the `types` entry

PR #505, opened 2019-05-18
docs: fix 3.x migration imports

lucasecdb:


docs/releases/migration-3.rst
```diff
@@ -41,7 +41,8 @@ is simplified and accepts ``i18n`` manager, which must be created manually:
 .. code-block:: diff
 
    - import { I18nProvider } from '@lingui/react'
-   + import { setupI18n, I18nProvider } from '@lingui/react'
+   + import { setupI18n } from '@lingui/core'
+   + import { I18nProvider } from '@lingui/react'
```
tricoder42: Could you please simplify the diff a bit? Now when the `setupI18n` is imported from another module, we don't have to remove `import { I18nProvider } from '@lingui/react'`. So it should be just:

```diff
  import { I18nProvider } from '@lingui/react'
+ import { setupI18n } from '@lingui/core'
  import catalogEn from './locale/en/messages.js'
```

lucasecdb: oh, sure!

PR #509, opened 2019-05-21
Fixes --config option processing

brandonc:
### Overview

Processing process.argv inside getConfig can cause problems when used in conjunction with @lingui/loader since the --config arg is also used by webpack

Instead of processing process.argv, getConfig now accepts the path to a custom config file. Each command now passes the config option from commander into getConfig. Webpack loader now accepts an option "config" that can be used to specify a config file.

### Notes

Should all lingui options be allowed via webpack loader options? loader could be modified to assign all options to getConfig output.

I think I fixed an unrelated bug with lingui/loader. compileNamespace and pseudoLocale config options  were not being passed to createCompiledCatalog, so I assume that all loader compiled output was being exported as cjs.

### Testing

I smoke tested the fix by implementing lingui/loader with a config option while simultaneously using the --config option with webpack. I also verified that `lingui extract --config` continued to function as expected.

### DIFF
--- a/packages/cli/src/lingui-add-locale.js
+++ b/packages/cli/src/lingui-add-locale.js
@@ -55,7 +55,7 @@ if (require.main === module) {
 
   if (!program.args.length) program.help()
 
-  const config = getConfig()
+  const config = getConfig({ configPath: program.config })
   if (program.format) {
     const msg =
       "--format option is deprecated and will be removed in @lingui/cli@3.0.0." +
--- a/packages/cli/src/lingui-compile.js
+++ b/packages/cli/src/lingui-compile.js
@@ -140,7 +140,7 @@ if (require.main === module) {
     })
     .parse(process.argv)
 
-  const config = getConfig()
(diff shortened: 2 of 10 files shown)


brandonc:
@tricoder42 Thanks! What is the next step?

tricoder42:
Sorry I completely forgot. I'll release it in few minutes.

tricoder42:
Released in [v2.8.2.](https://github.com/lingui/js-lingui/releases/tag/v2.8.2)

semoal:
Hi guys actually my code is crashing due this release.
Captura de pantalla 2019-05-22 a las 22 46 16

Any suggerence? @tricoder42

tricoder42:
@semoal Sorry for that. The tests for loader are broken, so this bug sneaked into the release :/

I'll fix it in few minutes.

tricoder42:
@semoal Fixed in [v2.8.3](https://github.com/lingui/js-lingui/releases/tag/v2.8.3). Thanks for report!

Issue #511, opened 2019-05-24
[loader] 2.8.3 SyntaxError: Unexpected token export


rwu823:
**Describe the bug**
Throws errors:

```
SyntaxError: Unexpected token export
```

**To Reproduce**
Steps to reproduce the behavior, possibly with minimal code sample, e.g:

```
import en from "@lingui/loader!./locale/en/messages.json"
```

**Expected behavior**
Import the messages without any errors, version `2.8.0` is fine

**Additional context**

- jsLingui version `2.8.3`,

ktunkiewicz:
I had the same issue and managed to get around it. The reason for this behaviour is that the CLI tool is choosing to use Babel 6 because some of your other project dependencies pull in babel-core@6.

Correct me if I'm wrong, but as far as I understood the Lingui works with both Babel 6 or 7 and chooses to use one of these depends on which one is available. Obviously, it prefers 6 over 7.

Anyways, when you run the CLI tool, it chooses babel 6, and because (I'm guessing now) you don't have the `.babelrc` file (I'm guessing that you use `babel.config.js`) it will not load any babel presets nor plugins.

There is a way to inject custom Babel configuration into `lingui extract` run, like this:
https://lingui.js.org/ref/conf.html#extractbabeloptions

This entry in package.json solved my problem:
```json
  "lingui": {
    "sourceLocale": "en",
    "fallbackLocale": "en",
    "localeDir": "<rootDir>/src/locale",
    "srcPathDirs": [
      "<rootDir>/src"
    ],
    "format": "po",
    "extractBabelOptions": {
      "presets": [
        "babel-preset-flow",
        "babel-preset-react",
        "@lingui/babel-preset-react"
      ],
      "plugins": [
        "transform-object-rest-spread",
        "dynamic-import-node",
        "transform-class-properties"
      ]
    }
  }
```

The above basically mimics what's in my Babel 7 config, just using the version 6 packages.
I also had to install all these packages:
```
npm install --save-dev babel-plugin-dynamic-import-node babel-plugin-transform-class-properties babel-plugin-transform-object-rest-spread babel-preset-flow babel-preset-react
```

I would love to see an option to simply **force** lingui CLI to always use Babel 7 @tricoder42 is there a way to do that?

tricoder42:
[...]

PR #517, opened 2019-06-05
Add extractBabelVersion option

ktunkiewicz:
This is a proposal to add config entry to select between Babel 6 and 7 to be used by the extract tool.

Related issue: https://github.com/lingui/js-lingui/issues/511

When this PR is merged you can use this config option:
```
extractBabelVersion: 6 (default) or 7
```
to force the extract tool to import babel 6 (`babel-core`) or babel 7 `(@babel/core)` core package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
   },
   "devDependencies": {
     "babel-core": "^6.26.3",
+    "@babel/core": "^7.0.1",
     "babel-eslint": "^8.2.6",
     "babel-jest": "^23.4.0",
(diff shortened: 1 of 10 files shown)


ktunkiewicz:
ah, sorry, my local npm settings spoiled the lock file, will correct that in a minute

tricoder42:
Hey @ktunkiewicz, thank you for your PR 👍  Lingui v3 is going to use Babel 7 only. We might revisit backward compatibility after v3 release.

Issue #457, opened 2019-02-19
Allow escaping placeholders


SilentImp:
**Describe the bug**

Both react [FilePond](https://pqina.nl/filepond/docs/patterns/frameworks/react/) and [lingui.js](https://lingui.js.org/) use `{}` chars to mark variables.
When you try to add FilePonf label for translations you get an error because lingui expect you will pass a value of the variable, but it actually FilePond's variable.

**To Reproduce**
Steps to reproduce the behavior, possibly with minimal code sample, e.g:

```javascript
import { t } from '@lingui/macro';
import { I18n } from '@lingui/react';
import { FilePond } from 'react-filepond';

export default function App() {
   return (<I18n>
        {({ i18n }) => (<FilePond
                  …
                  labelMaxFileSize={i18n._(t`Maximum file size is {filesize}`)} />))}
        </I18n>);
}
```

You will get `TypeError: Cannot read property 'filesize' of undefined` — it came from `/node_modules/@lingui/core/cjs/core.development.js:134`. I suppose that lingui.js think that filesize is lingui.js variable and expect filesize value should be passed. But it's should be passed by FilePhond.

**Expected behavior**
It would be nice to escape `{}` somehow.
I have tried / and // but I either get an error, either it just doesn't translate it. 

**Additional context**

- jsLingui version ``lingui --version`` **1.4.5**
- Babel version ``npm list babel-core``

```
$ npm list babel-core
instapro-client@1.0.0 /Users/silentimp/Work/instapro-client
└─┬ next@7.0.3
  └── babel-core@7.0.0-bridge.0 
```

- Your Babel config (e.g. `.babelrc`) or framework you use (Create React App, Meteor, etc.)
```json
{
  "presets": [
    [
      "@babel/preset-env",
      {
        "modules": false,
        "debug": false,
        "useBuiltIns": "usage",
        "targets": {
          "browsers": [">0.2%", "not dead", "not ie < 11", "not op_mini all"]
        }
      }
    ],
    "next/babel"
  ],
  "plugins": ["macros"]
}
```

tricoder42:
LinguiJS uses ICU MessageFormat. Could you please try [escape argument](http://userguide.icu-project.org/formatparse/messages#TOC-Quoting-Escaping using apostrophes?

```
Maximum file size is '{filesize}'
```

SilentImp:
@tricoder42  I have tried to do it this way:

```javascript
import { t } from '@lingui/macro';
[...]

PR #520, opened 2019-06-12
test: Add tests to check for escape apostrophes

hjylewis:
Syntax escaping is supported by ICU MessageFormat by using apostrophes http://userguide.icu-project.org/formatparse/messages#TOC-Quoting-Escaping

The bug that caused #457 has already been fixed in the `next` branch.

This PR adds some tests to prevent regressions.


tricoder42:
Looks great, thank you!

Issue #489, opened 2019-04-17
Mozilla's Fluent Message Format Support


chrischen:
**Is your feature request related to a problem? Please describe.**
Mozilla's new Fluent format is similar to MessageFormat from ICU but supports asymmetric localization (where the source and target language don't have a different number of possible translations depending on a variable or number or context).

```
tabs-close-warning-multiple = {$count ->
    [few] Chystáte se zavřít {$count} panely. Opravdu chcete pokračovat?
   *[other] Chystáte se zavřít {$count} panelů. Opravdu chcete pokračovat?
}
```

**Describe the solution you'd like**
Consider the support of Fluent formatted translation files: https://hacks.mozilla.org/2019/04/fluent-1-0-a-localization-system-for-natural-sounding-translations/

**Describe alternatives you've considered**

**Additional context**

tricoder42:
I'm in favor of supporting Fluent, but I don't have capacity to work on it.

- `@lingui/macro` was refactored, that the code is parsed into tokens and message is generated from these tokens. Right now the only implemented message format is [ICU](https://github.com/lingui/js-lingui/blob/next/packages/macro/src/icu.ts). You could implement Fluent in the same way.

- `@lingui/conf` probably needs another option, `format`:

```js
{
  parser: "icu" | "fluent"
}
```

- `@lingui/cli` and `@lingui/loader` - once we have configuration in `conf`, we could compile messages using specific parser

Bottom line: Packages are ready for another message format. We just need somebody to put the effort into it. Interested in sending a PR? 😉

nodkz:
I'm working on FLUENT integration right now.

For now implemented `format: fluent` for `@lingui/cli` and `@lingui/loader`: https://github.com/nodkz/js-lingui/commit/f38a9af95bc975d0f209eba6c1c818058c69905b

It remains to implement new `parser` option.

PS. @tricoder42 please review https://github.com/lingui/js-lingui/pull/530 I don't know why tests fail cause it perfectly works locally. This PR 530 is required for proper testing of `fluent` formatter and implemented on top of this changes. Tnx.

nodkz:
Ouch, `parser` implemented in the `next` branch.
Ok, I'll port my changes to this branch.

@tricoder42 you may close #530 it not required anymore.

alesn:
This is still relevant ...

tricoder42:
@alesn Are you going to work on this?

tricoder42:
[...]

PR #531, opened 2019-06-27
[next] Add support for FLUENT format

nodkz:
According to https://github.com/lingui/js-lingui/issues/489#issuecomment-497781933 needs to implement:
- [x] format in `cli`
- [x] processToken in `macro`
- [ ] add config property `parser`
- [ ] check compile
- [ ] polishing & refactoring

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -47,6 +47,7 @@
     "cli-table": "^0.3.1",
     "commander": "^2.20.0",
     "date-fns": "^1.29.0",
+    "fluent-syntax": "^0.13.0",
     "fuzzaldrin": "^2.1.0",
     "glob": "^7.1.4",
     "inquirer": "^6.3.1",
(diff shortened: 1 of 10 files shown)


nodkz:
@tricoder42 
- need some help with introducing `parser` option to config
- need some info on how to compile phase works

tricoder42:
@nodkz How can I help?

nodkz:
@tricoder42 please add `parser` option as you noted here https://github.com/lingui/js-lingui/issues/489#issuecomment-497781933 

How I saw it will require a small code refactoring. And will be better if you make it.
For me it quite difficult to fix all places where new `parser` option will be used. How I understood it should be applied to babel-extractor and cli-compiler.

I added [macro/fluent.ts](https://github.com/lingui/js-lingui/pull/531/files#diff-69886d92ee256176ca66cf8809a99389) but it is no way to use in the code yet.

tricoder42:
I'm on it. What if we name it `messageFormat` instead of `parser`? While `parser` is technically valid name, I find `messageFormat` more descriptive.

nodkz:
@tricoder42 I'm agree. messageFormat is the better name

tricoder42:
Hey @nodkz, this PR seems to be stale and new major release has just been published. If you'd like to finish it, please rebase it against `main` branch and reopen. Thanks for your efforts! 👍

PR #552, opened 2019-08-16
Documentation - Add "LocalEthereum" to list of projects using LinguiJS

localethereumMichael:
We've been using LinguIJS with LocalEthereum for some time now to support 12 languages.

Thank you for the great project.


tricoder42:
Looks great, thanks!

PR #545, opened 2019-07-20
Raise an error when translation is missing

landsman:
change for edge case in issue: #544, for better debugging

PR #554, opened 2019-08-19
fix(compile): fallback gracefully when source translation is missing

zenyr:
- lingui version: `2.8.3`
# Goal of this PR
- This PR fixes `lingui compile` failing whan a locale has a string that the source locale doesn't.

```
(ko/messages.po, downloaded from translation platform)
...
msgid "Hello"
msgstr "안녕"
...
```
 
```
(en/messages.po, extracted from latest source)
...
#~ msgid "Hello" (now removed)
...
``` 

# Expected
- `lingui compile` should compile both `ko`, `en` without problem
# Actual
- `lingui compile` errors with following error:
```
/Users/.../node_modules/@lingui/cli/api/catalog.js:131
        return catalogs[locale][key].translation;
                                     ^

TypeError: Cannot read property 'translation' of undefined
    at getTranslation (/Users/.../node_modules/@lingui/cli/api/catalog.js:131:38)
```

### DIFF
--- a/packages/cli/src/api/catalog.js
+++ b/packages/cli/src/api/catalog.js
@@ -104,7 +104,10 @@ export default (config: LinguiConfig): CatalogApi => {
     },
 
     getTranslation(catalogs, locale, key, { fallbackLocale, sourceLocale }) {
-      const getTranslation = locale => catalogs[locale][key].translation
+      const getTranslation = anotherLocale =>
+        (catalogs[anotherLocale][key] &&
+          catalogs[anotherLocale][key].translation) ||
+        undefined
 
       return (
         // Get translation in target locale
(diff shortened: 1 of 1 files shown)


zenyr:
After searching through PRs, I found that #545 tries to fail on such case while this PR tries to recover gracefully.

tricoder42:
Hi @zenyr, good observation. Let's figure out the API.

`lingui extract` uses silenty empty string as a last resort, unless you provide `--strict` option.

`lingui compile` should probably do the same - fallback to empty string silently, unless you provide `--strict` option.

What do you think? @landsman, since this is related to #545 as well, what's your opinion on this?

landsman:
@tricoder42 @zenyr I see that we are struggling on same thing.
In my eye there would be good have there console.log by default, without `--strict` option. 
Save time on debugging.

zenyr:
I agree, `--strict` would fail compile while we could warn users and continue by default. Silent recovery could be too "magical" for users indeed.

tricoder42:
I believe this was already in some form resolved in #545. If not, plese reopen this PR agains `next` branch.

Thank you for your effort and sorry for such a long delay!

PR #563, opened 2019-09-06
Add optional sort by origin/filename

skovhus:
This is a proposal to add a config option `sorting` to expose a new sort by origin feature.

Unfortunately, `master` is in a rather bad state with a bunch of failing tests. Let me know what I can do to help.

Context https://github.com/lingui/js-lingui/issues/562

PR #564, opened 2019-09-08
Fix unit tests

skovhus:
First, thanks for a great library! 👏 

As a new contributor, it would be great if all test passed when cloning the project. In order to do so, I suggest blocking any merges to master if CI is failing. This can be configured in the repository settings by enabling branch protection on master... When looking over the git commits, most of them actually also failed on CI, so it can be hard to know when a bug was introduced.

This PR tries to stabilize CI and fixes bugs introduced in https://github.com/lingui/js-lingui/pull/518 (where CI actually showed a bug)

## TODO
- [x] Fix test error when running locally
- [ ] Fix CI timing out (`node scripts/test.js --skip-build`)
- [ ] Fix CI build errors
- [ ] Disable or fix now deployment? (I see this was introduced in the 3.x branch)

## TODO Maintainers 
- [ ] Update project setting to not allow merges to master when CI fails


tricoder42:
Hey @skovhus, thanks for taking the initiative!

As I mentioned in #563:
- `lingui-loader › should compile catalog` is tricky, because it works locally and only sometimes fails in CI. This error also appears when the loader doesn't finish compilation or if there's an error inside loader.
- `compile › should handle an error if message has syntax errors` - From what I see in the CI report, the strings are equal even jest reports they're not. I'll try it locally.

> Disable or fix now deployment? (I see this was introduced in the 3.x branch)

Yeah, I added this for v3, but it's just documentation. Shouldn't be any harm.

> Update project setting to not allow merges to master when CI fails

Sure, once we figure out all problems.

Thank you again, I'll take a look too 👍

skovhus:
Now that I understand that all development is going on in the next branch, I think we can close this (which changes master).

tricoder42:
Hey @skovhus! After recents changes in `next` I'm having a green build when running locally 💚 

The `erorrs` typo was already fixed in other PR, so I'm gonna just add jest timeout to `next` branch and close this PR. That should fix the CI timeout errors.

Now fails only on master, but I want to focus on `next`.

> Update project setting to not allow merges to master when CI fails

Let's postpone it until `next` is merged into master and the proejct is a bit more stable.

Thanks for working on it!

tricoder42:
Hey @skovhus! After recents changes in `next` I'm having a green build when running locally 💚 

Now fails only on master, but I want to focus on `next`.

> Update project setting to not allow merges to master when CI fails

Let's postpone it until `next` is merged into master and the proejct is a bit more stable.

Thanks for working on it!

Issue #562, opened 2019-09-06
Add optional sort by origin/filename


skovhus:
**Is your feature request related to a problem? Please describe.**
Currently, lingui sorts by message key. This might make sense in some project, but sorting by origin can often be more natural, e.g. translations from the same file grouped in the .po files.

Note that [GNU gettext](https://www.gnu.org/software/gettext/manual/gettext.html) there is an option called `--sort-by-file`.

I saw that this is also mentioned in the [[Epic] Road to LinguiJS v3](https://github.com/lingui/js-lingui/issues/334).

**Describe the solution you'd like**

Changes in `@lingui/cli`:
- preserve version 2 backward compatibility
- add a configuration option `sorting` with values `key` (default) or `origin`

**Additional context**
I have a proof of concept for this, that I'll make a PR with. @tricoder42 let me know if you have any comments.

skovhus:
When cloning `js-lingui` I get 12 test failures on `master`... Would be nice to have thosed fixed before doing any changes. Any advice? 

```
Summary of all failing tests
 FAIL  packages/core/src/i18n.test.js (7.156s)
  ● I18n › .use should return new i18n object with switched locales

    expect(received).toEqual(expected)

    Expected value to equal:
      "٢ الكتب"
    Received:
      "2 الكتب"

      171 |         other: "# الكتب"
      172 |       })
    > 173 |     ).toEqual("٢ الكتب")
          |       ^
      174 | 
      175 |     const uae = i18n.use("ar", "en-UK")
      176 |     expect(

      at Object.toEqual (packages/core/src/i18n.test.js:173:7)

 FAIL  packages/react/src/Select.test.js
  ● Categories › Plural › should use plural forms based on locales

    expect(received).toEqual(expected)

    Expected value to equal:
      "1.000"
    Received:
      "1,000"

      84 |       const t = () => node.find("Render").text()
      85 | 
    > 86 |       expect(t()).toEqual("1.000")
         |                   ^
      87 | 
      88 |       node.setProps({ value: 2500 })
      89 |       expect(t()).toEqual("2.500")

      at Object.toEqual (packages/react/src/Select.test.js:86:19)

  ● Categories › Plural › should use plural forms based on locales set on i18n

    expect(received).toEqual(expected)

    Expected value to equal:
      "1.000"
    Received:
      "1,000"

       98 |       const t = () => node.find("Render").text()
       99 | 
    > 100 |       expect(t()).toEqual("1.000")
          |                   ^
[...]

PR #565, opened 2019-09-09
Add orderBy option (origin or messageId)

skovhus:
This is a proposal to add a config option `orderBy` to expose a new order by origin feature. 

Context #562


tricoder42:
Ups, I got a little bit confused by the second PR. What's the difference between this and #563? I think I merged the wrong one...

skovhus:
This is for version 3, the Next branch. The other one was for the version 2, master.

tricoder42:
Merde, I've just merged `master` into `next` and ported sorting feature before I realized there's another PR for it :) Anyway, `orderBy` is a better name than `sorting`, so I'm gonna keep this version.

Thank you for the hard work! 🥇

PR #571, opened 2019-10-02
don't write unchanged extract files

gabormagyari:
Hi,

I want to create a pre-push hook to prevent untranslated strings by running `yarn extract`, then check if there is any uncommited changes (e. g. locale files). I'm using this command:

`yarn extract && git diff-index HEAD`

The problem is that extract always writes the locale files so they show up in `git diff` if I'm running the command above. This PR changes this behaviour and solves my problem.


tricoder42:
Hey @gabormagyari, sorry for delay. I've just rebased it against latest `next` branch and merge it.

Thank you for your contribution 🎉

PR #572, opened 2019-10-02
fix typo in test

gabormagyari:
Fixes a typo that makes the test fail.


tricoder42:
Hey @gabormagyari, thank you for PR! Interesting it's only in test and not in the code... I ported the fix to next branch as well in 6b71c69 where was typo even in the code. 🤷‍♂

PR #576, opened 2019-10-12
Remove functionality related to loading/activating locales

tricoder42:
# Motivation

Setting locale is a one-time action either done automatically or manually by user when automatic detection fails. It's rarely required to switch locale afterwards. Therefore it's not a problem to reload page completely on locale change. Considerable amount of code can be saved by removing code related to message loading and locale activating.

# Checklist

- [x] Remove functionality related to loading/activating locales
- [x] Update Node.js example
- [ ] Add function to detect and override user's locale
- [ ] Update React example


tricoder42:
Deadend ☠️

PR #585, opened 2019-11-08
Fix issues with Flow type checking

d4wei:
Currently, `Flow` isn't actually checking any type imported from `core` due to its root-level `index.js` file missing a `@flow` header. Fixing this revealed a few more errors which I also included fixes for in this branch.

Not all tests are passing in this branch; however all failing tests also fail on `master`, so I'm just ignoring them for now.

### DIFF
--- a/packages/core/index.js
+++ b/packages/core/index.js
@@ -1,1 +1,2 @@
+// @flow
 export * from "./src"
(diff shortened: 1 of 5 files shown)


tricoder42:
Looks good to me! Thank you!

PR #586, opened 2019-11-11
fix: compile strict skips pseudo locale

### LINKED ISSUE #584: `compile --strict` should skip missing pseudo translations
**Describe the bug**
`lingui compile --strict` shouldn't fail with pseudo language missing translations

**To Reproduce**
Having pseudo locale enabled in config `pseudoLocale: 'pseudo'` run `lingui compile --strict` to compile messages catalogs. The command will fail even when all translations are available because pseudo lang is taken into account when checking for missing translations

**Expected behavior**
the `--strict` flag should compile succesfully when all translations are available even when I use pseudo locales for testing

**Additional context**
- jsLingui version ``lingui --version`` 2.8.3
- Babel version ``npm list babel-core`` 7.0.0-bridge.0
- Your Babel config (e.g. `.babelrc`) or framework you use (Create React App, Meteor, etc.) custom config

danielchabr:
closes #584

### DIFF
--- a/packages/cli/src/lingui-compile.js
+++ b/packages/cli/src/lingui-compile.js
@@ -65,7 +65,11 @@ function command(config, options) {
       }))
     )
 
-    if (!options.allowEmpty && config.sourceLocale !== locale) {
+    if (
+      !options.allowEmpty &&
+      config.sourceLocale !== locale &&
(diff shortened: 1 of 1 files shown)


tricoder42:
Looks good, thank you!

Issue #593, opened 2019-11-20
formatElements fails with multiple uses of same tag in same scope


dannysellers:
**Bug Description**
When formatting a string with multiple usages of the same component in the same scope, `formatElements` only considers the outermost opening and closing tags

**To Reproduce**
Use a Trans tag like `<Trans id="<0>Superscript</0> regular <0>super</0>" components=[{<sup />}] />`, observe rendered HTML is `<sup>Superscript&lt;/0&gt; regular &lt;0&gt;super</sup>`

```javascript
it("should format multiple usages of same element in the same scope", function() {
  expect(
    html(formatElements("<0>First</0> second <0>third</0> <0>fourth</0>", [<sup />]))
  ).toEqual(
   wrap("<sup>First</sup> second <sup>third</sup> <sup>fourth</sup>")
  )
})
```

**Expected behavior**
The formatter should correctly replace multiple instances of the same component

- jsLingui version `2.8.3`
- Babel version `6.26.3`
- This issue affects a project created with [simple-react-app](https://github.com/Kornil/simple-react-app), but also affects Lingui's `master`

tricoder42:
Fixed in #594

tricoder42:
It has just been released in [2.9.0](https://github.com/lingui/js-lingui/releases/tag/v2.9.0).

PR #594, opened 2019-11-20
#593: lazily split messages by value tags

dannysellers:
#593 Adds support for reusing the same tag multiple times in the same line in the same scope.
Given a message: `"msg": "<0>first</0> second <0>third</0>"`, used with `<Trans id="msg" components=[{<sup />}] />`
We expect: `<sup>first</sup> second <sup>third</sup>`

In practice we get: `<sup>first&lt;/0&gt; second &lt;0&gt;third</sup>`

Current: https://regex101.com/r/LHoVEP/2
New: https://regex101.com/r/LHoVEP/1

### DIFF
--- a/packages/react/src/format.js
+++ b/packages/react/src/format.js
@@ -2,7 +2,7 @@
 import { cloneElement } from "react"
 
 // match <0>paired</0> and <1/> unpaired tags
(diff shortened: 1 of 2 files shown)


tricoder42:
Interesting! It seems to work. Well done 👍

Issue #592, opened 2019-11-20
<Trans> component's render method returns a string, which causes React to fail


toolness:
I am having problems using a relatively trivial integration of lingui with a Create React App (CRA) application.  The project is open-source and the PR I'm using to test out lingui is at https://github.com/JustFixNYC/who-owns-what/pull/184.  (This PR contains information about what versions of CRA, Babel, and lingui I'm using.)

The problem was that I had set up an `<I18nProvider>` with a very simple `<Trans>Who owns what in NYC?</Trans>` component as a descendant, and adding the latter to my component tree caused the initial render to fail with the following exception:

```
Uncaught Invariant Violation: Render.render(): A valid React element (or null) must be returned. You may have returned undefined, an array or some other invalid object.
```

After setting a devtools breakpoint in the `render()` method of the `Render` component in  `node_modules/@lingui/react/cjs/react.development.js`, I found that the method is ultimately just returning a raw string, which is what React is upset about:

https://github.com/lingui/js-lingui/blob/fef70fc591d289f57be37fddca4bb6cc6f04dc64/packages/react/src/Render.js#L21-L26

It's even specified in the Flow typings for the `<Render>` component that `props.value` can be a string, so the fact that it _is_ a string isn't unexpected:

https://github.com/lingui/js-lingui/blob/fef70fc591d289f57be37fddca4bb6cc6f04dc64/packages/react/src/Render.js#L10-L12

However, it seems **render() methods can't return strings**!  For example, take the following trivial functional React component:

```
function MyBrokenComponent(props) {
  return "boop";
}
```

Using the above component will fail with the exact same Uncaught Invariant Violation described above.

After reading the source code for `<Render>` I noticed that it tries using a default render method provided by the `<I18nProvider>`, so [I defined that](https://github.com/JustFixNYC/who-owns-what/pull/184/commits/640653c3f4633a0edf50980cc947f3ac0ebc7359) and it seemed to fix things--although I really wanted to return a React fragment from it so that it wouldn't be wrapped in an extra DOM element, I ran into problems with this, so I ended up just wrapping the translated text in a `<span>`.

Am I somehow using lingui the wrong way?  Or is this a legitimate bug?

toolness:
[...]

PR #597, opened 2019-11-25
Drop support for React 15.

toolness:
Some components in `@lingui/react` return strings from their `render()` methods, which is a convention introduced in React 16 (see the first entry under "New Features" in the [v16.0.0 release notes](https://github.com/facebook/react/releases/tag/v16.0.0)).

This just updates the `peerDependencies` entry of `package.json` to clarify this fact. (See #592 for more details.)

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -37,7 +37,7 @@
     "cjs/"
   ],
   "peerDependencies": {
-    "react": "^15.0.0 || ^16.0.0"
+    "react": "^16.0.0"
   },
   "dependencies": {
     "@lingui/core": "0.0.0-managed-by-release-script",


toolness:
Umm, I am not sure what the ZEIT Now bot is complaining about.  It doesn't _seem_ to be related to anything I've changed in this PR...

tricoder42:
@toolness Yeah, ignore Now bot, it only builds documentation on `next` branch.

Anyway, I would rather remove the peerDependency in 3.x, when we introduce new context API which actually requires a new React. Right now it's possible to use the library with React 15.x when you configure `defaultRender` correctly, so maybe we should rather document it?

toolness:
Ah ok, that's doable too... I was thinking it might be hard to put that in a place where everyone on 15 would see it, but I just noticed that it's possible to [detect the version of React at runtime](https://stackoverflow.com/questions/36994564/how-can-one-tell-the-version-of-react-running-at-runtime-in-the-browser), so what if, in addition to documenting it, we added a `if (process.env.NODE_ENV !== 'production')`-gated conditional in the `render()` method of the `Render` component that throws a helpful error (or at least logs a helpful warning) if it's about to return a string on React 15?  The message could include a pointer to the docs too, similar to how React's own helpful error messages works. Anyways, just an idea!

toolness:
Er, yet another option would be to, in `render()`, simply check if the react version was 15 and if so just wrap the returned string in a `<span>` so that no exception is ever thrown.  But that would also be a funky variation in behavior that might confuse people, too...

tricoder42:
Warning under `if (process.env.NODE_ENV !== 'production')` would be the best I guess 👍

tricoder42:
Hey @toolness, thank you for this PR. The `next` branch (upcoming v3) already has React 16.8 as a minimal supported version, so I'm going to close this PR.

Cheers 👍

PR #600, opened 2019-12-03
Log a helpful warning if we're returning a string from render() in React 15

toolness:
This is an alternative to #597 that, instead of dropping support for React 15, logs a helpful warning if `render()` is about to return a string in React 15, to make it easier for developers to debug #592.

@tricoder42 does this kind of fix seem OK?  It doesn't actually update the docs yet, but I wanted to get your feedback on the code change and the text of the warning.

## To do

- [ ] Add docs that mention the workaround for React 15.

### DIFF
--- a/packages/react/src/Render.js
+++ b/packages/react/src/Render.js
@@ -23,6 +23,17 @@ export default class Render extends React.Component<RenderComponentProps> {
     let render = this.props.render || this.context.linguiDefaultRender
 
     if (render === null || render === undefined) {
+      if (process.env.NODE_ENV === "production" && typeof value === "string") {
+        if (/^15\./.test(React.version)) {
+          console.warn(
+            "lingui is about to return a string from render() in React 15.x, " +
+              "which will likely raise an Uncaught Invariant Violation error. " +
+              "to resolve this, please provide a custom defaultRender function " +
+              "that wraps the string in a component of your choice, or upgrade " +
+              "to React 16."
+          )
+        }
+      }
       return value || null
     } else if (typeof render === "string") {
       // Built-in element: h1, p
--- a/packages/react/src/Render.js
+++ b/packages/react/src/Render.js
@@ -23,7 +23,7 @@ export default class Render extends React.Component<RenderComponentProps> {
     let render = this.props.render || this.context.linguiDefaultRender
 
     if (render === null || render === undefined) {
-      if (process.env.NODE_ENV === "production" && typeof value === "string") {
+      if (process.env.NODE_ENV !== "production" && typeof value === "string") {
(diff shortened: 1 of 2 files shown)


packages/react/src/Render.js
```diff
@@ -23,6 +23,17 @@ export default class Render extends React.Component<RenderComponentProps> {
     let render = this.props.render || this.context.linguiDefaultRender
 
     if (render === null || render === undefined) {
+      if (process.env.NODE_ENV !== "production" && typeof value === "string") {
+        if (/^15\./.test(React.version)) {
```
toolness: I wanted to make this just one `if` statement but eslint got mad and I have to figure out how to run prettier to calm it down.

tricoder42:
Looks good to me 👍  Thanks for the work!

toolness:
Awesome, thanks!

So I'm looking at the docs and I'm not actually sure where the best place to put a blurb about React 15 is, because in our React tutorial, we have developers put a `<Trans>` component in before they put an `<I18nProvider>`.  That makes it harder to advise them to pass a `defaultRender` prop to `<I18nProvider>`, if we haven't yet introduced that component yet...

scripts/build/modules.js
```diff
@@ -9,6 +9,7 @@ const { UNIVERSAL } = require("./bundles").bundleTypes
 const HAS_NO_SIDE_EFFECTS_ON_IMPORT = false
 
 const importSideEffects = Object.freeze({
+  "react": HAS_NO_SIDE_EFFECTS_ON_IMPORT,
```
toolness: I had to do this in order to get CircleCI to pass; otherwise it was [failing](https://circleci.com/gh/lingui/js-lingui/4340) with this message:

```
Error: An external module "react" is used in a DEV-only code path but we do not know if it is safe to omit an unused require() to it in production. Please add it to the `importSideEffects` list in `scripts/rollup/modules.js`.
 ```

I don't _think_ importing `react` has any side effects on import, so I _think_ this should be OK?  Though honestly I was still confused because it seemed like the `react` module should always be imported regardless of whether we're in production or development, since we need to call `React.createElement()` to process any of the JSX, but maybe JSX is processed later in the pipeline. 🤷‍♂

toolness:
Er, any idea why test-node-8 and test-node-latest are [failing](https://circleci.com/gh/lingui/js-lingui/4344?utm_campaign=workflow-failed&utm_medium=email&utm_source=notification)?  It's giving a bunch of errors that might be unrelated but I'm not sure...

tricoder42:
Hey @toolness, thanks for the PR but I have to drop React 15 anyway, because Lingui uses new context and hooks. I appreciate your support 🙏

PR #603, opened 2019-12-16
Add `babel-plugin-macros` to default TypeScript plugins to make message extraction much easier to setup for non-CRA projects

MartijnHols:
Add `babel-plugin-macros` to plugins since it's required for any project type. The plugin is often included by default by presets such as `react-app` and `tsdx`, but since only `react-app` is supported by default, extracting in other libraries (in my case TSDX) is much more time consuming to configure.

### DIFF
--- a/packages/cli/src/api/extractors/typescript.js
+++ b/packages/cli/src/api/extractors/typescript.js
@@ -47,6 +47,7 @@ const extractor: ExtractorType = {
       linguiTransformJs,
       linguiTransformReact,
       [linguiExtractMessages, { localeDir }],
+      "babel-plugin-macros",
       ...(babelOptions.plugins || [])
     ]
 


FredyC:
It was added as a dependency in the past, but it was removed. Why do you consider adding one extra dependency as "time-consuming"?

MartijnHols:
It has been a while since I ran into this, but I guess it must have been hard to discover the cause for failed extraction and that the package was missing and required. 

I haven't run into this issue since I solved it for that project and it probably won't bother me again, so you're welcome to close this without merging.

tricoder42:
Thanks @MartijnHols. I'll close this PR because typescript extractor will be moved to external package.

Thank you for your work and sorry for late response.

PR #605, opened 2019-12-20
Fix flow types on withI18n

ktunkiewicz:
Hello guys,

I don't know how that works for others, but the existing flow types in withI18n throw a lot of errors in our project.
I figured out that the proposed change fixes this.

As far as I understand flowtype and the withI18n component, the return type is indeed `React.ComponentType<$Diff<P, withI18nProps>>` and not an intersection of `C` and the above.

Please take a look if it makes sense for you too.

### DIFF
--- a/packages/react/src/withI18n.js
+++ b/packages/react/src/withI18n.js
@@ -19,7 +19,7 @@ type withI18nProps = {
 const withI18n = (options: withI18nOptions = {}) =>
   function<P, C: React.ComponentType<P>>(
     WrappedComponent: C
-  ): C & React.ComponentType<$Diff<P, withI18nProps>> {
+  ): React.ComponentType<$Diff<P, withI18nProps>> {
     if (process.env.NODE_ENV !== "production") {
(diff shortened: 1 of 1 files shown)


ktunkiewicz:
duh.. ok Mr "now" bot, I admit, I didn't read any readme about contributing :)
I just used GitHub's "edit" function. Should I do something more to make this PR pass automated tests @tricoder42 ?

tricoder42:
Thanks, this looks good!

tricoder42:
Fixed in [v2.9.1.](https://github.com/lingui/js-lingui/releases/tag/v2.9.1) 🎉

tricoder42:
Unfortunatelly, these changes caused few flow errors. I had to disable flow linter in `master`, because I'm planning to remove flow completely in v3.

If you were interested fixing it, I've created branch [`feature/fix-flow`](https://github.com/lingui/js-lingui/tree/feature/fix-flow) with updated flow definitions.

Issue #599, opened 2019-12-02
[2.9.0] - Crash pipeline cannot find module 'typescript'


semoal:
**Describe the bug**
Crash on Node projects without typescript

**To Reproduce**
Just use 2.9.0 without typescript installed as peer dependency.

```
import { Trans } from "@lingui/react"

export default function App() {
   return <Trans>This should be translated!</Trans>
}
```

**Expected behavior**
Not crash, should be built or something like that.

- jsLingui version: 2.9.0

What's the reason to put typescript as peer dependecy, when a lot of projects don't need typescript for anything.
Shouldn't be built before publish the package? In plain Javascript will work for all the projects.

tricoder42:
Hey @semoal, what's the error message please? Could you please the code required to reproduce this error to a repository for quicker debugging?

@FredyC: My guess is that cli loads all modules, including Typescript transformer, and that breaks for projects which doesn't use typescript at all? Before these project get typescript installed without notice.

semoal:
> Hey @semoal, what's the error message please? Could you please the code required to reproduce this error to a repository for quicker debugging?
> 
> @FredyC: My guess is that cli loads all modules, including Typescript transformer, and that breaks for projects which doesn't use typescript at all? Before these project get typescript installed without notice.

Hi, I can't provide all my code, but it's just a react app with a custom webpack and babel. (not using create-react-app that probably provides typescript under the hood)

Here is the crash:

```
12:58:14  > npm run extract && npm run compile
12:58:14  
12:58:15  
12:58:15  > xxxx-xxxx-front@1.0.0 extract /app
12:58:15  > lingui extract
12:58:15  
12:58:15  [91minternal/modules/cjs/loader.js:638
12:58:15      throw err;

12:58:15  Error: Cannot find module 'typescript'
12:58:15      at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
12:58:15      at Function.Module._load (internal/modules/cjs/loader.js:562:25)
12:58:15      at Module.require (internal/modules/cjs/loader.js:692:17)
12:58:15      at require (internal/modules/cjs/helpers.js:25:18)
12:58:15      at Object.<anonymous> (/app/node_modules/@lingui/cli/api/extractors/typescript.js:33:19)
12:58:15      at Module._compile (internal/modules/cjs/loader.js:778:30)
12:58:15      at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
[...]

PR #611, opened 2020-01-17
Use lazy require to import TypeScript

theKashey:
Should solve the problem when there is no TS used, but `TypeScript` is required - #599


theKashey:
3 lines of code, as promised 😎

theKashey:
Tests are failing due to timeout 🤷‍♂️

tricoder42:
Looks good! Thank you very much!

PR #613, opened 2020-01-29
Is locales better?

banyan:
I was following this [document](https://lingui.js.org/tutorials/setup-cra.html) and `localeDir` is written as `locales`. 
so is `locales` better than `locale`?


tricoder42:
It isn't worse or better. Just a matter of choice. I guess we could make it consistent in the docs. Thanks for the PR!

PR #615, opened 2020-02-03
Expand and clarify "lazy translation" patterns

agwells:
I've introduced several people to Lingui, and I find they often have the same questions about what the docs describe as "lazy translations". Which I take to mean, use-cases where the default translation can _not_ be placed inline in a `<Trans>` or `i18n._()` in the code location where the message will be displayed.

So, I'm submitting some changes for this section of the docs. This includes:

* Changing the example to use the names of colors instead of languages. (I find that newcomers are often confused by the language names, into thinking the example has something to do with selecting a translation language catalog)
* Referring more explicitly to the "message descriptor" objects created by the `t` macro, so that it's more obvious what is being put into variables and passed around.
* Added examples about passing lazy translation messages as component props
* Added an example of picking between messages based on the value of a variable

Thanks for the great library! I'm looking forward to version 3. :)

### DIFF
--- a/docs/tutorials/react-patterns.rst
+++ b/docs/tutorials/react-patterns.rst
@@ -198,39 +198,129 @@ a bit more setup:
 Lazy translations
 =================
 
-:jsxmacro:`Trans` can also translate messages from variables. We can use :jsmacro:`t`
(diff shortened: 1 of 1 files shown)


tricoder42:
Looks good, thank you!

Issue #368, opened 2018-10-24
Object spread in <Trans> causes an error in "lingui extract"


agwells:
**Describe the bug**
Let me preface this by saying, this is a fairly small bug, with a few different possible workarounds. But it looks like it may be an easy fix.

I was trying to do [lazy translation](https://lingui.js.org/tutorials/react-patterns.html#lazy-translations), but I wanted to support not just static strings, but also strings with placeholders (and potentially also plurals). So, instead of using `i18nMark()` (which only accepts static strings), I used the `t` macro to tag the language strings, and passed the resulting object in to the component that would render it. Because the string may have different sorts of fields depending on whether it has a placeholder, plural, etc, I used object spread syntax to turn them all into props: `<Trans {...myLangString} />`.

This evaluates and runs fine in dev, but it causes `lingui extract` to crash, with the error message:

```
node_modules/@lingui/babel-plugin-extract-messages/index.js:110
          var key = item.name.name;
```

**To Reproduce**

Here's a (very contrived) minimal example of the problem. Save the following file, and run `lingui extract` over it.

```javascript
import React from 'react';
import { Trans } from '@lingui/react';
import { t } from '@lingui/macro';

const adminName = 'Aaron';
const myTranslatedString = t`Please lazy-translate this, ${adminName}`;

export default function App() {
  return (
    <div>
      <Trans {...myTranslatedString} />
    </div>
  );
}
```

**Expected behavior**

`lingui extract` should not crash when it encounters this syntax. It should either silently ignore it, or possibly print a warning message like it does when it encounters a `<Trans>` tag that has no `id` prop.

**Additional context**

Fortunately there are several workarounds to this problem:

1. If your lazy-translated strings don't contain HTML, use `i18n._()` instead of `<Trans>`.
2. If your lazy-translated strings are just static string literals without placeholders or plurals, use `i18nMark()` instead of `t` or `plural()`.
3. Manually enumerate all the expected props from your lazy-translated strings: `<Trans id={myString.id} defaults={myString.defaults} values={myString.values} />`
4. Copy `<Trans>` to a local variable with a different name, and use that instead: `const NoExtractTrans = Trans; return <NoExtractTrans {...myString} />`
5. Add the file to your `srcPathIgnorePatterns` lingui config setting.

- jsLingui version: 2.7.0
- Babel version: 
```
└─┬ react-scripts@2.0.4
  ├── babel-core@7.0.0-bridge.0 
  └─┬ jest@23.6.0
[...]

PR #617, opened 2020-02-09
Correcting bugs in "lazy translations" sample code

agwells:
I've noticed some problems in the updates I pushed for the "lazy translations" section of the docs. :sweat_smile: 

The main thing is, it copied from the previous examples the idea that you could pass a message descriptor object from `t` to `<Trans>`'s `id` prop. That doesn't work. In fact, it causes various errors, because the `id` prop must be a string, but `t` returns an object. I suspect it's probably a holdover from an earlier version of the docs that referenced the `i18nMark()` function, which actually did return a string.

Here's a little example to demonstrate what I'm talking about

```jsx
import React from 'react';
import {Trans, t} from "@lingui/macro";

const msg = t`Hello world`;

// This is what the examples were saying to do, but it causes a type error
const bad = <Trans id={msg} />;

// This causes "lingui extract" to crash due to #368 
const bad2 = <Trans {...msg} />;

// This is what you must do
const good = <Trans id={msg.id} />;
```

So this merge request updates my examples to pass the message descriptor's ID instead of the whole message descriptor object. Which is also what I've been doing in my own project that uses lazy translations, but I'd forgotten about that when writing the examples! I also actually ran and tested the code in each example, and fixed some other unrelated issues in them.

The new examples are still a little bit simple, because they don't address how you need to also pass the `defaults` property if you're using custom message IDs (and other properties if you're trying to use lazy translations for parameterized messages). But I think that would probably be best addressed by fixing #368 so that `<Trans {...msg} />` will just be ignored by "lingui extract" instead of causing it to crash.

### DIFF
--- a/docs/tutorials/react-patterns.rst
+++ b/docs/tutorials/react-patterns.rst
@@ -201,27 +201,32 @@ Lazy translations
 Messages don't have to be declared at the same code location where they're displayed.
 Tag a string with the :jsmacro:`t` macro, and you've created a "message descriptor", which
 can then be passed around as a variable, and can be displayed as a translated string by
-passing it to :jsxmacro:`Trans` as its ``id`` prop:
+passing its ``id`` attribute to :jsxmacro:`Trans` as its ``id`` prop:
 
 .. code-block:: jsx
 
-   import { t, Trans } from "@lingui/macro"
+   import React from 'react';
+   import { t, Trans } from '@lingui/macro';
 
+   // Creates an array of message descriptor objects
    const favoriteColors = [
       t`Red`,
       t`Orange`,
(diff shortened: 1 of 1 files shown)


tricoder42:
Hey @agwells, sorry I haven't respond earlier. This PR is already obsolete, because in V3 there's a new `defineMessage` macro for lazy translations:

```js
import React from 'react';
import { defineMessage } from "@lingui/macro";

const msg = defineMessage({ message: t`Hello world` });

function Component() {
  const i18n = useLingui()

  return i18n._(msg)
}
```

I'm gonna close this PR as I'm already working on new docs.

Thank you anyway 🙏

PR #620, opened 2020-03-03
Add fs-extra to package.json for CLI which is using it.

anuraaga:
Currently when using `next`, the user has to add `fs-extra` to their package.json since the CLI is using it (in `catalog.ts`), but it's not listed as a dependency.


tricoder42:
Perfect, thanks 👏

Issue #560, opened 2019-09-02
date macro not working when used inside t macro


jbojcic1:
**Describe the bug**
I have a message with a date that I need to translate outside of the React component, so based on https://lingui.js.org/ref/macro.html#jsmacro-t I tried using `i18n._(t`Today is ${date(name)}`)`, but it's not working. I am getting `Danas je [object Object]` in the result.

**To Reproduce**
1. Create react app with Create React App.
2. Add lingui-js and set it up so you can use translations outside the components as described here: https://lingui.js.org/tutorials/react-patterns.html#translations-outside-react-components
3. outside of the React component do:
```
import { date } from '@lingui/macro';
import { i18n } from './i18n';
...
const today = new Date();
const msg = i18n._(t`Today is ${date(today)}`);
```
then render msg in the component:
```
{msg}
```
Sandbox example  - https://codesandbox.io/s/linguijs-playground-5gk1u

**Expected behavior**
Date should be normally localized.

**Additional context**

- jsLingui version `2.8.3`
- I am using React app generated with Create React App

semoal:
> **Describe the bug**
> I have a message with a date that I need to translate outside of the React component, so based on https://lingui.js.org/ref/macro.html#jsmacro-t I tried using `i18n._(t`Today is ${date(name)}`)`, but it's not working. I am getting `Danas je [object Object]` in the result.
> 
> **To Reproduce**
> 
> 1. Create react app with Create React App.
> 2. Add lingui-js and set it up so you can use translations outside the components as described here: https://lingui.js.org/tutorials/react-patterns.html#translations-outside-react-components
> 3. outside of the React component do:
> 
> ```
> import { date } from '@lingui/macro';
> import { i18n } from './i18n';
> ...
> const today = new Date();
> const msg = i18n._(t`Today is ${date(today)}`);
> ```
> 
> then render msg in the component:
> 
> ```
> {msg}
> ```
> 
> Sandbox example - https://codesandbox.io/s/linguijs-playground-5gk1u
> 
> **Expected behavior**
> Date should be normally localized.
> 
> **Additional context**
> 
> * jsLingui version `2.8.3`
> * I am using React app generated with Create React App

Did you find any solution mate? I'm having the same problems.

```jsx
i18n._(number(payload.value, {
        style: "percent",
        maximumFractionDigits: 2,
        minimumFractionDigits: 2,
}))
```

For example, this is causing for me [ObjectObject]

jbojcic1:
[...]

PR #621, opened 2020-03-11
Fix for date and number to return proper values instead of [object Object] 

igorDolzh:
This issue is mentioned here: https://github.com/lingui/js-lingui/issues/560

Briefly
```
i18n._(t`Interest rate is ${number(0.05, "percent")}.`
```
may return 
```
Interest rate is [object Object]
```
instead of expected value 
```
Interest rate is 5%
```

The reason is the order of imports!

If you have imports (1)
```
import { t, number } from "@lingui/macro";
//returns `Interest rate is 5%`
//extracts `Interest rate is {0,number,number0}`
```
everything is working correctly.

But, if you have imports  (2)
```
import { number, t } from "@lingui/macro";
//returns `Interest rate is [object Object]`
//extracts `Interest rate is {0}` 
```
you will receive `[object Object]`

So, the fact is that the order of imports affects the order macros will be processed. In the (1) example, `t` is exported first and when it is executed it deals with with `number` as a function, (when this function will be executed and will return `{0, number,number0}` 
But when `number` is exported first (in (2)), it is executed first (it returns something like `{0, number,number0}` with value {0: value})  and then `t` treats this value as object and as the result, this value is converted to `[object Object]` in the string.

Also, we have the same problem with `date` with `t` and even `select` with `plural`. And I noticed some patterns that macro tags should be processed in the following order as they appear in the code. With this code, you could write your imports in any order and everything should work

### DIFF
--- a/packages/macro/src/index.js
+++ b/packages/macro/src/index.js
@@ -9,9 +9,14 @@ function macro({ references, state, babel }) {
   const jsxTransformer = new JSXTransformer(babel)
   const reactImportsToCarryOver = ["DateFormat", "NumberFormat"]
   const reactImports = []
+  let allTags = []
 
   Object.keys(references).forEach(tagName => {
     const tags = references[tagName]
+
+    if (tags) {
+      allTags = allTags.concat(tags)
+    }
     const macroType = getMacroType(tagName)
 
     if (macroType === "jsx") {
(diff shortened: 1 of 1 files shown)


tricoder42:
Hi! Thanks for this PR! Order of imports doesn't matter in v3. I've just checked.

PR #623, opened 2020-03-13
fix: allow imports from @lingui/macro to be in any order

ckknight:
previously, the references were processed in the order that they were
imported, leading to troubling results if one attempted to import
`plural` before `t` and use them together.

This transforms each reference in a deterministic order, regardless of
the order of import.

### DIFF
--- a/packages/macro/src/index.js
+++ b/packages/macro/src/index.js
@@ -10,8 +10,11 @@ function macro({ references, state, babel }) {
   const reactImportsToCarryOver = ["DateFormat", "NumberFormat"]
   const reactImports = []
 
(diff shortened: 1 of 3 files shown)


tricoder42:
Hey @ckknight, thank you for your PR! This is already fixed in v3, although a bit differently. I've just checked Next.js example and the order of imports doesn't matter.

Cheers 👍

ckknight:
@tricoder42 is v3 production-ready or do I have to rely on manually patching @lingui/macro until it is?

tricoder42:
It’s not yet, but it’ll be in 2 weeks

PR #643, opened 2020-04-12
Use micromatch to handle Win path sep

### LINKED ISSUE #642: Cannot read property 'locale' of null for `@lingui/loader`
**Describe the bug**
App crashes when using `@lingui/loader`

**To Reproduce**
Install deps and run example of `create-react-app` from `next` branch.

**Expected behavior**
It should work :)

**Additional context**
image

It's coming from this line. Apparently the `getCatalogForFile` returns `null` for some reason. I will investigate it further, but I suspect it's a classic issue of wrong path separators (I am on Windows).

https://github.com/lingui/js-lingui/blob/24d521ff385647f69f73188fd35e07079efdc1f3/packages/loader/src/index.js#L41

- jsLingui version ``lingui --version``: 3.0.0-10

FredyC:
Fixes #642

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -53,6 +53,7 @@
     "messageformat-parser": "^4.1.2",
     "micromatch": "4.0.2",
     "mkdirp": "^1.0.4",
+    "normalize-path": "^3.0.0",
     "ora": "^4.0.3",
     "pofile": "^1.0.11",
     "pseudolocale": "^1.1.0",
@@ -62,7 +63,6 @@
     "@types/micromatch": "^4.0.1",
     "@types/normalize-path": "^3.0.0",
     "mockdate": "^2.0.2",
-    "normalize-path": "^3.0.0",
     "typescript": "^3.8.3"
   },
   "peerDependencies": {
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -59,7 +59,10 @@
     "ramda": "^0.27.0"
   },
   "devDependencies": {
+    "@types/micromatch": "^4.0.1",
+    "@types/normalize-path": "^3.0.0",
     "mockdate": "^2.0.2",
+    "normalize-path": "^3.0.0",
     "typescript": "^3.8.3"
   },
   "peerDependencies": {
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -51,7 +51,7 @@
     "inquirer": "^7.1.0",
     "make-plural": "^6.2.1",
     "messageformat-parser": "^4.1.2",
-    "minimatch": "^3.0.4",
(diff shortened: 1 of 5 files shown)


FredyC:
Not sure why tests are failing with this change, apparently there are duplicate `@types/react` packages. I have good experience with `npx yarn-deduplicate -s fewer yarn.lock` to fix that.

Also, you said that Appveyor tests are ok, but since they are only running build then it's obvious. It should also run tests and then you would see anything related to Windows path failing due to snapshot testing.

FredyC:
Hm, looks like that `micromatch` is also a piece of crap. When it gets Windows path, it fails too. I guess I will just approach it differently and try to normalize any path to POSIX format instead of relying on these weird tools.

tricoder42:
> Also, you said that Appveyor tests are ok, but since they are only running build then it's obvious. It should also run tests and then you would see anything related to Windows path failing due to snapshot testing.

Ups, good point. I'll fix it. 

> Btw that failing linter check should be easily fixable by running npx yarn-deduplicate -s fewer yarn.lock. Duplicate types packages are evil :)

Yeah, my bad. I've fixed `yarn.lock` last night with `npx yarn-deduplicate` but forgot to commit... It's already in `next` branch and CI build is green.

FredyC:
This is harder than it looks, I will get back to it later today.

tricoder42:
Yeah, this is tricky. I'm having fun with #644 myself.

I think we should convert all paths to POSIX as soon as possible (e.g. when catalog is created) and internally use only POSIX. If `minimatch` or other glob library supports it, then this is the way to go. Otherwise we'll have more troubles down the line, especially with snapshot testing.

For example, message origins are always converted to POSIX paths - that way message catalogs are the same no matter if you extract them on win or other system.

FredyC:
So, should I close this PR? Because you seem to be solving the same thing in #644. Not sure why such double work.

FredyC:
Although I took a bit different approach. Perhaps if you have other work, I can figure it out in catalogs and then you can continue with Appveyor.

FredyC:
@tricoder42 I think it's ready now, tests are passing on WIndows as well, at least for catalog. I skipped other test files :)

The test `Catalog make should collect and write catalogs` gave me a hard time because some other test writes to a real file system without the cleanup and this one reads it making test extremely flaky locally.

tricoder42:
> So, should I close this PR? Because you seem to be solving the same thing in #644. Not sure why such double work.

I just wanted to enable tests in Appveyor and fix any related problems. The I realized that all failing tests are the ones you're trying to fix, so I stepped aside and waited.

> I skipped other test files :)

I don't see any skipped tests. Everything looks good. Great job!

PR #647, opened 2020-04-13
fix: windows compatibility in catalogs (#643)

FredyC:
Small oversight for paths not starting with `.\`

cc @tricoder42


tricoder42:
Thanks. I've just merge it into #644. It's definitely improvement. I managed to fix failing snapshot tests and it seems to be working now. Waiting for final build.

Thanks for debugging 👍

PR #648, opened 2020-04-13
Fallback useLingui to runtimeConfigModule

FredyC:
@tricoder42 I have something like this in mind, but the `runtimeConfigModule` is a bit of pickle as it cannot be loaded dynamically like that (build is failing) :/ So the other solution that comes to my mind is to pass `i18n` to `Trans` props and then macro would inject it there based on config.

### DIFF
--- a/packages/react/src/I18nProvider.tsx
+++ b/packages/react/src/I18nProvider.tsx
@@ -1,6 +1,7 @@
 import React, { FunctionComponent } from "react"
 import { I18n } from "@lingui/core"
 import { TransRenderType } from "./Trans"
+import { getConfig } from "@lingui/conf"
 
 interface I18nContext {
   i18n: I18n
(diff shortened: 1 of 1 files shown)


FredyC:
What I generally find weird about the current approach that there is a `runtimeConfigModule` but you still require to manually import `i18n` (whenever it is for you) and pass it into `I18nProvider`. It creates room for error if you ie. pass different `i18n` instance to the provider.

packages/react/src/I18nProvider.tsx
```diff
@@ -11,13 +12,14 @@ export interface I18nProviderProps extends I18nContext {}
 
 const LinguiContext = React.createContext<I18nContext>(null)
 
+const config = getConfig({ configPath: process.env.LINGUI_CONFIG })
```
tricoder42, line 15: This definitely won't work in browser because `getConfig` needs access to file system... You would need to turn it into a macro which loads config at compile time.

packages/react/src/I18nProvider.tsx
```diff
@@ -11,13 +12,14 @@ export interface I18nProviderProps extends I18nContext {}
...
+
 export function useLingui(): I18nContext {
   const context = React.useContext(LinguiContext)
 
-  if (process.env.NODE_ENV !== "production") {
-    if (context == null) {
-      throw new Error("useLingui hook was used without I18nProvider.")
-    }
+  if (context == null) {
+    return require(i18nImportModule)[i18nImportName]
```
tricoder42, line 22: I wouldn't bother importing `i18nImportModule` here... This should be only for the simplest cases when you use the default `i18n` from `@lingui/core`. As soon as you start using custom `i18n` instance, you would have to use `I18nProvider`

FredyC:
Closing in favor of #649

PR #651, opened 2020-04-13
Improve macro typings

FredyC:
It bugged me that `plural` macro accepts `Object` instead of correct options. And I slightly improved small things around. Also removed `defineMessages` declaration which no longer exists, does it?


tricoder42:
Looks great, thanks 👍

PR #652, opened 2020-04-13
Enable add-locale with deprecation message

FredyC:
Even though `add-locale` is deprecated, it's probably better to keep in the CLI so it can show deprecation message. I was fairly confused myself when I tried to run it and it was giving me rather cryptic error messages.

Also `init` command is some thing of the past, so I removed it from there.


tricoder42:
Pretty good, thanks! 👍

PR #653, opened 2020-04-13
Add labeler workflow to mark changed packages in PRs

lucasecdb:
Adds the [labeler](https://github.com/actions/labeler) workflow to automatically add the labels for pull-requests (like the ":package: cli", etc)


lucasecdb:
nevermind, this won't work for PRs made from forks

PR #654, opened 2020-04-13
Improve React related docs

FredyC:
Just initial attempt for now so we can discuss over specific parts of the docs.


docs/tutorials/setup-cra.rst
```diff
@@ -77,24 +68,7 @@ Install
       # or using Yarn
       yarn extract
 
-   There should be no error and you should see output similar following:
```
FredyC, line 80: I have removed this because I noticed stats are commented out in extract command. Then I noticed a similar table is in many places in the docs. I guess it will have to be updated globally when stats are fixed.

tricoder42: Yeah, I've forgot to fix stats. Now with multiple catalogs it isn't trivial :)

tricoder42:
Hey @FredyC, what's the status of this PR? Should we merge it as it is (it doesn't have to be finished) or wait for more improvements?

FredyC:
Well, apparently there is a conflict and I don't really have the capacity to deal with now. Sorry.

PR #660, opened 2020-04-19
Fix babel plugin-syntax-jsx package name in TypeScript extractor

lucasecdb:


tricoder42:
Good catch, thank you! Just curious - do you use Typescript directly or rather `@babel/preset-typescript`?

lucasecdb:
@tricoder42 I use the babel preset. Mostly because the webpack loader for TypeScript isn't as fast as babel.

PR #661, opened 2020-04-19
Feature: add support for `formatOptions`

### LINKED ISSUE #634: Add formatOptions.origin to PO format
Working on reference documentation for configuration I've found that `formatOptions.origins` is missing in `@lingui/conf` package (probably due to wrong merge). Moreoever, even the original [commit](https://github.com/lingui/js-lingui/commit/b1a109aa30f3fc3d8279cf978cd56528f7ca197a) adds this feature to `lingui` format only and not PO format. 

This feature removes origin lines (e.g. `App.js:3`) from message catalogs which is useful to avoid unnecessary merge conflicts.

Tasks:

- [x] Add `formatOptions` to `@lingui/conf` package
- [x] Check that both `lingui` and `po` formats use `formatOption.origins`
- [x] If possible, write tests to cover cases with and without this option.

Additional info:

- Use `next` branch as a base
- Use original [commit](https://github.com/lingui/js-lingui/commit/b1a109aa30f3fc3d8279cf978cd56528f7ca197a) as a reference.

kirillku:
Fixes #634 

Tasks:
- [x] Add formatOptions to `@lingui/conf` package
- [x] Check that both `lingui` and `po` formats use `formatOption.origins`
- [x] If possible, write tests to cover cases with and without this option

Had to fix some typings and update snapshots to pass ci check.

### DIFF
--- a/packages/cli/src/api/__snapshots__/catalog.test.ts.snap
+++ b/packages/cli/src/api/__snapshots__/catalog.test.ts.snap
@@ -295,8 +295,8 @@ Object {
     obsolete: false,
     origin: Array [
       Array [
-        1,
         catalog.test.ts,
+        1,
       ],
     ],
     translation: Label,
@@ -316,8 +316,8 @@ Object {
     obsolete: false,
     origin: Array [
       Array [
-        1,
         catalog.test.ts,
+        1,
       ],
     ],
     translation: A,
@@ -326,8 +326,8 @@ Object {
     obsolete: false,
     origin: Array [
       Array [
-        1,
         catalog.test.ts,
+        1,
       ],
     ],
     translation: B,
@@ -336,8 +336,8 @@ Object {
     obsolete: false,
     origin: Array [
       Array [
-        1,
         catalog.test.ts,
+        1,
       ],
     ],
     translation: C,
@@ -346,8 +346,8 @@ Object {
     obsolete: false,
     origin: Array [
       Array [
-        1,
         catalog.test.ts,
+        1,
       ],
     ],
     translation: D,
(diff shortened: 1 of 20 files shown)


kirillku:
@tricoder42 could you help me with ci, there is some test failed which is not clear for me how to fix

tricoder42:
Unit tests are fine, but integration tests fail (`yarn release:test` to run full test suit or just` yarn build && yarn test:integration` to skip unit tests).

I'm suspicous it's caused by circular imports (`formats/index.ts` imports `lingui.ts` which imports types from `formats/index.ts`). I tried to remove them, but it didn't help. I'll try it again in the afternoon.

Thank you very much anyway! It looks pretty good 👍

tricoder42:
All circular imports are just types... Is there a better way how to use types across the codebase rather than importing them? I saw in few projects that they put then in `*.d.ts` files and just use them (like global types).

kirillku:
@tricoder42 I found the issue. It was because i'm importing types from `@lingui/cli/src/api/formats` in `@lingui/conf`. Like using nested path in other package. I fixed that, but now linter is failed. For some reason types are not getting exported from `@lingui/conf`

tricoder42:
@kirillku Yeah, types from conf package weren't exported nor created. I've added them manually for now and I'll fix build process when I have more time.

I also noticed there're few broken types in tests an done feature in catalogs, which was broken earlier (`prevFormat`). I'll fix it later as well.

tricoder42:
Thanks for the hard work! I've just resolved all conflicts and merge it 8d4af154. Out of habbit I've rebased this branch to `next` instead of merging it :/

Anyway, thank you again! This PR really helped a lot.

PR #662, opened 2020-04-19
Export MessageDescriptor interface in core package

lucasecdb:
This is required because we are [importing it in `@lingui/macro`](https://github.com/lingui/js-lingui/blob/next/packages/macro/index.d.ts#L2) but it wasn't exported in the `@lingui/core` package, causing it to be treated as `any` and forcing users to use the `skipLibCheck` compiler option in their `tsconfig.json`.

### DIFF
--- a/packages/core/src/i18n.ts
+++ b/packages/core/src/i18n.ts
@@ -1,5 +1,5 @@
 import { interpolate } from "./context"
-import { isString, isFunction, isEmpty } from "./essentials"
+import { isString, isFunction } from "./essentials"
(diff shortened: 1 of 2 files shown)


packages/core/src/i18n.ts
```diff
@@ -1,5 +1,5 @@
 import { interpolate } from "./context"
-import { isString, isFunction, isEmpty } from "./essentials"
```
lucasecdb, line 2: this was an unused import

tricoder42:
Good point, thanks 👍

PR #671, opened 2020-04-20
fix: type of compiled messages

anuraaga:
`AllMessages` is a map from locale to messages, but this is the type for an individual, compiled locale.

### DIFF
--- a/packages/cli/src/lingui-compile.ts
+++ b/packages/cli/src/lingui-compile.ts
@@ -84,8 +84,8 @@ function command(config, options) {
         const typescriptPath = compiledPath.replace(/\.jsx?$/, "") + ".d.ts"
         fs.writeFileSync(
           typescriptPath,
-          `import { AllMessages } from '@lingui/core';
-declare const messages: AllMessages;
+          `import { Messages } from '@lingui/core';
+declare const messages: Messages;
 export = messages;
 `
         )
--- a/packages/cli/src/lingui-compile.ts
+++ b/packages/cli/src/lingui-compile.ts
@@ -86,7 +86,7 @@ function command(config, options) {
           typescriptPath,
           `import { Messages } from '@lingui/core';
 declare const messages: Messages;
-export = messages;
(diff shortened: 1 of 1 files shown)


packages/cli/src/lingui-compile.ts
```diff
@@ -84,8 +84,8 @@ function command(config, options) {
         const typescriptPath = compiledPath.replace(/\.jsx?$/, "") + ".d.ts"
         fs.writeFileSync(
           typescriptPath,
-          `import { AllMessages } from '@lingui/core';
-declare const messages: AllMessages;
+          `import { Messages } from '@lingui/core';
+declare const messages: Messages;
 export = messages;
```
tricoder42, line 89:
```suggestion
export { messages };
```

I believe we should also change this, because messages are no longer default export, but a named one.

anuraaga: Ah yup

tricoder42:
Thanks a lot 🚀

anuraaga:
By the way, I found the ux or the default import better FWIW. Now I always have to add some boilerplates

```
import { messages as enMessages} from "en/messages";
import { messages as esMessages } from "es/messages";
```

vs


```
import enMessages from "en/messages";
import esMessages from "es/messages";
```

Not a big deal though just to share

tricoder42:
I don't like it much either, but compilers add `default` key to exported object, ie:

```
export default {
  "Hello": "Salut"
}
```

becomes something similar to this:

```
var messages = {
  "Hello": "Salut"
}

module.exports = messages
module.exports['default'] = messages
```

I guess it does for compatability reasons. I was afraid that now when I would iterate over messages, I would also get `default` key, which isn't desired.

I'm not 100% sold to this design decision yet, feel free to suggest better approach. What I do right now to avoid import aliases is:

```
import enMessages from "./locale/en/messages.js"

i18n.load(en, enMessages.messages)
```

Issue #591, opened 2019-11-19
Translation validation


mpazik:
In our team, we have a problem with the verification of translations. Sometimes translators do a typo in variable names or forgot to close a tag or mess up tags orders and nesting.
If there is a problem in translation js lingui in production mode throws a fatal exception as it can not match variables to the translation and the page is not rendered. Which is really bad.

Right now we have a test parse source and translation for each language and compares but it's not ideal.
It would be nice to have a built-in tool in js lingui CLI, something like `lingui verify` or `lingui compile --check` or be part of `lingui compile --strict`

I could contribute with PR adding this functionality. 
Please let me know if there is some other way to validate the translations that I missed and if not what is preferred way of implementing it from the CLI perspective :)

This ticket could be related to
https://github.com/lingui/js-lingui/issues/397
https://github.com/lingui/js-lingui/issues/4

tricoder42:
Hey @mpazik, I would love to see this feature in Lingui.

1. How do you validate typos in message arguments? We could extend macros by adding "expected" arguments into message metadata and use it for validation. This would be useful e.g. when you use custom IDs, which doesn't include placeholders:

```
<Trans id="Label / Hello">Hello ${name}</Trans>
//<Trans id="Label / Hello" message="Hello {name}" meta={{ args: ['name'] }} />
```

2. Validation of HTML should be possible without other changes I guess.

Please let me know if you're still interested. If so, please use `next` branch as a base for your PR. It'll be released very soon (in matter of days or 1-2 weeks)

theKashey:
(we are on one team) Right now we just checking that all required placeholders exists. Nothing more. That solves 99% of problems.
The check is separated per `{}` and `<>` placeholders, so they cannot change their types.

tricoder42:
> all required placeholders exists

How do you know what placeholders the message uses? You have them in message id?

```js
{
  // if placeholders are included in message, then it's easy
  "Hello {name}": "Salut {nom}",
  
  // if you use custom message Ids, then it's not obvious unless you check the code 
  "Label / Greetings": "Salut {nom}"
}
```

mpazik:
@tricoder42 We load the source file (extracted translations) to memory and then compare translation agains it.
Not sure if loading a file to memory is an issue. Our source file have less then 100KB but it think it should be fine up to like 100MB which is a lot of text. Not sure if there are such cases.

Regarding what exactly we do check:
- all declared ICU variables in translation exists in the source (translation can omit unused variables.
[...]

PR #675, opened 2020-04-22
Fail gracefully for wrong use of elements in translation

mpazik:
Every few times we pull translation updates it turns out that translators do typo or miss-use either element or ICU formats.
Some of these mistakes can cause severe issues. It happens once that broken translation was released to prod and total break our application for customers.
I heard that there are other teams in my company that face similar issues with js-lingui.

We internally created a translation validator and as discussed in the issue https://github.com/lingui/js-lingui/issues/591 I'll try to rise a PR with that feature to the original library.

However, I still believe that any typo or mistake in the translation should not cause the page to break.

In this PR I tried to address the issue with errors caused by wrong use of react components.
- I introduced tests for the cases I thought could cause the error.
- I fixed the case where a not declared element is used in translation
- I fixed for the case where an unpaired component is used as paired (I couldn't find a nice way to implement it, I ended up coping the list of elements from react source)

All suggestions are welcome :)

### DIFF
--- a/packages/react/src/format.test.tsx
+++ b/packages/react/src/format.test.tsx
@@ -51,4 +51,29 @@ describe("formatElements", function () {
       'Before <a href="/about">Inside <strong>Nested</strong> Between <br> After</a>'
     )
   })
+
+  it("should ignore non existing element", function() {
+    expect(html(formatElements("<0>First</0>"))).toEqual("First")
+    expect(html(formatElements("<0>First</0>Second"))).toEqual("FirstSecond")
+    expect(html(formatElements("First<0>Second</0>Third")))
+        .toEqual("FirstSecondThird")
+    expect(html(formatElements("Fir<0/>st"))).toEqual("First")
+  })
+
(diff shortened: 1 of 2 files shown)


packages/react/src/format.js
```diff
@@ -30,7 +48,20 @@ function formatElements(
...
+    if (!element || (voidElementTags[element.type] && children)) {
+      if (!element) {
+        console.error(`Can use element at index '${index}' as it is not declared in the original translation`)
+      } else {
+        console.error(`${element.type} is a void element tag therefore it must have no children`)
+      }
+
+
+      // ignore problematic element but push its children and elements after it
+      element = createElement(Fragment)
```
mpazik, line 62: I tried few approaches but to make code readable, without many branches, I decided to create an empty element.

tricoder42:
Hey @mpazik, this looks good to me! Compared to #676, we indeed need a runtime check. I like the graceful ignoring of invalid elements and still logging a warning. Good job!

mpazik:
Thanks @tricoder42 🙂

PR #676, opened 2020-04-22
Handle icu errors gracefully

mpazik:
Every few times we pull translation updates it turns out that translators do typo or miss-use either element or ICU formats.
Some of these mistakes can cause severe issues. It happens once that broken translation was released to prod and total break our application for customers.
I heard that there are other teams in my company that face similar issues with js-lingui.

We internally created a translation validator and as discussed in issue #591 I'll try to rise a PR with that feature to the original library.

However, I still believe that any typo or mistake in the translation should not cause the page to break.

In this PR I tried to address the issue with errors caused by a wrong use of ICU formats.
- I introduced tests for the cases I thought could cause the error.
- I fix case where plurals are not declared - I assumed that declaration of them is not mandatory but a translator could still use it
- No more error when custom ICU format is used
- I handle exception from formatters on wrong arguments

### DIFF
--- a/packages/core/src/formats.ts
+++ b/packages/core/src/formats.ts
@@ -7,8 +7,13 @@ export function date(
 ): (value: string | Date) => string {
   const formatter = new Intl.DateTimeFormat(locales, format)
   return (value) => {
-    if (isString(value)) value = new Date(value)
-    return formatter.format(value)
+    try {
+      if (isString(value)) value = new Date(value)
+      return formatter.format(value)
+    } catch (error) {
+      console.error(`Could not format a date: ${value}`, error)
+      return `${value}`;
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


packages/core/src/formats.ts
```diff
@@ -5,17 +5,39 @@ export function date(
...
-  const formatter = new Intl.DateTimeFormat(locales, format)
-  return (value) => {
-    if (isString(value)) value = new Date(value)
-    return formatter.format(value)
+  try {
+    const formatter = new Intl.DateTimeFormat(locales, format)
+    return (value) => {
+      try {
+        if (isString(value)) value = new Date(value)
+        return formatter.format(value)
```
mpazik, line 13: In the case of the date, the error is thrown by `format` method when a value is in the wrong type.
In the case of the number, the error is thrown when formater is created without sufficient params. I tried few cases but the wrong data type seemed to be handled well by the number formater.

Neverthanless I couldn't find a prove that it can not throw an exception so I wrapped both cases in try/catch.

mpazik: I give it another tried and I removed any try/catch I couldn't reproduce in order to not lower the test coverage

tricoder42:
Hey @mpazik, I understand it's annoying when prod build breaks because of one message... However, since we already have a compile command, what if we check all messages at compile time? Then we could use just one top-level try/catch block to catch any runtime errors. That way we can catch any possible errors at compile time when it's still time to fix it.

PR #678, opened 2020-04-23
Adds missing dependency

arcanis:
Required here: https://github.com/lingui/js-lingui/blob/next/packages/macro/src/index.ts#L2

### DIFF
--- a/packages/macro/package.json
+++ b/packages/macro/package.json
@@ -29,7 +29,7 @@
   ],
   "dependencies": {
     "@babel/runtime": "^7.9.2",
-    "@lingui/conf": "^3.0.0-11",
+    "@lingui/conf": "0.0.0-managed-by-release-script",
     "ramda": "^0.27.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


packages/macro/package.json
```diff
@@ -29,6 +29,7 @@
   ],
   "dependencies": {
     "@babel/runtime": "^7.9.2",
+    "@lingui/conf": "^3.0.0-11",
```
tricoder42, line 32: Good catch! Just use placeholders for `@lingui/*` dependency versions, which is replaced with the correct version during build:

```suggestion
    "@lingui/conf": "0.0.0-managed-by-release-script",
```

tricoder42:
Thanks a lot 👍

arcanis:
Btw, note that Yarn 2 now [automatically the dependency ranges](https://yarnpkg.com/features/release-workflow#auto-updated-dependencies) if needed 🙂

tricoder42:
Is it released already? I still use yarn `1.22.4` but I haven't checked for a while

arcanis:
Yep it got released in January - the latest on npm is still 1.22.4 because we switched to a per-project package manager versioning. More details here: https://yarnpkg.com/getting-started/install

tricoder42:
Nice! Thanks a lot! 🙏

PR #685, opened 2020-04-29
fix: Support comments inside JSX macros by ignoring empty jsx expressions

### LINKED ISSUE #684: Support comments inside JSX macros by ignoring empty jsx expressions
**Describe the bug**
When using a `<Trans>` element with a nested element like `{/* This is a comment */}`, it complains of @lingui/macro: Property value of ObjectProperty expected node to be of a type ["Expression","PatternLike"] but instead got "JSXEmptyExpression" Learn more: https://www.npmjs.com/package/@lingui/macro

**To Reproduce**

```

   return (<Trans>
    {/* TODO: Come up with a better translation than this */}
    This should be translated!
  </Trans>)
}
```

**Expected behavior**
The comment should be ignored completely, leaving just the "This should be translated!" text.

**Additional context**

- jsLingui version ``2.9.1``
- Babel version ``babel-core@7.0.0-bridge.0``
- Your Babel config (e.g. `.babelrc`) or framework you use (Create React App, Meteor, etc.): Create React App

ckknight:
- this adds explicit handling of JSXEmptyExpression within
  packages/macro/src/icu.ts, where the token is skipped over entirely
- Fixes #684

### DIFF
--- a/packages/macro/src/icu.ts
+++ b/packages/macro/src/icu.ts
@@ -7,6 +7,7 @@ export default function ICUMessageFormat() {}
 ICUMessageFormat.prototype.fromTokens = function (tokens) {
   return (Array.isArray(tokens) ? tokens : [tokens])
     .map((token) => this.processToken(token))
+    .filter(Boolean)
     .reduce(
       (props, message) => ({
         ...message,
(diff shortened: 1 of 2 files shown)


tricoder42:
Wow, that was fast! It looks great and build is green. Thanks a lot! 👍 

I should issue another pre-release by the end of this week. I want to go through all PRs and merge as much as possible.

PR #693, opened 2020-05-14
Feature: add csv support

AOHUA:
I don't actually like the csv format, however in my current situation, the translators only accept csv files for convenience, and I have to convert the po file into csv and convert it back when the translation is ready. I'm not familiar with ramda, not sure if some of the code can be replaced by using ramda.

### DIFF
--- a/packages/cli/src/api/formats/csv.ts
+++ b/packages/cli/src/api/formats/csv.ts
@@ -1,0 +1,52 @@
+import fs from "fs"
+
+import { writeFileIfChanged } from "../utils"
+import { MessageType } from "../types"
+
+const serialize = (catalog) => {
+  var row = ""
+  for (const key of Object.keys(catalog)) {
+    var cur = catalog[key]
+    row += `\"${key}\"`
+    row += `,\"${cur.translation}\"` || ""
+    row += "\n"
+  }
+  return row
+}
+
+const deserialize = (raw: string): { [key: string]: MessageType } => {
+  const rows = raw.split("\n")
+  var rawCatalog = {}
+  rows.forEach((row) => {
+    for (let i = 0; i < row.length; i++) {
+      if (row[i] === "," && row[i - 1] === '"' && row[i + 1] === '"') {
+        rawCatalog[row.substring(1, i - 1)] = {
+          translation: row.substring(i + 2, row.length - 1),
+          obsolete: false,
+          message: null,
+          origin: [],
+        }
+      }
+    }
+  })
+  return rawCatalog
+}
+
+export default {
+  catalogExtension: ".csv",
+
+  write(filename, catalog) {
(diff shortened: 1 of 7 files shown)


packages/cli/src/api/formats/csv.ts
```diff
@@ -0,0 +1,52 @@
...
+
+import { writeFileIfChanged } from "../utils"
+import { MessageType } from "../types"
+
+const serialize = (catalog) => {
+  var row = ""
+  for (const key of Object.keys(catalog)) {
+    var cur = catalog[key]
+    row += `\"${key}\"`
+    row += `,\"${cur.translation}\"` || ""
```
tricoder42, line 11: What if the key or translation contain a separator (comma in this case)? Shouldn't we escape the value? Feel free to use any external dependency for CSV files. I'm planning to move formats to external packages anyway, so it wouldn't add dependency to everyones project.

AOHUA: I have [enclosed the field in quotes](https://stackoverflow.com/questions/4617935/is-there-a-way-to-include-commas-in-csv-columns-without-breaking-the-formatting), so we can include separators in either keys or translations. I have tested with `,` and `"`. It works fine in my current project. I'd love to explore some of the external packages but couldn't find a 2 ways conversion library. Any suggestions?

tricoder42: Hmm, that's really interesting. I would expect it to fail, if the message has one double quote, for example. 

AFAIK, if the message is unquoted, then comma is problem. If the message is surrounded by quotes, then the quotes inside the message become a problem and you need to escape them properly.

I'm not aware of any CSV packages, I'll check them later tonight.

AOHUA: You are right, it does fail in the case that you mentioned. I will try to find some CSV packages too. Thank you for reviewing my PR.

AOHUA: Hi @tricoder42 , I used the [papaparse](https://www.papaparse.com/) package to convert the csv file. Please take a look. Thank you.

AOHUA: I have created tests for it as well. But still didn't pass the coverage check.

AOHUA:
Hi @tricoder42 , I will need this feature very soon, would you like to merge it into next? Thank you.

tricoder42:
Hi @AOHUA, looks good to me. Thank you for all your hard work! I'll publish a pre-release.

AOHUA:
> Hi @AOHUA, looks good to me. Thank you for all your hard work! I'll publish a pre-release.

Thank you!

tricoder42:
Just published to npm [3.0.0-12](https://www.npmjs.com/package/@lingui/cli/v/3.0.0-12)

PR #694, opened 2020-05-15
CLI: Explicitly sort directory listings

ionut-botizan:
When reading the files in a directory in order to extract translation strings, the order in which the files are returned varies based on OS, sometimes resulting in unnecessary diffs when extraction is performed on different operating systems.

For example, runing `lingui extract` on Windows after a file was previously "extracted" on MacOS would generate a diff similar to this:
```patch
-#: src/components/admin/content/admin-event-page/AdminEventPage.tsx:217
 #: src/components/admin/content/AdminNugget.tsx:411
+#: src/components/admin/content/admin-event-page/AdminEventPage.tsx:217
 msgid "Preview"
 msgstr "Vorschau"
```

### DIFF
--- a/packages/cli/src/api/extract.js
+++ b/packages/cli/src/api/extract.js
@@ -55,6 +55,7 @@ export function extract(
     if (fs.statSync(srcFilename).isDirectory()) {
       const subdirs = fs
         .readdirSync(srcFilename)
+        .sort()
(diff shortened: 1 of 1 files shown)


tricoder42:
Ah, I forgot to merge this one. Thanks for both PRs!

PR #695, opened 2020-05-15
CLI: Explicitly sort directory listings

ionut-botizan:
When reading the files in a directory in order to extract translation strings, the order in which the files are returned varies based on OS, sometimes resulting in unnecessary diffs when extraction is performed on different operating systems.

(Basically the same thing as pull request #694 but for the `next` branch)


tricoder42:
This make sense! Thank you 👍

tricoder42:
Just published to npm [3.0.0-12](https://www.npmjs.com/package/@lingui/cli/v/3.0.0-12)

PR #696, opened 2020-05-15
Update react-int's rich text formatting example

petersendidit:
`react-intl` supports [rich text formatting](https://formatjs.io/docs/react-intl/components#rich-text-formatting) that is very similar to the way `LinguiJS` does. This updates the doc's to correctly show that.


tricoder42:
Looks good, thanks 👍

PR #698, opened 2020-05-18
Use type instead of interface

FredyC:
@tricoder42 It was just a couple of places in the end. I've also included ESLint rule to enforce this so it stays consistent.


FredyC:
@tricoder42 Do you want to merge this or something isn't right?

tricoder42:
Looks good, thank you!

PR #699, opened 2020-05-20
Fix import of setupI18n in @next React docs.

tstirrat15:
`setupI18n` isn't exported by the react module, so it has to be imported from core. I'm not sure if this is an issue with the lib or with the docs, though.


tricoder42:
Yeah, this is docs issue. Thanks for the fix!

PR #704, opened 2020-06-05
[next] Plural bug demo

tienpham94:
This is only a demo for the plural bug in @lingui/next and not to be merged
When upgrading to @lingui/react v3/next when using `{Plural or plural} from "@lingui/macro"` I have an error: `plurals is not a function`
image

I made this demo to show the Plural bug that exists in @lingui/next
Steps: 
- Checkout the branch
- cd into examples/create-react-app
- yarn start
Without `<Plural>` and `{i18n._(plural(` it works
image
When I add `<Plural>` or `{i18n._(plural(` it breaks with message `plurals is not a function`
```
{i18n._(
          plural(2, {
            one: "You have # friend",
            other: "You have # friends",
          })
        )}
        <Plural
          value={friends}
          one="You have # friend"
          other="You have # friends"
        />
```
image
Added localeData manually (already there in the create-react-app template) but the problem still persists

```
import {i18n} from '@lingui/core';
import { en, ru } from 'make-plural/plurals'
i18n.loadLocaleData('en', { plurals: en })
i18n.loadLocaleData('ru', { plurals: ru })
```


tienpham94:
Link to the bug that I opened #705

tricoder42:
Hey @tienpham94, thanks for the bug report. It appears that the problem is caused by initial render before locales are loaded. Locales and messages are loaded using async `import` so during the first render there're no messages nor locale data (plurals).

One solution is to add `i18n.activate(defaultLocale)` right above `activate('en')` in `i18n.config.js`. It hides the error, but sometimes you end up with untranslated strings because messages are probably loaded too late. Another option is a sipmle component with re-renders the tree when locale changes, like this: https://github.com/lingui/js-lingui/blob/1ef1afcacd55cee0e37f0077c5bf06d3ef03c549/examples/next-js/pages/_app.tsx#L24-L38

But I guess we should make this behavior default with optional opt-out. I'll add it to the provider.

tricoder42:
Hi @tienpham94, since 3.0.0-18 LinguiJS prevents rendering before messages are loaded:

https://github.com/lingui/js-lingui/blob/48268e014ce47c782e979bbafbe6c608f4348ffb/packages/react/src/I18nProvider.tsx#L64-L71

Please check it out and let me know if you still have problems. Cheers 👍

PR #711, opened 2020-07-01
Prevent trimming defaults for Trans component in Production env

### LINKED ISSUE #404: t and Trans behaviour is inconsistent
**Describe the bug**

Assume there is no fallbackLocale. And we have following code

```js
i18n._(t("component.name")`Your name`)

<Trans id="component.name1">
  Your name
</Trans>
```

After compiling it will be

```js
b._({id:"component.name",defaults:"Your name"})

i.a.createElement(m.Trans,{id:"component.name1"}))
```

First one preserve default in source code, but second one not.

**To Reproduce**

```js

return [
i18n._(t("component.name")`Your name`),
<Trans id="component.name1">
  Your name
</Trans>
}]
```

**Expected behavior**

I expect them to behave consistently either both remove default from compiled version or both leave it

**Additional context**

- jsLingui version ``2.7.1``
- Babel version 
├─ babel-core@7.0.0-bridge.0
├─ babel-register@6.26.0
│  └─ babel-core@6.26.0
└─ jest-runtime@22.4.4
   └─ babel-core@6.26.0
- Your Babel config CRA `"react-scripts": "2.0.0-next.a671462c"`

I can provide reproducible example if required

Related: https://github.com/lingui/js-lingui/issues/396, https://github.com/lingui/js-lingui/issues/383

NMVikings:
This PR fixes #404 and prevents trimming `defaults` for the Trans component.
So now `t` and `Trans` have similar behavior

```
i18n._(t("component.name")`Your name`)
<Trans id="component.name1">
  Your name
</Trans>
```

After compiling it will be
```
b._({id:"component.name",defaults:"Your name"})
i.a.createElement(m.Trans,{id:"component.name1", defaults:"Your name"}))
```

### DIFF
--- a/packages/babel-plugin-transform-react/src/transformer.js
+++ b/packages/babel-plugin-transform-react/src/transformer.js
@@ -425,8 +425,7 @@ export default class Transformer {
 
     if (process.env.NODE_ENV === "production") {
       node.openingElement.attributes = attrs.filter(
-        node =>
-          !this.isDefaultsAttribute(node) && !this.isDescriptionAttribute(node)
+        node => !this.isDescriptionAttribute(node)
       )
     }
   }
--- a/packages/babel-plugin-transform-react/test/fixtures/env-production-remove-defaults/actual.js
+++ b/packages/babel-plugin-transform-react/test/fixtures/env-production-remove-defaults/actual.js
@@ -1,3 +1,0 @@
-import { Trans } from '@lingui/react';
-
-<Trans id="msg.hello" defaults="Hello World" />
--- a/packages/babel-plugin-transform-react/test/fixtures/env-production-remove-defaults/expected.js
+++ b/packages/babel-plugin-transform-react/test/fixtures/env-production-remove-defaults/expected.js
@@ -1,3 +1,0 @@
-import { Trans } from '@lingui/react';
-
-<Trans id="msg.hello" />;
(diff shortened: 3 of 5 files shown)


tricoder42:
Hi @NMVikings, thanks for the PR, but it should be actually the other way round — we should remove `defaults` from `t` macro. The reason is that fallback messages are resolved at compile time and there's no reason to keep them in production.

NMVikings:
Why so?
If I have only dynamic translations that I fetch from some service like https://locize.com/ or any other and if loading failed I don't want to show i18n keys. I want to show my default value and I expect that behavior. Or if I have a typo in a key I still want to have text, not a key.

That's the reason for me – I don't want to show keys in production for users, I want to have a fallback if something is wrong with translations.

tricoder42:
I agree with you but LinguiJS does all you describe during build when message catalogs are compiled. Otherwise your bundle is cluttered with default messages which you might not even use. If you have all messages translated, then all defaults are useless and removed from the bundle. Lingui does it during the build phase — it removes default messages from code and adds them to missing keys in compiled locales. That way you can always have a fallback without having duplicate messages in source code.

NMVikings:
I see. I add extracting and compiling messages on the prebuild phase of my webpack config, so now I have compiled default translations and use them as a fallback if something goes wrong with our i18n service that provides translations. And all of it to avoid showing translation keys in production.

PR #717, opened 2020-08-01
Update deprecated ref in react doc

margori:
Package 'babel-plugin-lingui-transform-react' is deprecated and '@lingui/babel-plugin-transform-react' should be used in react doc.


tricoder42:
Good catch! Thanks 👍

PR #723, opened 2020-08-20
import setupI18n from core

dominikdosoudil:
import `setupI18n` from `@lingui/core` instead of `@lingui/react`

(`setupI18n` not found in `@lingui/react`)

### DIFF
--- a/docs/ref/react.rst
+++ b/docs/ref/react.rst
@@ -94,7 +94,7 @@ variables and components inside messages. Usage of this component depends on
 whether or not you're using LinguiJS Babel plugins.
 
 Each message is identified by **message ID**.
-``babel-plugin-lingui-transform-react`` automatically generates message ID from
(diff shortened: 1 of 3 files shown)


tricoder42:
Hey, thanks for the PR. Could you please rebase it against `next` branch? That should be the place with latest migration docs.

dominikdosoudil:
@tricoder42 yep, sure. (That's what I've been thinking about, but the "Edit on GitHub" took me to `master`.)

dominikdosoudil:
When I am looking into the next migration guide, I am kinda confused. It's recommended to use default `i18n` instance instead of `setupI18n`. However it was giving some type errors when passing into I18nProvider as i18n prop. Using latest core and react. I'll need to look into it more deeply. This PR is probably wrong, sorry

tricoder42:
No worries. In the v3 your indeed advised to use the default `i18n` instance, but you can also use a custom one. Feel free to reach me if you need anything

Issue #722, opened 2020-08-12
Add a CLI command to validate the catalog on CI


Spy-Seth:
**Is your feature request related to a problem? Please describe.**
One of the common problem with localization process is that as a developer you have to not forgot to run `lingui extract` before committing new/edited/removed string.

**Describe proposed solution**
Add a CLI command like `lingui check-catalog` that verify that all the key of the application are present in all catalog. It will return an errored exit code if some string are not referenced in the catalogs.
 
A verbose mode could also print the location of the missing messages.

I work on it if your ok with the idea

makenosound:
Would love this too. I’m planning on doing a manual version by comparing files before/after `extract` on CI, but it feels like a very useful bit of core functionality.

tricoder42:
Hey @Spy-Seth, good idea!

> A verbose mode could also print the location of the missing messages.

Do you mean location of messages which weren't extracted or which are missing translations? [`lingui compile --strict`](https://lingui.js.org/ref/cli.html#cmdoption-compile-strict) already checks for missing translations.

I'm thinking if we could actually create a `lingui check` command which would run a series of checks:
- Check that running `lingui extract` won't add any new message to catalog (i.e. all messages are already extracted)
- Check that all messages are translated (we would need a way to mark messages which don't require translations)
- Anything else?

I don't mean we should implement everything now. I'm just trying to figure out if there's a usecase for additional checks. Otherwise we might also go with `lingui extract --check` or `--dry-run`.

Spy-Seth:
> Do you mean location of messages which weren't extracted or which are missing translations?

The location of the message that need to be extracted.

> we could actually create a lingui check command which would run a series of checks:

My initial point was only focus on the missing message in the catalog. 

It's a good idea to also provide some check at compile time to detect earlier errors in the catalogs.

But given that each team have its own workflow (external translation tools, external translator, automated translation platform...) and its own quality target (for some team using a fallback on the main language is not an issue) : checking if all message are translated and the requirement should be configurable. Maybe the tolerance on missing translation are not the same in all language too. 
This subject seems more touchy than validating that all string are extracted from the source code.

Depending on the team workflow, the `lingui check` command should be able to only check the message that need to be extracted

tricoder42:
Good points! Alright, feel free to send a PR. Could you please base it on `next` branch? Thanks a lot!

Spy-Seth:
I will propose you something in the next weeks

tricoder42:
[...]

PR #724, opened 2020-08-21
Add `lingui check` cli

Spy-Seth:
This PR aim to implement the `lingui check` command as discuted in issue #722

The core diffing algorithm to detect the missing message have been written. It's simple but seems to be enough to build a detailed output.

### TODOs

- [ ] Write the CLI with a basic output
- [ ] Improve the CLI output to match the proposed one
- [ ] Add a verbose option

### Output

The output should show at lead these informations:
- the message id that is not extracted
- the location of the missing message
- the information in which catalogs the message is missing: a locale list (eg: `cs, fr`) or the catalogs path (eg: `src/messages/cs.po, src/messages/fr.po` )

#### Proposal one

```
$ lingui check

"some.new.message.id" 
  (from `src/componentA/foo.js:1`, and 1 other location)
     ✗ the message is missing in "src/messages/en.po"
     ✗ the message is missing in "src/messages/cs.po"
     ✗ the message is missing in "src/messages/fr.po"

"some.partially.missing.message.id" 
    (from `src/componentA/index.js:1`)
         ✗ the message is missing in "src/messages/fr.po"

✗ There is 2 issues in the catalogs. 

(use "yarn lingui extract" to update catalogs with new messages)
```

#### Proposal two

```
$ lingui check

The file "src/componentA/foo.js: contains 1 message not referenced in all catalogs:
    - "some.new.message.id" at line 1
         ✗ is missing in the catalog "src/messages/en.po"
         ✗ is missing in the catalog "src/messages/cs.po"
         ✗ is missing in the catalog  "src/messages/fr.po"

The file "src/componentA/index.js" contains 1 message not referenced in all catalogs:
    - "some.partially.missing.message.id" at line 3
         ✗ the message is missing in "src/messages/fr.po"

✗ There is 2 issues in the catalogs. 

(use "yarn lingui extract" to update catalogs with new messages)
```

### DIFF
--- a/packages/cli/src/api/__snapshots__/catalog.test.ts.snap
+++ b/packages/cli/src/api/__snapshots__/catalog.test.ts.snap
@@ -1,5 +1,13 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Catalog check should check for missing message in the catalog 1`] = `
+Object {
+  en: Array [
+    How are you today?,
+  ],
+}
+`;
+
 exports[`Catalog collect should extract messages from source files 1`] = `
 Object {
   Hello World: Object {
--- a/packages/cli/src/api/catalog.test.ts
+++ b/packages/cli/src/api/catalog.test.ts
@@ -132,6 +132,24 @@ describe("Catalog", function () {
     })
   })
 
+  describe("check", function () {
+    it("should check for missing message in the catalog", function () {
+      const catalog = new Catalog(
+        {
+          name: "messages",
+          path: fixture("check/locales/{locale}"),
+          include: [fixture("check/src")],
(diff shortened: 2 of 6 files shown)


tricoder42:
@Spy-Seth The CLI test suit isn't ideal. I usually unit test only the core of the command. Then I publish and manualy test the CLI from user perspective. Would be great to automate it

tricoder42:
Hey @Spy-Seth, what's the status of this PR? Are you still working on this?

Spy-Seth:
I'm not currently working on it. Having no way to execute the command in a test is for me a big issue to continue: sadly I did not enough OSS time to resolve this. If you want to close or finish it, I will understand

Issue #731, opened 2020-09-09
When exported as ES module, use the .mjs file extension


laszbalo:
**Is your feature request related to a problem? Please describe.**
For a while node.js natively supporting es modules, but expecting those files to have .mjs extension. At the moment I can set the module format to `es`, but in order to use that, I have to use module bundler.

**Describe proposed solution**
When module format set to `es`, the file extension of the generated files should be `.mjs`.

**Describe alternatives you've considered**
Manually renaming all the generated files after `lingui compile`:
```shell
for f in $(find ./locale/ -type f -name '*.js'); do cp "$f" "${f%.js}.mjs"; done
```

tricoder42:
Hey @laszbalo, this sounds like a good idea. Feel free to send a PR and use `next` branch as a base 👍

laszbalo:
Hi,

I tried adding this, but after half an hour of trying, I was not even be able to run the tests  :(

```shell
$ npm i
$ npm test // -> most of the tests are failing
```

For the time being I will just stick to the above one-liner to rename the message files. I might revisit this later, when I have more time.

P.S. Actually I have never worked with Typescript before. Should I have had Typscript in my class path?

tricoder42:
Not sure if `npm` can handle `yarn` workspaces. Could you please try to install dependencies using `yarn`? If there're still errors, could you please post them here? Thanks 👍

laszbalo:
```shell
$ yarn
$ yarn watch   // then pressed 'a' to run all the tests
```

One snapshot test is failing:

```shell
 FAIL  packages/cli/src/api/formats/po.test.ts
  ● pofile format › should write catalog in pofile format

    expect(received).toMatchSnapshot()

    Snapshot name: `pofile format should write catalog in pofile format 1`

    - Snapshot  - 1
    + Received  + 1

    @@ -1,8 +1,8 @@
      msgid ""
      msgstr ""
    - "POT-Creation-Date: 2018-08-27 10:00+0000\n"
    + "POT-Creation-Date: 2018-08-27 11:00+0100\n"
      "Mime-Version: 1.0\n"
      "Content-Type: text/plain; charset=utf-8\n"
      "Content-Transfer-Encoding: 8bit\n"
      "X-Generator: @lingui/cli\n"
      "Language: en\n"

      71 |     const pofile = fs.readFileSync(filename).toString()
      72 |     mockFs.restore()
    > 73 |     expect(pofile).toMatchSnapshot()
         |                    ^
[...]

PR #734, opened 2020-09-12
File extension of ES modules will be .mjs

laszbalo:
#731 Hi this PR attempts to solve this issue.

### DIFF
--- a/packages/cli/src/api/catalog.ts
+++ b/packages/cli/src/api/catalog.ts
@@ -252,7 +252,7 @@ export class Catalog {
   }
 
   writeCompiled(locale: string, compiledCatalog: string, namespace?: string) {
-    const ext = `.${ namespace === "es" ? "m": ""}js`
+    const ext = `.${ namespace === "es" ? "mjs": "js"}`
     const filename = this.path.replace(LOCALE, locale) + ext
 
     const basedir = path.dirname(filename)
--- a/packages/cli/src/api/catalog.ts
+++ b/packages/cli/src/api/catalog.ts
@@ -251,8 +251,9 @@ export class Catalog {
     this.locales.forEach((locale) => this.write(locale, catalogs[locale]))
   }
 
-  writeCompiled(locale: string, compiledCatalog: string) {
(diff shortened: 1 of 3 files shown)


laszbalo:
I am actually not super happy how I test if the compiled catalog is in ES module. Because the user can override the `namespace` parameter from the command line, I can not just use the `catalog.config.namespace` field. For a more robust alternative, I could have modified the the `catalog.writeCompiled()` method to take `namespace` as its third argument.

packages/cli/src/api/catalog.ts
```diff
@@ -251,8 +251,9 @@ export class Catalog {
     this.locales.forEach((locale) => this.write(locale, catalogs[locale]))
   }
 
-  writeCompiled(locale: string, compiledCatalog: string) {
-    const filename = this.path.replace(LOCALE, locale) + ".js"
+  writeCompiled(locale: string, compiledCatalog: string, namespace?: string) {
+    const ext = `.${ namespace === "es" ? "m": ""}js`
```
tricoder42, line 255: Could you please rewrite this condition so it's easier to find it? E.g. when someone search for `mjs` across the codebase

```suggestion
    const ext = `.${ namespace === "es" ? "mjs": "js"}`
```

tricoder42 (commented):
Looks pretty good, thank you! Just one minor suggestion

packages/cli/src/api/catalog.ts
laszbalo: Good idea. I am not super familiar with the Github UI. It offered me an option to `Commit suggestion`, so just did it. Hope it will not mess up the history.

tricoder42:
Great, thanks a lot! I'll publish pre-release tonigh so you can test all your changes 👍

laszbalo:
Thank you.

Also, there is something which I am puzzled with about the way comments are treated in the `next` version.
If you'll have the time, could you please reply to [this](https://github.com/lingui/js-lingui/issues/334#issuecomment-693634628) question as well.

Issue #739, opened 2020-09-15
Make the Babel extractor to support all the file extensions that Babel supports


laszbalo:
**Is your feature request related to a problem? Please describe.**
Hi, it would be really great, if Lingui's extractor could extract the messages from .mjs files. Since Lingui's is using Babel for that, why not to add support for all the [default extensions Babel](https://babeljs.io/docs/en/babel-core#default_extensions) supports?

**Describe proposed solution**
Make the Babel extractor to use the `@babel/core`'s read-only `DEFAULT_EXTENSION` array to determine which files to extract messages from.

**Describe alternatives you've considered**
Only use the `.js`, `.jsx` subset of the default extensions.

tricoder42:
Hey @laszbalo, sounds good! Feel free to send a PR 👍 

PS: I'm gonna review the other PRs of yours today. I've just got back from vacation and I'm kinda busy

laszbalo:
No worries. Just take your time. Hope it was a good one.

tricoder42:
Closed by #740

PR #740, opened 2020-09-15
Make the extractor support Babel's default file extensions. Babel sup…

laszbalo:
…ports the

following file extensions by default: .js, .jsx, .es6, .es and .mjs.
In the current version, Lingui only passes files to Babel with the .js and .jsx extension, limiting the set of file extensions Babel would be able to process otherwise. Allowing Babel to process .mjs files is really handy, when someone would like to use it with native ES modules in recent version of Node.

This PR attempts to solve this issue: #739


tricoder42:
This is great! Thanks a lot! 🚀

Issue #657, opened 2020-04-14
[next] Refactor pseudolocale


FredyC:
I never used pseudo localization so I am not sure how it's supposed to work, but it's currently commented out from extract command. This should be fixed before V3 comes out otherwise it's a breaking change.

https://github.com/lingui/js-lingui/blob/03ce90086a698e0cc030d188c0a7e2cb4d0b787b/packages/cli/src/lingui-extract.ts#L42

tricoder42:
Pseudolocalization creates a catalog in another locale where all translations (or ids) and passed through `pseudolocale` function. I think the only missing piece here is add an extra locale at runtime and catalogs should handle the rest. This happens in `createCompiledCatalog`, so when you run `lingui compile`

tricoder42:
Closed in #741. Thanks @laszbalo

PR #741, opened 2020-09-16
Add back support for pseudo-locale.

laszbalo:
Hi,

once you have the time, please have a look at this one as well.

Works with multiple catalogues, but the `command` function is not exported, therefore I did not write any tests.

#657


tricoder42:
Cool, thanks a lot! Looks good to me. 

> Works with multiple catalogues, but the command function is not exported, therefore I did not write any tests.

Yeah, we should figure out e2e tests for cli at some point... Let's merge it as it is for now.

Thank you!

PR #749, opened 2020-10-09
Update migration-3.rst - `setupI18n` is not exported by @lingui/react anymore

sladek-jan:
Starting from v3.0.0-4, package `@lingui/react` apparently does not export the function `setupI18n` anymore


tricoder42:
Good catch! Thank you

PR #658, opened 2020-04-17
Add option to merge catalogs to single file

johnparn:

PR #754, opened 2020-10-11
568 merge catalogs

### LINKED ISSUE #568: Generate aggregated, compiled messages from multiple po-source files in custom location
Hi,

and thanks for your hard work on linguijs!

We are working in a monorepo and are in the process of changing how handle the locale files. At the moment we use extract and compile on the whole project to generate two files, `/locales/sv.js` and `/locales/en.js`. Using react contex we handle the language switching. But as the project has grown and we are about to engage other teams to work on the code we would like to keep the translations as close to each package and component as possible. 

Looking at https://github.com/lingui/js-lingui/issues/257 and https://lingui-docs-git-next.lingui-js.now.sh/rfc/002_catalogs.html I've done a test setup so that when running extract and compile all locale files are placed in each package and component. I'm using the _lingui@next_ version for this. The structure looks something like this:

```
/
  web/
    locales/
      en.js en.po sv.js sv.po
    src/
      index.js app.js routes.js
  layouts/
    common/
      header/
        locales/
          en.js en.po sv.js sv.po     
        src/
          index.js
    private/
      accounts/
        locales/
          en.js en.po sv.js sv.po
        src/
          index.js
    public/
      contact/
        locales/
          en.js en.po sv.js sv.po
        src/
          index.js
  packages
    package-a/
      locales/
        en.js en.po sv.js sv.po
      src/
        index.js container.js style.scss
    package-b/
      locales/
        en.js en.po sv.js sv.po
      src/
        index.js container.js style.scss
  components/
    component-a/
      locales/
        en.js en.po sv.js sv.po
      src/
        index.js container.js style.scss
    component-b/
      locales/
        en.js en.po sv.js sv.po
      src/
        index.js container.js style.scss      
```

By default all compiled js-files end up in the same `locales` directories as their corresponding po-files. 

Is it possible to compile all po-files in the whole project to two files for example at `/locales/sv.js` and `/locales/en.js`?

That would mean the file tree would look like this:

```
/
  locales/
    en.js sv.js    
  web/
    locales/
      en.po sv.po
    ...
  layouts/
    ...
  packages
    ...
  components/
    ...      
```

The other way forward that comes to my mind is to handle the import of locale files in each and every package/component. I tried to make a stab at it but at the moment the `I18nProvider` does not render any children so I haven't been able to proceed with that.

darkowic:
Resolves #568

Continuation of https://github.com/lingui/js-lingui/pull/658 - I have no write access ofc so I started a new PR

Just thinking if we want to apply this proposition https://github.com/lingui/js-lingui/pull/658#issuecomment-706675746

### DIFF
--- a/docs/tutorials/cli.rst
+++ b/docs/tutorials/cli.rst
@@ -104,7 +104,7 @@ The ``locale`` directory now contains the source catalogs (``messages.json``) an
 the compiled ones (``messages.js``).
 
 It is also possible to merge the translated catalogs into a single file per locale
-by specifying :conf:`mergePath`. For example if :conf:`mergePath` is assigned 
+by specifying :conf:`catalogsMergePath`. For example if :conf:`catalogsMergePath` is assigned
 ``locales/{locale}`` then catalogs will be compiled to ``/locales/cs.js`` and 
 ``/locales/en.js``.
 
--- a/docs/tutorials/cli.rst
+++ b/docs/tutorials/cli.rst
@@ -103,6 +103,11 @@ plural rules to a production ready catalog::
 The ``locale`` directory now contains the source catalogs (``messages.json``) and
 the compiled ones (``messages.js``).
 
+It is also possible to merge the translated catalogs into a single file per locale
+by specifying :conf:`mergePath`. For example if :conf:`mergePath` is assigned 
+``locales/{locale}`` then catalogs will be compiled to ``/locales/cs.js`` and 
+``/locales/en.js``.
+
 Cleaning up obsolete messages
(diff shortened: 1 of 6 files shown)


tricoder42:
Looks pretty good! Thank you @darkowic 

I would definitely rename `mergePath`. I'm not 100% sure if `catalogsMergePath` is the best choice but I definitely find it better than `mergePath`. What do you think?

tricoder42:
Alright, I've renamed `mergePath` to `catalogsMergePath` to keep related config options together.

Thank you @darkowic for help! 👍

PR #755, opened 2020-10-11
Add Notos to showroom

renchap:
Thanks a lot for Lingui, I have been using it for I18n on the frontend for some time, replacing i18next, and it has been a nice improvement.

Will try to update to 3.0 soon, it seems even better!


tricoder42:
Cool, thank you!

PR #756, opened 2020-10-11
feat: introduced withI18n

semoal:
For an easier migration to v3.0.0, i've introduced a little hoc for an easy migration to big codebases where withI18n it's used a lot!

On master branch, this hoc accepts some configuration, should we accept that configuration?

Feel free to request changes

### DIFF
--- a/packages/react/src/I18nProvider.tsx
+++ b/packages/react/src/I18nProvider.tsx
@@ -1,4 +1,4 @@
-import React, { ComponentType, FunctionComponent, ReactNode, useContext } from "react"
+import React, { FunctionComponent } from "react"
 import { I18n } from "@lingui/core"
 import { TransRenderType } from "./Trans"
 
--- a/packages/react/src/I18nProvider.tsx
+++ b/packages/react/src/I18nProvider.tsx
@@ -37,7 +37,7 @@ export function withI18n(o?: object): <P extends { i18n: I18n }>(Component: Comp
           )
         }
       }
-      
+
       const { i18n } = useLingui();
       return <WrappedComponent {...props} i18n={i18n} />;
     }
--- a/packages/react/src/I18nProvider.tsx
+++ b/packages/react/src/I18nProvider.tsx
@@ -1,4 +1,4 @@
-import React, { FunctionComponent } from "react"
+import React, { ComponentType, FunctionComponent, ReactNode, useContext } from "react"
 import { I18n } from "@lingui/core"
 import { TransRenderType } from "./Trans"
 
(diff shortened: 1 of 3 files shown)


packages/react/src/I18nProvider.tsx
```diff
@@ -25,6 +25,34 @@ export function useLingui(): I18nContext {
...
+      if (typeof o === "function" || React.isValidElement(o)) {
+        throw new Error(
+          "withI18n([options]) takes options as a first argument, " +
+            "but received React component itself. Without options, the Component " +
+            "should be wrapped as withI18n()(Component), not withI18n(Component)."
+        )
+      }
+    }
+
+    return class extends React.Component<P> {
```
FredyC, line 40: Why use a class here? You can just go with `useLingui()` hook and functional component. That also means you can drop that check for Provider.

semoal: You're absolutly right Fredy, i'll change it asap, my mind was thinking hoc===classes, and that's why i've used a class

tricoder42: Originally I though about:

```js
function withI18n(Component) {
  return props => {
    const { i18n } = useLingui()
    return <Component {...props} i18n={i18n} />
  }
}
```

but then I saw your comment about refs.. Afterall, function components can't have refs (unless you wrap then in `React.forwardRef`).

But we can keep it simple and add support for refs as soon as someone raises the issue.

Anyway, thanks both for the input!

semoal: Done guys, refactored to functional component! Keeping the error advise if people use it like for me it's a best practice and will avoid a lot of issues of people asking why the withI18n doesnt work:

```
withI18n(Component)
```

instead of:
```
withI18n()(Component)
```

semoal: I'm not sure why tests are failing, on my machine
Test Suites: 1 skipped, 24 passed, 24 of 25 total
Tests:       15 skipped, 212 passed, 227 total

semoal: Tests fixed! :shipit:

tricoder42:
Looks good, thank you @semoal! 👍

renchap:
@semoal Thanks for this, but it looks like `<withI18n>` is not exported from `@lingui/react`. Is this expected?
Also Lingui 2 exported a `withI18nProps` type which was quite useful for TS users. I think it should be kept in 3 as well if we keep this HOC.

semoal:
> @semoal Thanks for this, but it looks like `<withI18n>` is not exported from `@lingui/react`. Is this expected?
> Also Lingui 2 exported a `withI18nProps` type which was quite useful for TS users. I think it should be kept in 3 as well if we keep this HOC.

Uh, you're absolutly right. Ill do the change asap

semoal:
> > @semoal Thanks for this, but it looks like `<withI18n>` is not exported from `@lingui/react`. Is this expected?
> > Also Lingui 2 exported a `withI18nProps` type which was quite useful for TS users. I think it should be kept in 3 as well if we keep this HOC.
> 
> Uh, you're absolutly right. Ill do the change asap

https://github.com/lingui/js-lingui/pull/775
Done =) once @tricoder42 merges, will be fixed, thx for the tip mate

Issue #747, opened 2020-10-07
Trans inside Plural has surrounding whitespace


ThiefMaster:
I have this JSX:

```jsx
      <Plural
        value={numSlots}
        one={
          <Trans>
            <strong>#</strong> slot added
          </Trans>
        }
        other={
          <Trans>
            <strong>#</strong> slots added
          </Trans>
        }
        _0="You haven't added any slots yet"
      />
```

When extracting, I get this:

```
msgid "{numSlots, plural, one {<0>#</0> slot added } other {<1>#</1> slots added } =0 {You haven't added any slots yet}}"
```

As you can see, the one/other cases with Trans have a space at the end of the string, while the 0 case with the normal string doesn't have this space. I think the extra space is a bug, since it doesn't make sense there and surrounding whitespace is generally ugly.

tricoder42:
Hi @ThiefMaster, what version do you use? 2.9 or the 3.0 pre-release?

ThiefMaster:
3.0 prerelease

fny:
A related problem:

`<Trans>Hello there are many <Plural value={this.state.number} one="child" other="children">`

tricoder42:
Indeed, I can confirm this is happening. I've added a test case to `feature/fix-trailing-whitespace` branch (see 0e51288e).

@ThiefMaster Could you please checkout the commit and try to fix it? It must be somewhere inside [macroJsx.ts](https://github.com/lingui/js-lingui/blob/next/packages/macro/src/macroJsx.ts) or [icu.ts](https://github.com/lingui/js-lingui/blob/next/packages/macro/src/macroJsx.ts). Let me know if you need any help.

@fny I believe you're using Lingui 2.9.x right? Please try to reproduce your bug in v3 and let me know.

yozi-developer:
I have the latest lingui and faced the same issue

```tsx

<Plural
      one={
        <Trans>
          Care Navigator{" "}
          <Badge
            aria-label={t`You have unread message`}
          >
            #
          </Badge>
        </Trans>
      }
      other={
        <Trans>
          Care Navigator{" "}
          <Badge
            aria-label={t`You have unread message`}
          >
            #
          </Badge>
        </Trans>
      }
      value={unreadSendBird}
/>

```

becomes

```
msgid " Care Navigator <0>#</0>"
msgid_plural " Care Navigator <1>#</1>"
msgstr[0] " Care Navigator <0>#</0>"
[...]

PR #760, opened 2020-10-15
fix: trailing whitespace

semoal:
Fixes https://github.com/lingui/js-lingui/issues/747

This line `.replace(keepSpaceRe, " ")` is creating the whitespace, probably that regex could be reimplemented to don't do that change on this cases, but with the new regex this issue is fixed


tricoder42:
Nice, thanks a lot! 🥇

PR #762, opened 2020-10-17
fix: fix children with same key, uniqueId

### LINKED ISSUE #757: React "two children with same key" warning from `<Trans>` macro
**Describe the bug**
When using the `<Trans>` macro, if the language string contains two instances of the same variable, _and_ that variable's value is a JSX element, Lingui generates two sibling elements with the same `key` prop. This causes React to complain, because the `key` prop is required to be unique for sibling elements.

**To Reproduce**
Here is a Storybook story that reproduces the issue:

```
import React from 'react';
import { Trans } from '@lingui/macro';
import { storiesOf } from '@storybook/react';
import { I18nProvider } from '@lingui/react';

const foo = <span>bar</span>;

storiesOf('lingui', module).add('two children with the same key', () => (
  <I18nProvider language="en">
    <Trans>
      {foo} {foo}
    </Trans>
  </I18nProvider>
));
```

**Expected behavior**
The story displays "bar bar" when viewed in the browser, with no warnings or errors in the JS console.

**Actual behavior**
The story displays correctly in the browser, but it generates React elements with the same key of `0`, and React prints a warning about that to the JS console:

```
Warning: Encountered two children with the same key, `0`. Keys
should be unique so that components maintain their identity across
updates. Non-unique keys may cause children to be duplicated
and/or omitted — the behavior is unsupported and could change in a
future version.
    in Render (created by Trans)
    in Trans (created by I18n)
    in I18n (created by WithI18n)
    in WithI18n (at lingui.stories.tsx:10)
    in I18nProvider (at lingui.stories.tsx:9)
    in storyFn
    in ErrorBoundary
```
When the rendered components are viewed using the React Dev Tools browser plugin, I can see this hierarchy of components:

* `WithI18n`
    * `I18n`
        * `Trans`
            * `Render`
                * `span key="0"`
                * `span key="0"`

**Additional context**
- jsLingui version 2.9.2
- Babel version 7.9.0
- Create React App (react-scripts 3.4.3)

semoal:
Fixes #757 

It's using a simpe lodash uniqueId function, just a const to keep the past id's 👍🏻 
I've introduced a little test to avoid a regression to this

### DIFF
--- a/packages/react/src/format.ts
+++ b/packages/react/src/format.ts
@@ -68,7 +68,7 @@ function formatElements(
     tree.push(
       React.cloneElement(
         element,
-        { key: uniqueId("$lingui$") },
+        { key: uniqueId() },
 
         // format children for pair tags
         // unpaired tags might have children if it's a component passed as a variable
@@ -106,7 +106,7 @@ function getElements(parts) {
   )
 }
 
-function uniqueId(prefix = "$lingui$") {
+function uniqueId(prefix: "$lingui$" = "$lingui$") {
   if (!idCounter[prefix]) {
     idCounter[prefix] = 0
   }
--- a/packages/react/src/format.ts
+++ b/packages/react/src/format.ts
@@ -3,7 +3,6 @@ import React from "react"
 // match <0>paired</0> and <1/> unpaired tags
 const tagRe = /<(\d+)>(.*?)<\/\1>|<(\d+)\/>/
 const nlRe = /(?:\r\n|\r|\n)/g
-const idCounter = {}
 
 // For HTML, certain tags should omit their close tag. We keep a whitelist for
 // those special-case tags.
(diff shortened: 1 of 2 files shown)


packages/react/src/format.ts
```diff
@@ -105,4 +106,14 @@ function getElements(parts) {
...
+function uniqueId(prefix: "$lingui$" = "$lingui$") {
+  if (!idCounter[prefix]) {
+    idCounter[prefix] = 0
+  }
+
+  const id =++idCounter[prefix]
+
+  return `${prefix}_${id}`
+}
+
```
tricoder42, line 118: I would prefer simple closure before global variable:

```suggestion
const makeCounter = (count = 0, prefix = "") => () => `${prefix}${count++}`
```

Then we can create a counter inside `formatFunction`:

```js
const uniqueId = makeCounter(0, '$lingui$')
```

I think it should be safe to create the counter on every call of `formatElements`. The problem was that elements without an index (variables) were getting the same id.

tricoder42 (commented):
Good idea, just one suggestion 👇

packages/react/src/format.ts
semoal: Yes, absolutly, i've changed to match your suggestion!!

semoal:
Changes made @tricoder42, we can :shipit:

tricoder42:
@semoal Pretty good! Thanks 🚀

PR #763, opened 2020-10-17
fix: render from prop 

### LINKED ISSUE #607: `render` in `Trans` should be a renderProp
**Describe the bug**
Let's pick example from the docs
```js
<Trans render={({ translation }) => <Icon label={translation} />}>
```

What it is "expected" to be? It's _ducks_ like a render prop, so it would __just__ render an `Icon` with `translation`

What it is? `React.createElement(render, { translation: value })`, so it would be a new, unique `Element` __every render__.

The bug is: that's just not usable at all. It shall be render prop, and could be it, without any breaking compatibity.

**To Reproduce**

<Trans render={({ translation }) => <Icon label={translation} />}>

**Expected behavior**

Should not remount internal component every render

Both v2 and v3 are affected.

** Proposed fix **

https://github.com/lingui/js-lingui/blob/next/packages/react/src/Trans.tsx#L77
```js
  // Component: (props) => <a title={props.translation}>x</a>
  return React.createElement(render as React.ElementType<TransRenderProps>, {
    id,
    translation,
    message
  })
```
👇👇👇👇
```js
  // Component: (props) => <a title={props.translation}>x</a>
  return render({
    id,
    translation,
    message
  })
```

semoal:
Fixes #607

### DIFF
--- a/packages/react/src/Trans.tsx
+++ b/packages/react/src/Trans.tsx
@@ -10,7 +10,7 @@ export type TransRenderProps = {
   message?: string | null
 }
 
-export type TransRenderType = React.ComponentType<TransRenderProps>;
+export type TransRenderType = React.ComponentType<TransRenderProps> | React.ElementType<TransRenderProps>;
 
 export type TransProps = {
   id: string
--- a/packages/react/src/Trans.tsx
+++ b/packages/react/src/Trans.tsx
@@ -10,9 +10,8 @@ export type TransRenderProps = {
 }
 
 export type TransRenderType =
-  | string
-  | React.ElementType<TransRenderProps>
-  | React.ReactElement
+  React.FunctionComponent<TransRenderProps> |
+  ((o: TransRenderProps) => React.ComponentType<TransRenderProps>);
 
 export type TransProps = {
   id: string
@@ -20,7 +19,7 @@ export type TransProps = {
   values: Object
   components: { [key: string]: React.ElementType | any }
   formats?: Object
-  render?: TransRenderType
+  render?: (o: TransRenderProps) => TransRenderType
 }
 
 export function Trans(props: TransProps) {
@@ -75,7 +74,7 @@ export function Trans(props: TransProps) {
   }
 
   // Component: (props) => <a title={props.translation}>x</a>
-  return React.createElement(render as React.ElementType<TransRenderProps>, {
+  return render({
     id,
     translation,
     message,
--- a/packages/react/src/Trans.tsx
+++ b/packages/react/src/Trans.tsx
@@ -1,4 +1,4 @@
-import React, { DOMElement } from "react"
+import React from "react"
 
 import { useLingui } from "./I18nProvider"
 import { formatElements } from "./format"
@@ -11,7 +11,7 @@ export type TransRenderProps = {
 
 export type TransRenderType =
   string |
-  DOMElement<any, any> |
+  React.DOMElement<TransRenderProps, any> |
   React.ReactElement<TransRenderProps> |
   React.FunctionComponent<TransRenderProps> |
   ((o: TransRenderProps) => TransRenderType);
--- a/packages/react/src/Trans.tsx
+++ b/packages/react/src/Trans.tsx
@@ -82,9 +82,9 @@ export function Trans(props: TransProps) {
     });
   } else if (Component) {
     // Component: component={Text}
-    return <Component>{translation}</Component>
+    return <Component>{translation}</Component>;
   } else {
-    return DefaultComponent ? <DefaultComponent>{translation}</DefaultComponent> :<>{translation}</>;
+    return DefaultComponent ? <DefaultComponent>{translation}</DefaultComponent> : <>{translation}</>;
   }
 }
 
--- a/packages/react/src/Trans.tsx
+++ b/packages/react/src/Trans.tsx
@@ -10,7 +10,9 @@ export type TransRenderProps = {
   message?: string | null
 }
 
-export type TransRenderType = React.ComponentType<TransRenderProps> | React.ElementType<TransRenderProps>;
+export type TransRenderType =
+  | React.ComponentType<TransRenderProps>
+  | React.ElementType<TransRenderProps>
 
 export type TransProps = {
   id: string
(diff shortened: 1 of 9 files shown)


packages/react/src/Trans.tsx
```diff
@@ -10,17 +10,16 @@ export type TransRenderProps = {
 }
 
 export type TransRenderType =
-  | string
-  | React.ElementType<TransRenderProps>
-  | React.ReactElement
```
tricoder42, line 15: We need to refactor this function a bit. First of all, if we remove support for `string`, `React.ElementType<TransRenderProps>` and `React.ReactElement` we also need to remove the if-branches which handles these cases and related tests:

https://github.com/lingui/js-lingui/blob/7b0ff4cba08d9b2d6d13e5c2632f8a84b20904b7/packages/react/src/Trans.tsx#L66-L74

Also, this is going to be a breaking change which we need to document.

---

Finally, I don't think we can support `React.FunctionComponent<TransRenderProps>` here. Since we're calling the function directly, I believe fucntion component won't work as ithey might contain hooks and possible side effects.

semoal: Alright, so i'll take a look more in depth with your notes

semoal: I did some changes Tomáš, to keep the compatibility with string and added a new `if` to check if the render prop is a component or a class, so we use the current method `React.createElement`, and if the user is just using `<Trans render={(props) => <Element {...props />`, we use the `render({})` alternative

I've added some testing to check that the is still rendering the content correctly

packages/react/src/Trans.tsx
```diff
@@ -69,19 +73,43 @@ export function Trans(props: TransProps) {
...
+      message
+    })
+  } else if (React.isValidElement(render)) {
     // Element: <p className="lear' />
     return React.cloneElement(render, {}, translation)
   }
 
   // Component: (props) => <a title={props.translation}>x</a>
-  return React.createElement(render as React.ElementType<TransRenderProps>, {
+  return render({
```
theKashey, line 92: any function is `isFunctionalComponent`, so you will never execute this line

semoal: I tough only functions that returns "internally" React.createElement are functional components, we cant render a functional component as a simple function, Wont work correctly

tricoder42:
Apparently, it's impossible to distinguish functional component from a render prop function:

```js
function MyComponent({ translation }) {
   const value = useMySideEffect()
   // do something
   return translation
}

function ExampleComponent() {
  return ( 
    <Trans render={MyComponent}>This is rendered in custom cumponent</Trans>
  )
}

function ExampleRenderProp() {
  return ( 
    <Trans render={({ translation }) => <Component>{translation}</Component>}>This is rendered using render prop function</Trans>
  )
}
```

So, we need to simplify the API. I believe there's no need to support every possible option.

1. Built-in components - I would say we can omit this usecase.

```js
<Trans render="h1">Heading</Trans>

// Alternative
<h1><Trans>Heading</Trans></h1>
```

2. Custom components - I believe we should keep this functionality. Maybe add a `component` prop to distinguish between a render prop and a custom component? But see my point about `4. Render prop` below

```js
import { Text } from "react-native";

<Trans render={Text}>Works in React Native</Trans>

// This is in real world used as a default prop of I18nProvider

<I18nProvider defaultRender={Text}>
   <Trans>Works in React Native</Trans>
</I18nProvider>
```

Also, custom component are handy if you want to e.g. create UI for in-context translations for development.

3. Custom element — Not sure if this is actually useful. It could be probably superseeded by `4. render prop` and removed.

```js
<Trans render={<MyComponent prop={42} />}>Shortcut for adding props to custom components</Trans>
```

4. Render prop — Do we actually need it? What is it required for?

```js
<Trans render={({ translation }) => <img alt={translation} />}>Image title</Trans>

// Alternative
<img alt={t`Image title`} />
```

---

## Proposal

1. Keep custom component functionality, but move it to `component` prop:

```js
<Trans component={CustomTranslation}>Works in React Native</Trans>
```

2. Rename `defaultRender` to `defaultComponent` in `I18nProvider`. Also add it to migration docs, as it's a breaking change.

3. Drop support for 1. builtin components and 3. custom elements.

4. What to do with a render prop functions? Do we need it? Is there a usecase which can't be covered with js macros (`t` etc)?

theKashey:
1. 👍 . Would be consistent with, for example, react-router API. 
2. ok
3. 🤷‍♂️
4. Well, babel can distinguish "component" and "renderProp" - first one is a separate variable, second is "inlined"
4+. Can eslint help here?

tricoder42:
> Well, babel can distinguish "component" and "renderProp" - first one is a separate variable, second is "inlined"
4+. Can eslint help here?

I don't know what you mean. My question is if we actually need a support for render prop function if all possible usecases can be covered by using core functions.

```js
// In other words, is there a usecase similar to this:
<Trans render={({ translation }) => <img alt={translation} />}>Image title</Trans>

// Which can't be replaced with a simplified version like this?
import { t } from "@lingui/macro"
<img alt={t`Image title`} />
```

semoal:
Probably we should just keep the "general" use cases that people use in React.

Example of `react-router`:
```jsx
    component?: React.ComponentType<RouteComponentProps<any>> | React.ComponentType<any>;
    render?: (props: RouteComponentProps<any>) => React.ReactNode;
```

Task I considered:

- [ ] Keep custom component functionality, but move it to component prop:
```<Trans component={CustomTranslation}>Works in React Native</Trans>```
- [ ] Rename defaultRender to defaultComponent in I18nProvider. Also add it to migration docs, as it's a breaking change.

- [ ] Drop support for 1. builtin components and 3. custom elements.

- [ ] Maintain `render={() => {}}` for being friendly with React ecosystem

I can start working on it if you're comfortable with it

tricoder42:
@semoal Sure, sounds reasonable! Thanks 👍

semoal:
I think all is ready, I'll take a re-read to documentation if somewhere has something confusing.

Anyways, I have to review the CRA setup so probably I'll take a read to everything.. =)

tricoder42:
@semoal Thanks a lot! I've refactored the `Trans` component a bit and added all new features to migration docs. Looks pretty good 👍

PR #767, opened 2020-10-20
[766] compile should return true correctly if everything was fine

lucax88x:
return true is currently inside forEach by mistake, moved outside of it


tricoder42:
Perfect, thanks 👍

Issue #527, opened 2019-06-22
doc: not working link to rendering reference


landsman:
I wanted to send PR with fix but I can not find the right article:
https://github.com/lingui/js-lingui/blob/master/docs/ref/macro.rst#render

tricoder42:
Right article for what? Could you be more specific please?

landsman:
For not working link: https://github.com/lingui/js-lingui/blob/master/docs/ref/rendering-translations

tricoder42:
Ah, interesting. The link should point to this section: https://github.com/lingui/js-lingui/blob/master/docs/ref/react.rst#rendering-of-translations

landsman:
@tricoder42 but this is not react doc I think. This is vanilla-js solution, right?

tricoder42:
Not really, rendering of translations is related to React.

Issue #484, opened 2019-04-02
Ignore compiled catalogs?


dcorb:
In the documentation says that is safe to ignore the *compiled* catalogs ? `locale/**/*.js `https://lingui.js.org/guides/excluding-build-files.html

It doesn't make sense to me. I understand is ok to ignore ` _build` folder, but why to ignore the compiled `.js ` catalog that should be normally should be committed and loaded in production?

zenyr:
I _used to_ include .js catalogs in git but soon I found out that I want to build a fresh copy of .js catalogs from external .json files on each build anyway.

filipcro:
It is an auto generated file and you need to recreate it after each change anyway. You don't lose anything if you delete this file, because original translation files are enough to recreate it.

If you commit it you will produce conflicts, which somebody will need to solve, in this minimized and transpired (basically unreadable to human) file. Do you really want this? And this will happen often on project with multiple developers. Just think how often you add new translations to catalog. 

Also if you have workflow where you for some reason commit builded files to git, all files for deployment, including translations will be in root `build` folder anyway.

dcorb:
@zenyr  @filipcro  thanks for your answers. It has sense to add the compilation step to the CI
And it's true that is creating git conflicts sometimes (easily dismissed re-generating the compiled catalog)

But the doc says it's "safe" to ignore them, without giving explanations... if you git-ignore it, and you deploy your git repo content, translations will not work (aka, it's not "safe"). You need to run that compilation step

I consider only correct to say "it's safe" to ignore the _build folder. The compiled .js files don't live inside the build folder.

![image](https://user-images.githubusercontent.com/881333/56476484-f6d1cb80-64a0-11e9-84b4-39a175cce0e8.png)

filipcro:
If you have CI make it run compilation.

tricoder42:
Hi @dcorb, I consider compiled catalogs as build artifacts. I don't include production bundle (e.g. next.js `_next` folder or CRA `build` folder) in git, so I don't include `messages.js` either.

The documentation could be improved though. Would you like to send a PR?

Issue #415, opened 2018-12-12
No messages extracted outside of react


diegolacarta:
No messages get extracted when using:

```
import i18n from './utils/i18n'
import {t} from '@lingui/macro';

i18n._(t`hello`)
```

They get extracted when using it like:
```
i18n._("hello")
```

They also get extracted as expected when using a react component:

```
import {Trans} from '@lingui/react';

<Trans>Hello</Trans>
```

any ideas?

diegolacarta:
"@lingui/cli": "2.7.2",
    "@lingui/loader": "2.7.2",
    "@lingui/macro": "2.7.2",
"@lingui/react": "2.7.2",

tricoder42:
Could you please also post your Babel config? Do you have `babel-plugin-macros` enabled?

diegolacarta:
```
{
  "env": {
    "local": {
      "presets": ["@babel/preset-react", "@lingui/babel-preset-react"],
      "plugins": [
        ["@babel/plugin-syntax-dynamic-import"],
        ["macros"],
        [
          "emotion",
          {
            "sourceMap": true,
            "autoLabel": true
          }
        ],
        [
          "react-intl",
          {
            "messagesDir": "./i18n/messages/",
            "extractSourceLocation": true
          }
        ]
      ]
    }
  }
}
```

react-intl in there just while transitioning away from it

tricoder42:
That looks good. Could you also please show how babel transforms the file with i18n? `babel src/file.with.i18n.js`

Finally, does extract work for `Trans` imported from `@lingui/macro` package?

tricoder42:
There's also an example repo with js-macros (no React): https://github.com/lingui/js-lingui/tree/master/examples/vanilla-js Maybe it would help you find the root cause.

diegolacarta:
> That looks good. Could you also please show how babel transforms the file with i18n? `babel src/file.with.i18n.js`
> 
> Finally, does extract work for `Trans` imported from `@lingui/macro` package?

I'm using typescript, following the setup from https://github.com/lingui/js-lingui/blob/master/docs/guides/typescript.rst

I could debug at which point babel plugin is not doing what it should, or if you could point me into where to look that'd help

Trans doesn't work when imported from @lingui/macro:
`<Trans>random text</Trans>`
...
```
> Missing message ID, skipping.
> <Trans>random text</Trans>
```

tricoder42:
Right, so now we now the macros doesn't work.

[...]

Issue #388, opened 2018-11-08
Documentation: All catalogs including sourceLocale must be compiled. Default messages aren't available in production.


yalishizhude:
**Describe the bug**
`@lingui/core/cjs/core.production.min.js` Found a bug in this file.

```
i18n._('My name is {name}', {name: 'yalishizhude'}) 
//  as expected, it should be  'My name is yalishizhude'.
// but in fact it is 'My name is {name}'
```

**To Reproduce**
**Remember, must use the production js,development is ok.**
```
import { setupI18n } from '@lingui/core';
const i18n = setupI18n();
i18n._('My name is {name}', {name: 'yalishizhude'}) 
```

**Expected behavior**
My name is yalishizhude

**Additional context**
My name is {name}

- jsLingui version ``2.7.0``
- Babel version ``6.26.0``
- framework you use (Create React App)

tricoder42:
This might be similar issue as in https://github.com/lingui/js-lingui/issues/385. Could you please try to load compiled catalog with translated messages?

```js
import { setupI18n } from '@lingui/core';

// change location to your source catalog
import catalogEn from '@lingui/loader!locale/en/messages.po'

const i18n = setupI18n();
i18n.load({ en: catalogEn })
i18n._('My name is {name}', {name: 'yalishizhude'}) 
```

yalishizhude:
Yes, I have my own catalogs.The almost full code as follows:
```
// i18nFactory.ts
import { Catalogs, setupI18n } from '@lingui/core';

export default (catalogs: Catalogs) => {
  const i18n = setupI18n({
    catalogs
  });
  return i18n;
};
```

```
// i18n.ts
import en from '../locale/en';
import zh from '../locale/zh';
import factory from './i18nFactory';

export default factory({ en, zh });
```

```
// en.ts
const en = {
  ...
  'rule.same': '{srcFieldName} is different from {destFieldName}',
  ...
}
```

I cannot precompiled the catalogs, because as the codes show, `srcFieldName` and `destFieldName` are dynamic decided by variables(some logic or some requests' response, something like this), not static.

tricoder42:
@yalishizhude That's fine. You need to compile catalogs especially for messages with formatting.

Use `lingui extract` to get message catalogs and then `lingui compile` to get final js module. Don't write it manually.

Let me know if it helps

yalishizhude:
It doesn't work though I tried two diffrent ways~
[...]

PR #769, opened 2020-10-20
docs: refactored docs to match latest spec

semoal:
Will close some issues listed here:

#415 
#527 
#484 
#388


semoal:
Not finished yet, I'll finish in 1-2 hours max: 

- [ ] Check how missing param is working on v3 and replace the current doc
- [ ] Check dynamic loading of translations of v3 and chang
- [ ] Check Webpack loader docus
- [ ] Trans misspellings

docs/ref/react.rst
```diff
@@ -325,10 +325,16 @@ for debugging:
...
+   import { en } from 'make-plural/plurals'
+
+   i18n.loadLocaleData('en', { plurals: en })
+   i18n.load('en', messages)
+   i18n.activate('en')
 
    const App = ({ language} ) => {
         return (
-            <I18nProvider language={language} missing="🚨">
+            <I18nProvider i18n={i18n}>
```
tricoder42, line 337: I believe `missing` prop has been removed completely and this whole paragraph can be removed as well.

tricoder42: Ah, nevermind, I see it's still in your todo list 👍

semoal: Yes! :) I'll ping you once finished

semoal:
Screenshot 2020-10-20 at 16 39 34
Built a cool CRA with all the functions that Lingui uses, I'll rewrite to handle also a Hoc for dynamic loading.

@tricoder42 how is it handled on v3? Should I still use @lingui/loader?
Can we use i18n.on("change", () => {})?

examples/create-react-app/src/i18n.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const locales = {
+  en: "English",
+  cs: "Česky",
+};
+
+export const defaultLocale = "en";
+
+export const i18n = setupI18n();
+i18n.load({
```
semoal, line 15: @tricoder42 can't  we load plurals on this method?

tricoder42:
Take a look at next.js example https://github.com/lingui/js-lingui/blob/next/examples/next-js/lingui-example/i18n.ts. I'm using a custom function, which dynamicaly loads catalog, feeds i18n object and activates locale. I believe you could use `@lingui/loader` inside the dynamic import without problems.

Also, feel free to replace create-react-app example with your version.

examples/create-react-app/src/i18n.ts
tricoder42: No, this only handles messages.

Also, please don't use `setupI18n` in tutorial. That's for advanced use and it requires additional configuration in `lingui.config.js`. Simply import `i18n` from `@lingui/core`

semoal: Great, didn't know that I'll add that to the documentation

semoal:
@tricoder42 Pull request ready mate :)

There's only one question left:

```jsx
  date(value: string | Date, format: Intl.DateTimeFormatOptions): string {
    return date(this.locales || this.locale, format)(value)
  }

  number(value: number, format: Intl.NumberFormatOptions): string {
    return number(this.locales || this.locale, format)(value)
  }
```

I saw this methods options are not optional, but I think it should be optional if you're conformtable with base config of i18n methods, no?

tricoder42:
@semoal Yes, they should definitely be optional. Good point 👍 

And thanks for the PR! I'm gonna check it later tonight. Good work 🚀

semoal:
> @semoal Yes, they should definitely be optional. Good point 👍
> 
> And thanks for the PR! I'm gonna check it later tonight. Good work 🚀

I'll create other pull request to keep the versioning tags =)

tricoder42:
It looks perfect. Let's merge it so we can review it online. Thank you for tackling this 🚀

PR #771, opened 2020-10-20
Allow number properties in plurals macro

### LINKED ISSUE #770: Plural macro does not support _0
As per [beware of zeros](https://js-lingui-git-next.lingui-js.now.sh/tutorials/react.html#beware-of-zeroes) in linguijs' docs, some languages do not have a zero plural form, requiring the prop `_0` to be used instead of `zero`.

While using the [plural JS macro](https://js-lingui-git-next.lingui-js.now.sh/ref/macro.html#plural) it seems there is no property `_0` available, but only [the following](https://github.com/lingui/js-lingui/blob/bddb43fe07d1dd5cdf7cd764147037ef7ee95171/packages/macro/index.d.ts#L10):

```typescript
export type ChoiceOptions<T = string> = {
  offset?: number
  zero?: T
  one?: T
  few?: T
  many?: T
  other?: T
}
```

Considering there's no other alternative: support the property `_0` while using the plural JS macro.

plourenco:
Closes #770

### DIFF
--- a/packages/macro/index.d.ts
+++ b/packages/macro/index.d.ts
@@ -8,6 +8,7 @@ export function t(
 ): string
 
 export type ChoiceOptions<T = string> = {
+  [digit: number]: T,
   offset?: number
   zero?: T
   one?: T


tricoder42:
Perfect, thank you! 👍

PR #772, opened 2020-10-20
Make clear that v3 migration pre-release

Bertg:
I have wrongly tried to apply changes from this document to my code, as I assumed that the latest version is v3.
Ideally there should be a operate site for the v2 and v3 documentation.

### DIFF
--- a/docs/releases/migration-3.rst
+++ b/docs/releases/migration-3.rst
@@ -2,6 +2,9 @@
 Migration guide from 2.x to 3.x
 ********************************
 
+> ⚠️ LinguiJS v3 is in pre-release.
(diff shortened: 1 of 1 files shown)


tricoder42:
Good point! Thank you 👍

tricoder42:
Actually, I'll do it the other way round. I'm gonna remove migration docs from `master` branch which is deployed at https://lingui.js.org and instead keep them only in `next` branch which is available here https://js-lingui-git-next.lingui-js.now.sh/. I'll merge your commit to `next` branch

Bertg:
Even better! 
I didn't want to suggest such a big change; but this is the best solution indeed.

PR #773, opened 2020-10-20
fix: required options to date, number macro

semoal:
We add optional params to date and number cuz they aren't required in the native functions.


tricoder42 (approved):

tricoder42:
Awesome, thanks 👍

PR #775, opened 2020-10-22
hotfix: i18n props and exported component

semoal:


tricoder42 (approved):

tricoder42:
Pretty good! Thanks 👍

renchap:
Sorry I did not had time to have a look faster than @tricoder42 but I think you also need to export the type `withI18nProps` so TS users can use it (like in version 2)

tricoder42:
@renchap It is exported https://github.com/semoal/js-lingui/blob/b31883adf762db73866a7550701b916f89eacc39/packages/react/src/I18nProvider.tsx#L51

Maybe we could keep the `export` on the same line as a type (I missed that), but functionaly it's ok 👍

renchap:
I missed this line, all is fine then 🙂

PR #778, opened 2020-10-24
Update macro reference to latest version

tricoder42:
I've updated [`ref/macro`](https://js-lingui-git-feature-docs-macro.lingui-js.now.sh/ref/macro.html) docs to reflect latest changes in v3. @semoal, I noticed there were lot of wrong examples (especially macros wrapped in `i18n._`). Could you please review changes just in case it affects your codemod?

Here's the preview: https://js-lingui-git-feature-docs-macro.lingui-js.now.sh/ref/macro.html

### DIFF
--- a/docs/ref/macro.rst
+++ b/docs/ref/macro.rst
@@ -93,11 +93,10 @@ Examples of JS macros
 +-------------------------------------------------------------+--------------------------------------------------------------------+
 | .. code-block:: js                                          | .. code-block:: js                                                 |
 |                                                             |                                                                    |
-|    plural({                                                 |    i18n._(/*i18n*/{                                                |
-|      value: count,                                          |      id: "{count, plural, one {Message} other {Messages}}",        |
-|      one: "Message",                                        |      values: { count }                                             |
-|      other: "Messages"                                      |    })                                                              |
-|    })                                                       |                                                                    |
+|    plural(count, {                                          |    i18n._(/*i18n*/{                                                |
+|       one: "Message",                                       |      id: "{count, plural, one {Message} other {Messages}}",        |
+|       other: "Messages"                                     |      values: { count }                                             |
+|    })                                                       |    })                                                              |
 +-------------------------------------------------------------+--------------------------------------------------------------------+
 | .. code-block:: js                                          | .. code-block:: js                                                 |
 |                                                             |                                                                    |
@@ -310,10 +309,10 @@ are transformed automatically:
 .. code-block:: jsx
 
    import { plural } from "@lingui/macro"
-   const message = i18n._(plural(count, {
+   const message = plural(count, {
       one: `${name} has # friend`,
       other: `${name} has # friends`
-   }))
+   })
 
    // ↓ ↓ ↓ ↓ ↓ ↓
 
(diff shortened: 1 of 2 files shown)


docs/ref/macro.rst
```diff
@@ -3,86 +3,153 @@
...
-|                                                             |    }                                                               |
-|                                                             |                                                                    |
+|                                                             |    })                                                              |
 +-------------------------------------------------------------+--------------------------------------------------------------------+
 | .. code-block:: js                                          | .. code-block:: js                                                 |
 |                                                             |                                                                    |
-|    plural({                                                 |    /*i18n*/{                                                       |
+|    plural({                                                 |    i18n._(/*i18n*/{                                                |
```
semoal, line 96: on the new method is not like this?: 

```
js
plural(count, {
  one: "Message",
  other: "Messages"
})
```

On the codemod I have a method to rewrite this macro from { value: count ....} to (count, {...})

semoal (approved):

docs/ref/macro.rst
```diff
@@ -174,87 +279,149 @@ plural
...
 
-Second argument (optional) specifies date format.
+``selectOrdinal`` macro is similar to :jsmacro:`plural` but instead of using
+cardinal plural forms it uses ordinal forms:
 
 .. code-block:: jsx
 
-   import { setupI18n } from "@lingui/core"
-   import { t, date } from "@lingui/macro"
+   import { plural } from "@lingui/macro"
```
semoal, line 381: shouldnt be selectOrdinal?

docs/ref/macro.rst
```diff
@@ -174,87 +279,149 @@ plural
...
 
-Second argument (optional) specifies number format.
+``select`` macro works as a switch statement — it select one of the forms
+provided in ``options`` object which key matches exactly ``value``:
 
 .. code-block:: jsx
 
-   import { setupI18n } from "@lingui/core"
-   import { t, number } from "@lingui/macro"
+   import { plural } from "@lingui/macro"
```
semoal, line 411: shouldnt be select?

semoal (changes requested):

docs/ref/macro.rst
```diff
@@ -174,87 +279,149 @@ plural
...
+This time :jsmacro:`t` macro isn't required as template strings
+are transformed automatically:
 
-   // t macro isn't required for nested messages,
-   // template strings are transformed automatically.
-   const vars = i18n._(plural(count, {
+.. code-block:: jsx
+
+   import { plural } from "@lingui/macro"
+   const message = i18n._(plural(count, {
```
semoal, line 313: is this i18n._. required?, on the codemod i'm replacing all the i18n._(plural()), for just plural(value, {})

semoal (changes requested):

docs/ref/macro.rst
tricoder42: You're right, I missed this code example 👍

docs/ref/macro.rst
tricoder42: No, it's not 👍

docs/ref/macro.rst
tricoder42: Yes :)

docs/ref/macro.rst
tricoder42: Yes, it should 👍

tricoder42 (commented):
@semoal Thanks for your comments! I've just pushed requested changes

Issue #656, opened 2020-04-14
[next] Refactor extraction stats output


FredyC:
Currently, stats are commented out due to the complexity of multiple catalogs. 

https://github.com/lingui/js-lingui/blob/03ce90086a698e0cc030d188c0a7e2cb4d0b787b/packages/cli/src/lingui-extract.ts#L49

We can discuss here how it should look like to be actually useful. Alternatively, it might be worth to keep extract command simple with some short summarization and instead add a new command that would output stats based on eg. catalog name.

semoal:
Uh, I tough that the script wasnt working properly, because I was used to see the stats.
Whats the problem? I could work on it, after the codemod

FredyC:
AFAIK it was commented out because of a change in how catalogs work. I don't really know specifics.

tricoder42:
Yes, now with multiple catalogs it's a bit more complex to calculate the stats. I didn't have time to fix it.

Regarding the UI, I guess we could display the overal stats as before. Not sure if stats per catalog would be anyhow helpful

semoal:
> Yes, now with multiple catalogs it's a bit more complex to calculate the stats. I didn't have time to fix it.
> 
> Regarding the UI, I guess we could display the overal stats as before. Not sure if stats per catalog would be anyhow helpful

Tomáš I can work on it if you give me some guidelines, we are close enough to do some little corrections to the codemod so I can do it in parallel.

I'll write some docs on the website about the codemod an his usage in the migration-3.rst if you're agree with =)

tricoder42:
> Tomáš I can work on it if you give me some guidelines, we are close enough to do some little corrections to the codemod so I can do it in parallel.

Well, you need to calculate stats for each catalog, combine them together and then print to console. I don't think it's complex or difficult, I just didn't have time to tackle it.

https://github.com/lingui/js-lingui/blob/bcc37800c57638022bfe492b4ba58011ebe37879/packages/cli/src/lingui-extract.ts#L34-L49

As I mentioned before I don't know if we should print stats for each catalog or just the totals. I belive the totals could be enough. I've always used stats to check missing translations

> I'll write some docs on the website about the codemod an his usage in the migration-3.rst if you're agree with =)

Absolutely! Go ahead 👍

semoal:
Alright, lgtm I'll start later on

PR #779, opened 2020-10-25
fix: refactored stats logs 

semoal:
I'm not sure what's the correct way of handling this, I think the problem of users using lingui in a monorepo where "en" appears by each catalog it's solved in this way, we log for every catalog his stats

Feel free to rquest any changes or some improvements, i'm still pretty noobie about core aspects.

### DIFF
--- a/packages/conf/index.d.ts
+++ b/packages/conf/index.d.ts
@@ -15,7 +15,9 @@ export declare type LinguiConfig = {
     extractBabelOptions: Object;
     fallbackLocale: string;
     format: CatalogFormat;
+    prevFormat: CatalogFormat;
     formatOptions: CatalogFormatOptions;
+    localeDir: string;
     locales: string[];
     catalogsMergePath?: string;
     orderBy: OrderBy;
--- a/packages/conf/index.d.ts
+++ b/packages/conf/index.d.ts
@@ -15,9 +15,7 @@ export declare type LinguiConfig = {
     extractBabelOptions: Object;
     fallbackLocale: string;
     format: CatalogFormat;
-    prevFormat: CatalogFormat;
     formatOptions: CatalogFormatOptions;
-    localeDir: string;
     locales: string[];
     catalogsMergePath?: string;
     orderBy: OrderBy;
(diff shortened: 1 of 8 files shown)


packages/conf/index.d.ts
```diff
@@ -15,7 +15,9 @@ export declare type LinguiConfig = {
     extractBabelOptions: Object;
     fallbackLocale: string;
     format: CatalogFormat;
+    prevFormat: CatalogFormat;
     formatOptions: CatalogFormatOptions;
+    localeDir: string;
```
tricoder42, line 20: `localeDir` is an obsolete config replaced by `catalogs`. `prevFormat` is obsolete as well.  Why you've put them back? If they're required somewhere, we might fix that code instead.

tricoder42 (commented):
Looks good to me. Just one comment about obsolete configuration

packages/conf/index.d.ts
semoal: There are some tests using this configuration, how i passed the Partial<LinguiConfig> to mockConfig function, some tests were failing because this type doesnt exists. 

I'll remove this types from the base and i'll extend the type only for this tests cases? or just remove this tests cases that are deprecated?

tricoder42: Hmm, good question. `LinguiConfig` type which is used by `Catalog` is a different than `LinguiConfig` in `conf` package: https://github.com/lingui/js-lingui/blob/aff1d092fc2f2848eefb129b931c488e3232fac9/packages/conf/src/index.ts#L22-L36

I've searched the codebase for `localeDir` and I've found it's still being used in babel plugin... Let's merge this one and figure out how to clean the codebase in next PR

semoal: Perfect for me, it's a pretty easy task (for first constributors) if we document correctly on an independent issue what's the goal

tricoder42: I honestly don't know how it's possible that extract plugin works :) I'll take a look when I have some time, but if it works, it's good enough

tricoder42:
Closes #656 

Thanks a lot @semoal. Once again, great job 🚀

PR #783, opened 2020-10-26
Keep decoded html entities

### LINKED ISSUE #780: Handling of HTML entities
As @renchap pointed out [few days ago](https://github.com/lingui/js-lingui/issues/334#issuecomment-714557445), LinguiJS handles HTML entities differently that you would expect from JSX:

```js
<p><Trans>&ldquo;entities&rdquo;</Trans></p>
// rendered as &ldquo;entities&rdquo;

<p>&ldquo;entities&rdquo;</p>
// rendered as “entities”
```

This is due to the fact that we're actually rendering strings. For example, this wouldn't render html entities either:

```js
<p>{"&ldquo;entities&rdquo;"}</p>
// rendered as &ldquo;entities&rdquo;
```

Even this:

```js
React.createElement("p", null, "&ldquo;entities&rdquo;")
// rendered as &ldquo;entities&rdquo;
```

I assume this isn't neither caused by React or Lingui, but rather Babel JSX plugin. When it encounters HTML entities in children, it somehow converts them to the actual characters.

One option is to pass the translation to `dangerouslySetInnerHTML` which I don't like. Second option is to convert HTML entities as babel does either by using a 3rd party lib like https://github.com/mathiasbynens/he#hedecodehtml-options or figure out how Babel does it. I believe we should convert it in compile step, so we still have the escaped entities in message catalogs.

tricoder42:
Closes #780 

@semoal There's a small change in migration docs. Does it affect your codemod? I guess not, but I wanted to check with yout.

### DIFF
--- a/docs/releases/migration-3.rst
+++ b/docs/releases/migration-3.rst
@@ -160,8 +160,8 @@ for a single locale or multiple catalogs at once.
 - command ``lingui init`` was removed
 - command ``lingui add-locale`` was removed
 
-Whitespace and HTML entities
-----------------------------
+Whitespace
+----------
 
 Whitespace handling in plugins had few bugs. By fixing them, there might be few
 backward incompatible changes. It's advised to run :cli:`extract` and inspect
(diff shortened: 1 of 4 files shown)


semoal:
Actually doesnt brake the codemod, but probably could be interesting to migrate if some users are using unscaped entities to the correct entities

I'll add that feature if you're okay with it 👍🏻

semoal (approved):

tricoder42:
Actually don't worry about it. In v2 it works exactly the same. I thought we need to change it but I've just reverted it back. If you don't have it in codemods, then it's fine 👍

PR #784, opened 2020-10-26
feat: introduced some initial client detectors

### LINKED ISSUE #577: Detect locale
User's locale should be detected automatically if possible, but also allow customization. The approach should work both on server and in browser and it should follow guidelines for [multilingual sites](https://support.google.com/webmasters/answer/182192?hl=en&ref_topic=2370587)

## Proposal

Websites which serve different content for each locale (e.g. Wikipedia) will always use just URL to detect the current locale.

Locale-adaptive apps (i.e. same content, but UI is internationalized, like GitHub, Twitter, Facebook) should detect locale from following sources:

On server, detect locale from:

1. URL (subdomain `en.example.com`, path `example.com/en/` or search param `example.com?l=en`). This is useful even for locale-adaptive apps for search engines, i.e. user's will always use `example.com` and customize locale through user's profile or cookies, but search engine will be advised to follow `example.com/?l=en` to crawl site in specific locale.
2. Cookie (for user's overrides)
3. [HTTP Accept-Language header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language)

In browser, detect locale from:

1. `<html lang>` attribute (if set by server). 

This should be the only option for server-side rendered pages. For client-side rendered pages, there are other options:

2. URL (same motivation as above)
3. Cookie (for user's overrides)
4. [navigator.language](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language) 

## Implementation

Each _detector_ `<please insert better name>` should be just a plain function returning a locale or null (if not found):

```
type Detector<Config extends {}> = (config: Config) => string | null

// just an example with pseudocode
function detectFromCookie({ name, expiration }) {
  if (typeof window !== "undefined") {
    // server side code
    if (has cookie) {
      return req.cookie[name]
    }
  } else {
    // client-side code
  }
}
```

Then we can chain them together to get the first match:

```
const locale = detect(
  detectFromURL(),
  detectFromCookie({ cookieName: "locale" }),
  detectFromRequest()
)
```

## Resources

- [Guide - Detecting active locale in LinguiJS] - I've started writing draft of guide, it's available in [`nxt/detect-locale`](https://raw.githubusercontent.com/lingui/js-lingui/6f58868e7dbb50fddb444c251bf963f5d2b9ecab/docs/guides/active-locale.rst) branch. Feel free to use it or throw it away, I just needed to figure out a possible approach.
- [fluent-langneg](https://github.com/projectfluent/fluent.js/tree/master/fluent-langneg) - once we have a list of requested locales, we might compare it with list of available locales to get the best match. Probably only useful when detecting locale from `navigator.language` or `Accent-Language` header.
- [i18next/language-detector](https://github.com/i18next/i18next-browser-languageDetector/blob/master/src/index.js) includes several strategies to detect locale. The implementation seems to be tighly coupled to i18next, but it could definitely serve as an inspiration.

## Checklist

- [ ] Create a new package `@lingui/detect-locale` or find existing solution in JS ecosystem.
- [ ] Document how to detect locale in different kind of apps (static vs locale-adaptive, SSR vs CSR)

semoal:
Will close #577

### Tasks

- [X] Detector fromUrl(parameter: string, location?) - `returns from the query-string params the correct value`
- [X] Detector fromCookie(key: string) - `returns from the cookie the stored value`
- [X] Detector fromStorage(key: string, { useSessionStorage: true }),  - ` recovers from localstorage or session storage if the options passed the corresponding value`
- [X] Detector fromNavigator() - `detects the locale via navigator.language || navigator.userLanguage (ie 11 comaptibility)`

### Pending
- [ ] Some server side detectors... (to be discussed since next-js is already working on it, probably cookies and headers will be cool)
- [x] Html tags (like html=en)
- [x] Path and subdomain parameters 
- [x] Write documentation

All is fully tested, inclusive the utils that are being used like cookie-getter and the query-string parser 

### Current api

```js
import { detect, fromUrl, fromStorage, fromNavigator } from "@lingui/detect-locale"

// can be a function with custom logic or just a string, `detect` method will handle it
const DEFAULT_FALLBACK = () => "en"

detect(
  fromUrl("lang"),
  fromStorage("lang"),
  fromNavigator(),
  DEFAULT_FALLBACK
)
```


tricoder42:
This is really great! I would release it as it is and add server detectors in later version. Great job 👍

semoal:
@tricoder42 node10 is failing cuz out of memory, not sure what's the problem 

I'm confortable with lookin in the future to server side solutions, on all the detectors we implemented probably we cover the needs of the 80% users, once next-js gives the guideline we'll follow them :)

So, I think this feature is done, i don't know if i have to do something else to the package to be deployable

tricoder42:
Perfect! Let's ship it 🚢

PR #786, opened 2020-10-27
fix: import type declaration overrides @lingui/react

### LINKED ISSUE #777: Babel macros do not works with TS "import type" before the macro import
**Describe the bug**

When using `import type { … } from "@lingui/react"` before `import { Trans } from "@lingui/macro"` the macro does not use an import for `Trans`.
This results in an obscure `Uncaught ReferenceError: Trans is not defined` error.

Note: I can reproduce this with Lingui 2, not with Lingui 3, but I think this should be fixed as the error message is not explicit at all and finding out what is happening is painful.

**To Reproduce**

```ts
import * as React from "react"

import type { withI18nProps } from "@lingui/react"

export const Comp = () => <Trans>Foo</Trans>
```

This gets transpiled by Babel to:

```ts
"use strict";

var _interopRequireWildcard = require("@babel/runtime/helpers/interopRequireWildcard");

Object.defineProperty(exports, "__esModule", {
  value: true
});
exports.Comp = void 0;

var React = _interopRequireWildcard(require("react"));

var Comp = function Comp() {
  return /*#__PURE__*/React.createElement(Trans, {
    id: "Foo"
  });
};

exports.Comp = Comp;
```

Note that it is using `Trans` and not importing `@lingui/react` at all.

If you comment the `import type` line it works as expected.

If you have a standard `@lingui/react` import *before* the `import type`, it works fine. If you add it after, it fails the same way as above.

I think the macro wrongly detects the `import type` as a normal import, and tries to add the needed export (here `Trans`) to the import list, which is then removed from the final output by Babel as this is TS specific and needs to be trimmed from the final output.

**Additional context**

- jsLingui version ``lingui --version``: 2.9.2
- Babel version ``npm list @babel/core``: 7.12.3
```json
{
  "presets": [
    "@babel/preset-typescript",
    "@babel/preset-react",
    [
      "@babel/preset-env",
      {
        "useBuiltIns": "usage",
        "corejs": 3
      }
    ]
  ],
  "plugins": [
    [
      "babel-plugin-styled-components",
      {
        "ssr": false,
        "transpileTemplateLiterals": true,
        "minify": true,
        "displayName": false
      }
    ],
    "lodash",
    "macros",
    "@babel/plugin-syntax-dynamic-import",
    "@babel/plugin-proposal-object-rest-spread",
    "@babel/plugin-proposal-class-properties",
    [
      "@babel/plugin-transform-runtime",
      {
        "regenerator": false
      }
    ]
  ]
}
```

semoal:
Fixes #777, also tested 👍🏻

### DIFF
--- a/packages/macro/test/jsx-trans.ts
+++ b/packages/macro/test/jsx-trans.ts
@@ -217,6 +217,20 @@ export default [
       `,
   },
   {
+    name: "Production - import type doesn't interference on normal import",
+    production: true,
+    useTypescriptPreset: true,
+    input: `
+        import { withI18nProps } from '@lingui/react'
+        import { Trans } from '@lingui/macro';
+        <Trans id="msg.hello" comment="Hello World">Hello World</Trans>
+      `,
+    expected: `
+        import { withI18nProps, Trans } from "@lingui/react";
+        <Trans id="msg.hello" />;
+      `,
+  },
+  {
     name: "Strip whitespace around arguments",
     input: `
         import { Trans } from "@lingui/macro";
--- a/packages/macro/src/index.ts
+++ b/packages/macro/src/index.ts
@@ -61,8 +61,10 @@ function addImport(babel, state, module, importName) {
   const { types: t } = babel
 
   const linguiImport = state.file.path.node.body.find(
-    (importNode) =>
-      t.isImportDeclaration(importNode) && importNode.source.value === module
+    (importNode) =>t.isImportDeclaration(importNode) &&
+      importNode.source.value === module &&
+      // https://github.com/lingui/js-lingui/issues/777
+      importNode.importKind !== "type"
   )
 
   const tIdentifier = t.identifier(importName)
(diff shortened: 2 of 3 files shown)


packages/macro/test/jsx-trans.ts
```diff
@@ -216,6 +216,20 @@ export default [
         <Trans id="msg.hello" />;
       `,
   },
+  {
+    name: "Production - import type doesn't interference on normal import",
+    production: true,
+    useTypescriptPreset: true,
+    input: `
+        import { withI18nProps } from '@lingui/react'
```
renchap, line 224: You are not using `import type` in your tests

tricoder42: It wasn't related to `import type`. The same problem appeared even with regular import. I've checked it locally that with new changes it works (with regular and type import).

tricoder42 (approved):

tricoder42:
I can confirm this works. Perfect job! 👏

PR #787, opened 2020-10-27
fix: detect-locale package not deployed correctly

semoal:
Sorry Tomas didnt know that npm folder was required to publish, also i changed some babel transform methods that are deprecated and we'll be removed in babel 8.0.0 👍🏻


tricoder42:
No worries! Move fast and break things. It's still pre-release, no big deal

PR #799, opened 2020-10-31
Allow null override for default component

dhaigh:
the docs state that `render={null}` or `component={null}` should override the `defaultComponent` prop. (https://github.com/lingui/js-lingui/blame/v3.0.0/docs/ref/react.rst#L88)

suppose I had this

```ts
const Span = ({ children }) => {
    return <span>{children}</span>;
};

const App = () => {
    return (
        <I18nProvider i18n={i18n} defaultComponent={Span}>
            <Trans id='Hello' render={null} />
        </I18nProvider>
    );
};
```

current behaviour: `<App /> === <span>Hello</span>`
expected behaviour:`<App /> === 'Hello'`

### DIFF
--- a/packages/react/src/Trans.test.tsx
+++ b/packages/react/src/Trans.test.tsx
@@ -44,10 +44,10 @@ describe("Trans component", function () {
     const originalConsole = console.error
     console.error = jest.fn()
 
-    renderWithI18n(<Trans render="span" id="Just a text" />)
+    renderWithI18n(<Trans render="span" id="Some text" />)
     expect(console.error).toHaveBeenCalled()
 
-    renderWithI18n(<Trans render="span" id="Just a text" />)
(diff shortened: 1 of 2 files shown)


tricoder42:
Hey @dhaigh, thanks for the bug fix! You're right. Could you also please provide a test? Just add relevant test cases to `rendering` section in `Trans.test.tsx`:

https://github.com/lingui/js-lingui/blob/de7d07e6004fe379d8a702c056496536a44a285c/packages/react/src/Trans.test.tsx#L139-L204

dhaigh:
@tricoder42 i had a go, let me know what you think :)

tricoder42 (approved):
Pretty good!

tricoder42:
Released in [3.0.1](https://github.com/lingui/js-lingui/releases/tag/v3.0.1). Thank you for your contribution! 👍

PR #807, opened 2020-11-02
feat: accept t as function

### LINKED ISSUE #800: Support id and comment in `t` macro
This is often requested feature — `t` macro should accept custom `id` and/or `comment`. Using `defineMessage` in combination with `i18n._` is unnecessary verbose.

## Proposal

TLDR: Same api as `defineMessage`, but the output will be wrapped in `i18n._` as `t` and other macros does:

```
t({
  id: 'msgId,
  message: 'text'
})
// ↓ ↓ ↓ ↓ ↓ ↓
i18n._(/*i18n*/{ id: "msgId", values: { name }, message: 'text' })


t({
  id: 'msgId,
  message: 'text',
  comment: 'description for translators'
})
// ↓ ↓ ↓ ↓ ↓ ↓
i18n._(/*i18n*/{ id: "msgId", values: { name }, message: 'text', comment: 'description for translators' })


t({
  id: 'msgId,
  message: plural(value, { one: '...', other: '...' })
})

// ↓ ↓ ↓ ↓ ↓ ↓
i18n._(/*i18n*/{ id: "msgId", values: { name }, message: '{value, plural, one {...} other {...}' })

// Important! This should still work:
t`Hello World`
i18n._(/*i18n*/{ id: 'Hello World')
```

This should be fairly easy, because all logic for parsing messge descriptors is already inside `defineMessage` macro. It just needs an extra `if` in the right location, so the `t` macro can be used either as a template literal tag or a function.

## Tasks

- [x] update `t` macro in [`macroJs.ts`](https://github.com/lingui/js-lingui/blob/main/packages/macro/src/macroJs.ts#L213) to also accept message descriptor, the same way how `defineMessage` does
- [x] add tests to [`js-t.ts`](https://github.com/lingui/js-lingui/blob/main/packages/macro/test/js-t.ts) to cover new usecases
- [x] update `t` macro docs

Related issues: #797 #701

semoal:
Closes #800

Couldn't update the docs because i'm shit writing tables with RST.. any suggerence? haha

### DIFF
--- a/docs/ref/macro.rst
+++ b/docs/ref/macro.rst
@@ -138,6 +138,13 @@ Examples of JS macros
 +-------------------------------------------------------------+--------------------------------------------------------------------+
 | .. code-block:: js                                          | .. code-block:: js                                                 |
 |                                                             |                                                                    |
+|    t({                                                      |    i18n._(/*i18n*/{                                                |
+|       id: "msg.refresh",                                    |      id: "msg.refresh",                                            |
+|       message: "Refresh inbox"                              |      message: "Refresh inbox"                                      |
+|    })                                                       |    })                                                              |
++-------------------------------------------------------------+--------------------------------------------------------------------+
+| .. code-block:: js                                          | .. code-block:: js                                                 |
+|                                                             |                                                                    |
 |    describeMessage({                                        |    /*i18n*/{                                                       |
 |       id: "msg.refresh",                                    |      id: "msg.refresh",                                            |
 |       message: "Refresh inbox"                              |      message: "Refresh inbox"                                      |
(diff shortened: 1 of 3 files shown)


tricoder42:
I don't know either. It's pain. But it looks good when it's done. I'll write the docs tonight. Thanks for the PR, I'll review it tonight as well!

tricoder42:
Docs added, PR looks good! Great job 👍

Bertg:
Hi!

So I was trying to use this functionality today (using v3.0.3) but I'm getting a typescript error:

```tsx
          <li>
            {t({id: "long_value"})}
          </li>
```

```
Argument of type '{ id: string; }' is not assignable to parameter of type 'TemplateStringsArray'.
  Object literal may only specify known properties, and 'id' does not exist in type 'TemplateStringsArray'.ts(2345)
```

I'm guessing the typescript definitions did not get updated?

tricoder42:
It just wasn't released yet. We don't have an automatic release process for pre-releases and I wanted to make a new minor release next week on Monday. There's still few new features in development.

Bertg:
@tricoder42 Oh! Ok now i see! That makes sense.

However, the public documentation already mentions this feature. That is why I wanted to try it.
Maybe there is a need to look at the documentation release process?

tricoder42:
Yeah, this is a mistake... We should merge new features into a `next` branch next time, so we can publish pre-releases and documentation preview independently on what's currently in `main`. Good point, thanks 👍

PR #812, opened 2020-11-02
fix: accept catalog paths without ending slash

### LINKED ISSUE #809: Accept catalog paths without ending slash
Directories in catalog paths must end with slash, otherwise they're treated as files. This behavior is confusing for many users.

Detect directories not by checking ending slash, but by checking `fs.lsStats().isDirectory()`

Explain users over and over, that this isn't a bug but a feature.

**Additional context**
I believe we simply need to fix `normalizeRelativePath` function https://github.com/lingui/js-lingui/blob/cbe946ea4d4219630e16fd5333cc08483cd95d92/packages/cli/src/api/catalog.ts#L500

Since the test coverage of `normalizeRelativePath` is very low, please provide all related tests.

Closes #808

semoal:
Closes #809 
Added mark: true to node-glob because internally node-glob needs an slash at the end, will be the same of running normalizeRelativePaths to exclude array

I want to (if you are agree of course) migrate node-glob to fdir that is significantly more faster, will work on it with some benchmarks this week 👍🏻

### DIFF
--- a/packages/cli/src/api/__snapshots__/catalog.test.ts.snap
+++ b/packages/cli/src/api/__snapshots__/catalog.test.ts.snap
@@ -363,7 +363,7 @@ Object {
 }
 `;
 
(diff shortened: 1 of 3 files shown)


tricoder42 (approved):

tricoder42:
Looks good!

Honestly, I don't need to see a benchmark. If you want to refactor it and you know that it's faster, then it's fine.

PR #814, opened 2020-11-02
Add `lingui extract-template` command

### LINKED ISSUE #793: Write .pot file
Translation tools (Transifex, weblate, etc.) usually use a .pot file as an input file to load strings.
This is almost like a normal .po file but the msgstr is always empty.

Add a command to create/update a pot file. The path to it should be customizable

While the english .po file could in theory be used, it would be nicer if there was a way to explicitly write a .pot file. Also, one may put the english .po file on gitignore since having it under version control is yet another source for annoying merge/rebase conflicts.

ThiefMaster:
closes #793 which also contains details about why this is useful

### DIFF
--- a/docs/ref/cli.rst
+++ b/docs/ref/cli.rst
@@ -73,6 +73,17 @@ Convert message catalogs from previous format (see :conf:`format` option).
 
 Prints additional information.
 
+``extract-template``
+--------------------
+
+.. lingui-cli:: extract-template [--verbose]
+
+This command extracts messages from source files and creates a ``.pot`` template file.
+
+.. lingui-cli-option:: --verbose
+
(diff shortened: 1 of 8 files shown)


tricoder42:
Cool! I've realized this command is only useful when you use `po` format, but let's proceed. I don't think someone will need it if they use other formats.

I also added at least basic docs. Please check if it's enough or if you'd add something.

Thank you for your contribution! 👍

tricoder42 (approved):

ThiefMaster:
Thanks! I don't like writing docs that much myself so glad someone else did before I had the opportunity to do so :p

tricoder42:
Perfect! Thank you again! 👍

PR #815, opened 2020-11-02
Fix test environment

ThiefMaster:
The missing `mockFs.restore()` in #812 broke snapshot tests rather badly. Existing tests did not fail, but no new snapshots were written, and every test run created new ones (in some temp location).


tricoder42:
Ah, good catch! Thanks 👍

PR #817, opened 2020-11-03
Fix extracting with --overwrite

### LINKED ISSUE #616: CLI command `extract --overwrite` deletes msgstr values in source locale po file
**Describe the bug**

The first time `lingui extract --overwrite` is run then it generates po file entries for the source locale with identical `msgid` and `msgstr` values, e.g.

```
#: src/file.ts:105
msgid "Hello world"
msgstr "Hello world"
```

But then future runs of `lingui extract --overwrite` will remove the `msgstr` values in existing unedited entries, and changes the above to,

```
#: src/file.ts:105
msgid "Hello world"
```

This renders the source locale po file seemingly invalid, and breaks integrations with some cloud translation services like CrowdIn with errors like `The imported messages.po localization file is invalid. GetText PO validation module said: line ###: missing msgstr section`.

**To Reproduce**

Config looks like

```
module.exports = {
  sourceLocale: 'en',
  srcPathDirs: ['<rootDir>/src'],
  format: 'po',
  sorting: 'origin',
  localeDir: '<rootDir>/src/locale',
}
```

And we're just using the plain JS approach (not React), e.g.

```
import { t } from '@lingui/macro'
i18n._(t`Hello world`)
```

**Expected behavior**

Repeated runs of `lingui extract --overwrite` do not remove `msgstr` entries for existing translations in the source locale po file. Although this is an assumption, and my understanding of the process may be faulty.

**Additional context**

- jsLingui version `2.9.1`
- Babel version `babel-core@7.0.0-bridge.0`
- Your Babel config

```
module.exports = {
  presets: [
    [
      '@babel/preset-env',
      {
        targets: {
          node: '10',
        },
        modules: 'cjs',
      },
    ],
    '@babel/typescript',
  ],
  plugins: [
    '@babel/proposal-class-properties',
    '@babel/proposal-object-rest-spread',
    'macros',
  ],
}
```

ThiefMaster:
When extracting messages coming e.g. from `<Trans>something</Trans>` we only have the key and `origin`, but no `message` in the nextCatalog. Only when using `<Trans id="foo">something</Trans>` we get an actual `message` as well.

I'm not sure if this is the best way of fixing it, but it seems to do the job.
Needs updated unit tests of course, but not sure what's the best way to test  this...

fixes #616

### DIFF
--- a/packages/cli/src/api/catalog.ts
+++ b/packages/cli/src/api/catalog.ts
@@ -191,8 +191,10 @@ export class Catalog {
           (prevCatalog[key].translation === prevCatalog[key].message ||
             options.overwrite)
 
+        // when updating from defaults, use the message in case we have one (custom keys)
+        // otherwise (with keys being the source language string) just use the key itself
         const translation = updateFromDefaults
-          ? nextCatalog[key].message
+          ? nextCatalog[key].message || key
           : prevCatalog[key].translation
 
         return {


tricoder42:
Hmm, so what it does actually?

In the catalog for source locale I would expect that `message` is empty, unless we use a custom ID:

```po
# Extracted from <Trans>Hello World</Trans>
msgid "Hello World"
msgstr ""
```

```po
# Extracted from <Trans id="msg.hello">Hello World</Trans>
msgid "msg.hello"
msgstr "Hello World"
```

The reason is that once you change the `msgstr` in po file (e.g. you change it to a better copy) and then you change it in the source code, it's hard to figure out if the message should be updated because it could overwrite the version from catalog. That's what the `--overwrite` flag is for — it forces overwrite from source to catalog and ignore any changes. Obviously this is only useful when custom ids are used. By default, when `msgstr` is empty, the `id` is used, so the empty `msgstr` in the first catalog would work correctly.

Now, it looks like your change would fill the source catalog as well:

```po
# Extracted from <Trans>Hello World</Trans>
msgid "Hello World"
msgstr "Hello World"
```

Is it really intended?

ThiefMaster:
hm, i assumed msgid==msgstr was intended (and needed to compile the locale) since that's what i got during the very first extraction

tricoder42:
Alright, let me play with it tomorrow. I need to check what's going on

PR #818, opened 2020-11-03
fix: improved performance of formatters

### LINKED ISSUE #472: Performance issues with formatters

Lingui takes a lot of time to create number and date formatters, showing in the profiles consuming 25% to 85% of render time. 

This is due to repeatedly creating Number and Date formatters: https://github.com/lingui/js-lingui/blob/master/packages/core/src/formats.js#L31  and https://github.com/lingui/js-lingui/blob/master/packages/core/src/formats.js#L20 , which is a very expensive operation (at least in some browsers - I we tested this in Chrome)

We'd like to have a API to provide custom number/date formatter so we can use memoized formatters. This is also in the wishlist for Lingui 3, as I can see. 

As a temporary solution until we get a proper one we did this: https://github.com/SectorLabs/js-lingui/pull/6 in our fork of lingui

semoal:
Will close #472

I've added to format functions a third param to disable/enable the memoize function for testing purposes in that way we can be 100% sure that the cache is working properly.

Some results (1000x renders of `date`)
```
    ┌──────────────────────┬────────────────────┐
    │       (index)        │       Values       │
    ├──────────────────────┼────────────────────┤
    │  memoizedDateResult  │  6.58090500000003  │
    │ withoutMemoizeResult │ 263.14999799999987 │
    └──────────────────────┴────────────────────┘
```

Some results (1000x renders of `number`)
```
    ┌──────────────────────┬────────────────────┐
    │       (index)        │       Values       │
    ├──────────────────────┼────────────────────┤
    │ memoizedNumberResult │ 2.9453200000002653 │
    │ withoutMemoizeResult │ 45.79147499999999  │
    └──────────────────────┴────────────────────┘
```

### DIFF
--- a/packages/core/src/context.ts
+++ b/packages/core/src/context.ts
@@ -16,8 +16,7 @@ const defaultFormats = (
     return (ctx) => {
       const msg = isFunction(message) ? message(ctx) : message
       const norm = Array.isArray(msg) ? msg : [msg]
-      const formatter = new Intl.NumberFormat(locales)
-      const valueStr = formatter.format(value)
+      const valueStr = number(locales)(value)
       return norm.map((m) => (isString(m) ? m.replace("#", valueStr) : m))
     }
   }
--- a/packages/core/src/formats.test.ts
+++ b/packages/core/src/formats.test.ts
@@ -1,0 +1,49 @@
+import { date, number } from "./formats"
+
+describe("@lingui/core/formats", () => {
+  it("number formatter is memoized", async () => {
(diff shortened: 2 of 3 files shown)


tricoder42 (approved):

tricoder42:
Awesome! Pretty good performance boost 👍

tricoder42 (approved):

tricoder42 (approved):

tricoder42:
Nice! Looks good to me 🚀

PR #820, opened 2020-11-03
feat: fallback cldr

### LINKED ISSUE #791: Add fallback languages for the i18n instance.
First of all, I would like to thank all of you, your work on this library. Actually, v3 was launched before due date, so congrats!! :tada: 

Linguijs does not allow you to decide fallback locales for specific languages.

Instead, lingui will show source language strings.

Add fallback specific languages for the lingui instance. It could be similar to [i18next API](https://www.i18next.com/principles/fallback#fallback-language).
~~~
{
    es-MX: ['es'],
    es-AR: ['es'],
    default: ['en'],
}
~~~

For the moment, this handling relies on yourself, loading the catalog inside `if statements` in case of unknown/unsupported language. This is kinda tedious, because it's not that easy as "just load english for every new locale". For example, we would like to handle all hispanic locales ("es-MX, es-AR") to be have a fallback in "es", but, every other to use the default fallback locale (en).

semoal:
Will close #791	

- [x] Update docs
- [x] Add some extra testing
- [x] Clean some duplicated types

### DIFF
--- a/packages/cli/src/api/catalog.ts
+++ b/packages/cli/src/api/catalog.ts
@@ -238,6 +238,10 @@ export class Catalog {
     const getMultipleFallbacks = (locale) => {
       let i = 0
       const fL = fallbackLocales[locale]
+
+      // some probably the fallback will be undefined, so just search by locale
+      if (!fL) return null
+
       if (Array.isArray(fL)) {
         while(i <= fL.length) {
           const fallbackLocale = fL[i]
--- a/packages/cli/src/api/catalog.ts
+++ b/packages/cli/src/api/catalog.ts
@@ -44,9 +44,14 @@ export type MergeOptions = {
   overwrite: boolean
 }
 
+// @todo: import from @lingui/conf
+export type FallbackLocales = {
+  [locale: string]: string[] | string
+} | { default: string[] | string }
+
 export type GetTranslationsOptions = {
   sourceLocale: string
-  fallbackLocale: string
+  fallbackLocales: FallbackLocales
 }
 
 type CatalogProps = {
@@ -94,7 +99,7 @@ export class Catalog {
       )
     ) as unknown) as (catalog: AllCatalogsType) => AllCatalogsType
 
-    const sortedCatalogs = cleanAndSort(catalogs);
+    const sortedCatalogs = cleanAndSort(catalogs)
 
     if (options.locale) {
       this.write(options.locale, sortedCatalogs[options.locale])
@@ -227,7 +232,7 @@ export class Catalog {
     catalogs: Object,
     locale: string,
     key: string,
-    { fallbackLocale, sourceLocale }: GetTranslationsOptions
+    { fallbackLocales, sourceLocale }: GetTranslationsOptions
   ) {
     if (!catalogs[locale].hasOwnProperty(key)) {
       console.error(`Message with key ${key} is missing in locale ${locale}`)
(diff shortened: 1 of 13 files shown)


semoal:
Practically the feature is done, 
- if no fallbackLocales is passed we look the cldr parent locale of each locale in config file, and we create our own fallbackLocales through locales array
- if `fallbackLocales.default` is passed we just use that :)
- if `fallbackLocale` (deprecated), we create an object of `fallbackLocales.default` with that locale

I think all the cases are covered, I'm a bit tired now so probably something is not correct, I'll take a better look tomorrow

semoal:
I'll do some extra testing on a sample project, to check 100% if all the cases are covered and we're ready to :shipit:

semoal:
Tested on create-react-app of examples folder adding ES-MX and ES, translating some things in "es", appears in es-mx 👍🏻 

So I think, it works as intended

packages/conf/src/index.ts
```diff
@@ -241,22 +276,218 @@ export function replaceRootDir(
 }
 
 /**
- * Replace fallbackLanguage with fallbackLocale
- *
- * Released in lingui-conf 0.9
- * Remove anytime after 3.x
+ * Replace fallbackLanguage and fallbackLocale, by the new standard fallbackLocales
+ * - https://github.com/lingui/js-lingui/issues/791
+ * - Remove anytime after 3.x
```
tricoder42, line 281:
```suggestion
 * - Remove anytime after 4.x
```

We should bump the version here. The idea is that these deprecations shouldn't break current version, but they should be included in 4.x release notes. After that we can remove them.

packages/conf/src/index.ts
```diff
@@ -241,22 +276,218 @@ export function replaceRootDir(
...
+  })
+
+  const DEFAULT_FALLBACK = fallbackLocales?.default || fallbackLocale || fallbackLanguage
+  if (DEFAULT_FALLBACK) {
+    if (!config.fallbackLocales) config.fallbackLocales = {}
+    config.fallbackLocales.default = DEFAULT_FALLBACK
   }
+
+  return config
+}
```
tricoder42, line 312: I would remove `fallbackLanguage` completely. This change was introduced in Lingui 0.9 and there were two major bumbs since then. I don't think we need to keep it that long, what do you think?

If you agree, let's simplify it to converting `fallbackLocale` to `fallbackLocales` and just add one check if both `fallbackLocale` and `fallbackLocales.default` are set. That's the only case which should end up with hard error.

packages/cli/src/api/catalog.ts
```diff
@@ -227,26 +227,48 @@ export class Catalog {
...
+
     return (
       // Get translation in target locale
       getTranslation(locale) ||
-      // Get translation in fallbackLocale (if any)
-      (fallbackLocale && getTranslation(fallbackLocale)) ||
+      // Get translation in fallbackLocales.default (if any)
+      (fallbackLocales.default && getTranslation(fallbackLocales.default)) ||
+      // We search in fallbackLocales as dependent of each locale
+      getMultipleFallbacks(locale) ||
```
tricoder42, line 264: shouldn't the `getMultipleFallbacks` has a higher precedence before `fallbackLocales.default`?

For example:

```json
{
  "fallbackLocales": {
    "en-gb": "en",
    "default": "cs"
  }
}
```

Then for a locale `en-gb` I should first try "en" and only if it doesn't exist I should fallback to "cs". What do you think?

packages/cli/src/api/catalog.ts
```diff
@@ -227,26 +227,48 @@ export class Catalog {
...
+      if (!fL) return null
+
+      if (Array.isArray(fL)) {
+        while(i <= fL.length) {
+          const fallbackLocale = fL[i]
+          if (catalogs[fallbackLocale]) {
+            return getTranslation(fallbackLocale)
+          }
+          i++
+        }
```
tricoder42, line 252:
```suggestion
        for(const fallbackLocale of fL) {
          if (catalogs[fallbackLocale]) {
            return getTranslation(fallbackLocale)
          }
        }
```

Could we remove the `while` loop here?

packages/core/src/formats.test.ts
```diff
@@ -27,40 +27,4 @@ describe("@lingui/core/formats", () => {
...
-
-    const loop0 = performance.now()
-    for (let i = 0; i < 1000; i++) {
-      number("es", {}, false)(999666)
-    }
-    const loop1 = performance.now()
-    const withoutMemoizeResult = loop1 - loop0
-
-    expect(memoizedNumberResult).toBeLessThan(withoutMemoizeResult)
-  })
```
tricoder42, line 65: Why is it removed in this PR? You've accidentaly started branch from unfinished previous PR? Just checking :)

tricoder42 (commented):
Looking good, great work! Just a few comments bellow, but this is going to be a huge improvement 👍

packages/core/src/formats.test.ts
semoal: That two tests that were run inside a big loop weren't exact when running in jest parallelism. With all the other tests we can see if we have any regression in performance so it's enough

packages/cli/src/api/catalog.ts
semoal: Absolutely

tricoder42 (approved):

tricoder42:
Looking good! Thanks a lot 👍

MartinCerny-awin:
I am trying to use fallbackLocales and getting error
```
● Validation Warning:

  Unknown option "fallbackLocales" with value {"default": "en", "ps": ["en"]} was found. Did you mean "fallbackLocale"?
```

What is the correct way now to use it?

MartinCerny-awin:
Ok, I guess I am reading the latest docs for unreleased version

tricoder42:
@MartinCerny-awin Yes, sorry, there's currently a problem with latest release of docs... It'll be fixed early next week with new release setup

PR #821, opened 2020-11-05
fix: t macro as function typed

semoal:
Totally forgot the type 🤕 
Will fix @Bertg comment of #807
Screenshot 2020-11-05 at 12 26 06


tricoder42:
Ah, cool! Good catch 👍

PR #823, opened 2020-11-05
feat: lingui commands suggestion looks package.json

### LINKED ISSUE #727: `lingui extract` command prints possibly misleading messages
Not really a problem, more like a UX issue. 

After `lingui extract` command finishes there is a message which is likely to be irrelevant due to user's different script configuration. The message is:

```sh
(use "npm run extract" to update catalogs with new messages)
(use "npm run compile" to compile catalogs for production)
```

Print a message without assuming that there are user-defined npm scripts at all:
```sh
(run "lingui extract" command to update catalogs with new messages)
(run "lingui compile" command to compile catalogs for production)

# OR

(use "npx lingui extract" to update catalogs with new messages)
(use "npx lingui compile" to compile catalogs for production)
```
And, if Yarn is used then it should be `use "yarn lingui ...` appropriately. This is being detected already.

Not printing this block at all

**Additional context**
I have a project with junior developers and such messages can be misleading to them. Now I have to document this so they know that in our setup these two operations use different script names.

Here is my scripts section
```json
"scripts": {
  "locales:extract": "lingui extract",
  "locales:build": "lingui compile",
  "locales:dev": "onchange \"src/**/*.{ts,tsx}\" -- npm run locales:extract"
}
```

**Update:** proposed a solution with `npx`, added my scripts configuration which leads to irrelevant suggestions message

semoal:
Will close #727

Super easy feature, but will help a lot the introduction to not soo much experienced users on lingui

### DIFF
--- a/packages/cli/src/api/help.ts
+++ b/packages/cli/src/api/help.ts
@@ -1,5 +1,6 @@
 /**
- * Detect how cli command is being run (npm, yarn) and construct help
+ * Detect where's is the command lingui extract or lingui compile
+ * and how is being run (npm, yarn) and construct help
  * for follow-up commands based on that.
  *
  * Example:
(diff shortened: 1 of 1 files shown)


tricoder42 (approved):

packages/cli/src/api/help.ts
```diff
@@ -15,7 +16,26 @@
...
+      path.join(process.cwd(), "package.json")
+    ))
+  } catch (error) {}
+
+  if (findRootPkgJson?.scripts) {
+    const res = Object
+      .entries(findRootPkgJson.scripts)
+      .find(([_, value]) => value.includes(`lingui ${command}`))
+
+    if (res[0]) {
```
tricoder42, line 34: Isn't `res == null` when it doesn't find a matching command?

```suggestion
    if (res != null) {
```

semoal: Ahh, right damn it, i'll change it asap

semoal: I don't why I tought that entries was returning an empty array if not found, changes applied! :shipit:

tricoder42 (approved):

tricoder42:
Looking good! 👍

PR #824, opened 2020-11-05
fix: dx on plurals not called

### LINKED ISSUE #798: Improve error message when plurals aren't loaded
## Tasks

- [x] show `console.error` "Plurals for locale {locale} aren't loaded. Use i18n.loadLocaleData method to load plurals for specific locale. Using `other` plural rule as a fallback."
- [x] Use `other` plural rule as a fallback

I know this isn't 100% accurate, because some locales, e.g. pl, ru, actually use `many` as a fallback rule, while `other` is used fo decimal numbers. However, this shouldn't be used in production anyway. It's just for easier debugging.

---

Original post:

**Describe the bug**
Using plurals, when an unsupported language gets activated (via i18n.activate) which was not loaded via `loadLocaleData`, app breaks.
image

Without using plurals, just a console warn pops out:
image

An example of how this could happen, is when you activate users browser language, which potentially is not under the loaded plurals. 

**To Reproduce**
[CodeSandBox](https://codesandbox.io/s/dreamy-haze-5p5y9?file=/src/App.tsx)

**Expected behavior**
We expected the same behaviour as working without plurals. Plurals will be computed with fallback language or loaded plurals.

**Additional context**
Even when the issue is not the same (I already have some plurals loaded) this could be a related one: https://github.com/lingui/js-lingui/issues/437

- jsLingui version ``3.0.0``
- Create React App (3.3.0)

semoal:
Will close #798

The app doesn't crash if loadLocaleData is not called, but logs a console.error warning about.

### DIFF
--- a/packages/core/src/context.ts
+++ b/packages/core/src/context.ts
@@ -21,14 +21,18 @@ const defaultFormats = (
     }
   }
 
+  if (!plurals) {
+    console.error(`Plurals for locale ${locale} aren't loaded. Use i18n.loadLocaleData method to load plurals for specific locale. Using other plural rule as a fallback.`)
+  }
+
   return {
     plural: (value, { offset = 0, ...rules }) => {
-      const message = rules[value] || rules[plurals(value - offset)]
+      const message = rules[value] || rules[plurals?.(value - offset)] || rules.other
       return replaceOctothorpe(value - offset, message)
     },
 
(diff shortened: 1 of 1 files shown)


tricoder42:
Pretty good! 👍  Lingui 3.1 is gonna be a blast 🚀

PR #825, opened 2020-11-05
fix: exit 1 on strict mode of compile

### LINKED ISSUE #722: Add a CLI command to validate the catalog on CI

semoal:
Will close #722


tricoder42 (approved):
Yep 👍

PR #827, opened 2020-11-05
ci: release strategy refactored

semoal:
### Done
- [X] Added a `main-suite` workflow:
	- That is responsible of running all the linting, testing, building, and codecoverage on every pull request
	- Also is responsible when running the action manually of deploying the packages to "next" | "latest", (this is defined where the action is run)
	- Also has a matrix-strategy where all the suite is run for node 10,12 and 14.

- [X] Added `next-sync` workflow:
	- Is responsible of syncing `next` branch with `main`, when we push something to `main` automatically will create a pr with all the changes pointing to `next` 👍🏻 

- [X] Added `deploy-main-docs` workflow:
	- Is responsible of deploying automatically the documentation of `main` branch only when someone pushes any change in `/docs/**/*` folder, super cool.
	- `It can be easily changed to deploy to differnt subdomain of vercel if the branch is next`, I'm not an expert of vercel cli or configuration so that part is pretty dark to me, but I can help configuring the dynamism of the Github Action

- [x] Add `pr-lint` pr-lint action will check that the title of the pull request follow our standards, in that way, we only we'll need to merge that pr, and we'll be done. Example of use https://github.com/puppeteer/puppeteer/pull/6540
- [X] Introduced logic of [skip ci] for every workflow 
- [x] Decide if we want to do the full main-suite to every node version (for example we can skip building?, or integration?, dunno) `Right now, I put build, integration and release only for node 14 (lts)`
- [X] Create a `lingui-bot` account to do the commit deployments where's the changelog, package.json changes.. https://github.com/lingui-bot

### Todos

- [ ] Some pending environment variables, i have to check.
We have to define this secrets:
```yml
GH_BOT_PWD: "",
NPM_TOKEN: "",
NOW_TOKEN: ""
```
@tricoder42 i send you the gh_bot_pwd via email 💯 

- [ ] A careful read to check everything is fine.


.github/workflows/main-suite.yml
```diff
@@ -0,0 +1,66 @@
...
+
+      - name: Install dependencies if needed
+        if: steps.yarn-cache.outputs.cache-hit != 'true'
+        run: yarn install --frozen-lockfile --pure-lockfile
+
+      - name: Unit Testing
+        run: yarn test
+
+      - name: Linting & Types
+        run: yarn lint:all
```
tricoder42, line 54: Is worth it to `lint` files on all Node.js versions?

.github/workflows/next-sync.yml
```diff
@@ -0,0 +1,28 @@
...
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  next-sync:
+    runs-on: ubuntu-latest
+    if: "! contains(toJSON(github.event.commits.*.message), '[skip ci]')"
```
tricoder42, line 11: Shouldn't this actions run anytime there's a new commit in `main`? No matter if it's marked as `[skip ci]` or not.

.github/workflows/deploy-main-docs.yml
```diff
@@ -0,0 +1,60 @@
...
+        with:
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Install dependencies if needed
+        if: steps.yarn-cache.outputs.cache-hit != 'true'
+        run: yarn install --frozen-lockfile --pure-lockfile
+
```
tricoder42, line 43: Is it necessary to install Node when we're building documentation with Python? Even the `npx vercel` command is going to fetch it's own dependencies, isn't it?

```suggestion
```

.github/workflows/deploy-main-docs.yml
```diff
@@ -0,0 +1,60 @@
+name: deploy-main-docs
```
tricoder42, line 1: Do I undestand it correctly that the whole point of building the documentation in GitHub actions rather than in Vercel is that we can control *when* the documentation is build? I.e. only when files inside `docs` have changed and there isn't `[skip ci]` in commit message?

.github/workflows/release.yml
```diff
@@ -0,0 +1,72 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Setup node
+        uses: actions/setup-node@v2-beta
+        with:
+          always-auth: true
+          node-version: 12
```
tricoder42, line 27: What about using the latest LTS here? Can't we specify the version as `latest` or sth?

```suggestion
          node-version: 14
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,72 @@
...
+        run: yarn release:build
+
+      - name: Integration Testing
+        run: yarn test:integration
+
+      - name: Creates local .npmrc
+        run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > ~/.npmrc
+
+      - name: Prepare Lingui-Bot git account
+        uses: oleksiyrudenko/gha-git-credentials@v1
```
tricoder42, line 62: What does this action do? Can't we use a `GH_TOKEN` for authentication?

tricoder42 (commented):
This is great! I'm really looking forward seeing this machinery up and runnin.

I've added @lingui-bot to @lingui organization and it's `GH_TOKEN` to secrets. `NPM_TOKEN` is still from my personal account, but I guess it's fine.

Just left few comments to clarify some things, but overall — amazing job! Thank you for tackling this 👊

tricoder42:
> Decide if we want to do the full main-suite to every node version (for example we can skip building?, or integration?, dunno) Right now, I put build, integration and release only for node 14 (lts)

I would run tests on all versions and relese only on the latest

> It can be easily changed to deploy to differnt subdomain of vercel if the branch is next, I'm not an expert of vercel cli or configuration so that part is pretty dark to me, but I can help configuring the dynamism of the Github Action

It seems we should add [`--prod`](https://vercel.com/docs/cli#commands/overview/unique-options/prod) when deploying to prod domain

.github/workflows/main-suite.yml
semoal: Absolutely not, I'll lint only for LTS

.github/workflows/next-sync.yml
semoal: Hmm, I'm not sure because the [skip ci] commits will include package.json changes, main changelog, something that we must discard when merging to next. So it's not relevant the main changes into next.

.github/workflows/deploy-main-docs.yml
semoal: Totally, has sense

.github/workflows/deploy-main-docs.yml
semoal: Yes, i think is more interesting to control in the CI when the docs are deployed. Now Vercel is deploying the docs always also when the docs doesn't change, so probabbly this strategy it's better.

semoal: Vercel should continue creating the temporal domains for pull requests, but for main/next has more sense to me, to control it on the ci. But as you want, both options are valid.

.github/workflows/release.yml
semoal: Actually not, their docs says:

```
It will first check the local cache for a semver match. The hosted images have been updated with the latest of each LTS from v8, v10, v12, and v14.
```

.github/workflows/release.yml
semoal: This is basically doing

git config --add user.email=linguibot@gmail.com
git config --add user.pwd=github password account

We can't use lingui bot gh token, because will override the internal github token of the ci.

semoal: By default, if we use the github token, the commit will be like this:


INPUT_EMAIL=${INPUT_EMAIL:-'github-action@users.noreply.github.com'}
INPUT_NAME=${INPUT_NAME:-'GitHub Action'}

.github/workflows/next-sync.yml
tricoder42: Hmm, I thought that `next` should be always the same or ahead of `main`. What happens e.g after we release the new version, the `main` will have version `4.0.0`, but `next` will still have version `3.1.4`. Wouldn't it cause troubles down the line?

Ah, I think I got it. Since we have a branching model similar to [git flow](https://nvie.com/posts/a-successful-git-branching-model/), I recall that by default the main branch isn't merged back to develop (similar to what  this action would be). However, it works only because hotfixes (which are direclty merged to `main`) are merged back to `develop`.

I usually use the `git-flow-avh` extension, which automatically merges `main` to `develop` anytime there's a new change. That way I can run `git describe` and see how far ahead of `main` I'm.

semoal: Yes, I usually worked with git flow strategy that's why I introduced that.

.github/workflows/release.yml
tricoder42: Alright, I've just added `GH_BOT_PWD` to secrets. I just feel very unsecure by adding this password anywhere :/

package.json
```diff
@@ -14,11 +15,15 @@
...
     "lint:types": "tsc",
     "lint:eslint": "eslint packages",
-    "lint": "yarn lint:eslint; yarn lint:types",
+    "lint:all": "yarn lint:eslint; yarn lint:types",
     "release": "node ./scripts/release.js",
     "release:integration": "node ./scripts/integration.js",
     "release:build": "node ./scripts/build/",
     "release:test": "node ./scripts/test.js",
+    "version:next": "lerna version --no-private --preid next --create-release github --conventional-commits --conventional-prerelease --yes",
+    "version:latest": "lerna version --no-private --create-release github --conventional-commits --yes",
```
tricoder42, line 24: Does Lerna uses semantic-release internally or does it handle similar process on it's own (e.g. generating Changelogs, etc)?

tricoder42 (approved):
Alright, let's push it and see how it works 👍

package.json
semoal: Yes does it internally, much better because works out of the box with monorepos.

.github/workflows/release.yml
semoal: If you want we can do it manually, doing the config.set without any external action. It's pretty simple

tricoder42: That's fine. Let's ship it :) Can't wait to see it in action! 👍

.github/workflows/next-sync.yml
```diff
@@ -19,6 +19,14 @@ jobs:
         with:
           node-version: 14
 
+      - name: Prepare Lingui-Bot git account
+        uses: oleksiyrudenko/gha-git-credentials@v1
+        with:
+          name: 'Lingui Bot'
+          email: 'linguibot@gmail.com'
+          actor: 'lingui-bot'
+          token: '${{ secrets.GH_BOT_PWD }}'
```
tricoder42, line 28: Hey @semoal, sorry to bother you, but could you please explain to me again what's the benefit of using password over a token?

I've checked the `gha-git-credentials` repository and they use it in their example:

```yml
    # publish to a branch in current repo using GITHUB_TOKEN and other default settings
    - uses: oleksiyrudenko/gha-git-credentials@v1
      with:
        token: '${{ secrets.GITHUB_TOKEN }}'
```

semoal: oh, didn't see that, i only saw the first case that was the case of remote repo. Sorry mate was my fault

semoal: I change it now

semoal: I have one question about that because if we use the GITHUB_TOKEN, will do the login correctly to the linguibot account?

semoal: I changed it to github token, looks if we are in the same repo/org works perfectly with GITHUB_TOKEN and we'll override the committer.

semoal: Nevermind my last sentence, that's wrong I've tried on my lingui fork and the pull request is created by the github action bot account, it's not using the lingui git account.

semoal: we can create a personal access token on lingui bot account, to use that instead of the password I guess. what do you think?

tricoder42: I've already created a personal access token, but I've add it as `GH_TOKEN` and not `GITHUB_TOKEN`. I've just checked that I can't add `GITHUB_TOKEN`, I guess it's reserved for GitHub Actions account.

semoal: Perfect so, I'll change the name asap

semoal: Done Tom

tricoder42 (approved):
Looks good, let's try it 👍

tricoder42:
Just to confirm: What happens after I merge it? Do I need to setup anything manually? I should probably just disable vercel and circleci automatic builds, right?

semoal:
> Just to confirm: What happens after I merge it? Do I need to setup anything manually? I should probably just disable vercel and circleci automatic builds, right?

How main is the default branch automstically will find the new workflows and Will be done. Yes we can disable that

tricoder42:
Cool, let's 🚢  it! Good work!

tricoder42:
The `next-sync` action failed. Not sure why. I've created the `next` branch, but it's still failing. I'll take a look later today.

PR #829, opened 2020-11-06
Add typescript to next example dependencies

MartinCerny-awin:
I have added typescript to dependencies as I think it is better developer experience than assume that they have correct version installed globally. 

This error is shown when running extract without typescript installed
```
Extracting messages from source files…
✖ Cannot find module 'typescript'
Require stack:
- /Volumes/sites/lingui/examples/next-js/node_modules/@lingui/cli/api/extractors/typescript.js
- /Volumes/sites/lingui/examples/next-js/node_modules/@lingui/cli/api/extractors/index.js
- /Volumes/sites/lingui/examples/next-js/node_modules/@lingui/cli/api/catalog.js
- /Volumes/sites/lingui/examples/next-js/node_modules/@lingui/cli/lingui-extract.js
✖ Cannot find module 'typescript'
Require stack:
```

### DIFF
--- a/examples/next-js/lingui.config.js
+++ b/examples/next-js/lingui.config.js
@@ -4,7 +4,7 @@ module.exports = {
   catalogs: [
     {
       path: "<rootDir>/locale/{locale}/messages",
-      include: ["<rootDir>"],
+      include: ["<rootDir>/"],
       exclude: ["**/node_modules/**"],
     },
   ],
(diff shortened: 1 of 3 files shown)


tricoder42:
Thanks! Make sense. I usually miss dependencies in examples, because they're leaked from top-level node_modules...

PR #831, opened 2020-11-06
fix: mandatory ext on @lingui/loader

### LINKED ISSUE #830: @lingui/loader! Not working as advertised (typescript?)
**Describe the bug**

Using dynamicActivate (async import) of pre compiled files works. 
Prefixing ```@lingui/loader!``` to the import path fails hard.

**To Reproduce**

I have set up lingui (v3.0.3) with Typescript. 

.tsconfig
```json
{
  "compilerOptions": {
    "target": "es5",
    "lib": ["dom", "dom.iterable", "esnext"],
    "allowJs": true,
    "skipLibCheck": true,
    "esModuleInterop": true,
    "allowSyntheticDefaultImports": true,
    "strict": true,
    "forceConsistentCasingInFileNames": true,
    "module": "esnext",
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "isolatedModules": true,
    "noEmit": false,
    "jsx": "react",
    "baseUrl": "src"
  },
  "include": ["src"]
}
```

.linguirc
```json
{
   "locales": ["en", "nl"],
   "sourceLocale": "en",
   "catalogs": [{
      "path": "<rootDir>/locales/{locale}/messages",
      "include": ["<rootDir>/**"],
      "exclude": ["**/node_modules/**"]
   }],
   "pseudoLocale": "en-fo",
   "rootDir": ".",
   "format": "po"
}
```

.bablerc
```json
{
  "presets": [
    "@babel/preset-env",
    "@babel/preset-react",
    "@babel/preset-typescript"
  ],
  "plugins": [
    ["@babel/plugin-transform-runtime",
    {
      "regenerator": true
    }],
    "macros"
  ]
}
```

I have a function ```buildDynamicActivate```

```ts
const buildDynamicActivate = (i18n: any) => async (locale: string) => {
  const { messages } = await import(`../../locales/${locale}/messages`);
  i18n.load(locale, messages);
  i18n.activate(locale);
};
```

This works fine, but when I introduce ```@lingui/loader!```

```ts
const buildDynamicActivate = (i18n: any) => async (locale: string) => {
  const { messages } = await import(`@lingui/loader!../../locales/${locale}/messages`);
  i18n.load(locale, messages);
  i18n.activate(locale);
};
```

the runtime throws this error.

```
Unhandled Promise Rejection: Error: Module build failed (from ./node_modules/@lingui/loader/index.js):
TypeError: Cannot read property 'locale' of null
    at Object._default (/app/node_modules/@lingui/loader/index.js:70:35)
```

**Expected behavior**

Page not to crash, and compilation to happen on the fly as advertised.

**Additional context**

Not that the structure of this app is

```
ROOT/
  locales/
    en/...
  src/
    utils/
      dynamic_activate.ts
```

This is possibly related to #642

semoal:
will close #830

### DIFF
--- a/docs/ref/loader.rst
+++ b/docs/ref/loader.rst
@@ -25,12 +25,10 @@ Usage
 Simply prepend ``@lingui/loader:`` in front of path to message catalog you want to
 import. Here's an example of dynamic import:
 
-Extension is mandatory, if you use minimal format, you should use ``.json``, in case of using po format, use ``.po``
-
 .. code-block:: jsx
 
    export async function dynamicActivate(locale: string) {
-      const { messages } = await import(`@lingui/loader!./locales/${locale}/messages.json`)
+      const { messages } = await import(`@lingui/loader!./locales/${locale}/messages.js`)
       i18n.load(locale, messages)
       i18n.activate(locale)
    }
--- a/docs/ref/loader.rst
+++ b/docs/ref/loader.rst
@@ -25,10 +25,12 @@ Usage
 Simply prepend ``@lingui/loader:`` in front of path to message catalog you want to
 import. Here's an example of dynamic import:
 
+Extension is mandatory, if you use minimal format, you should use ``.json``, in case of using po format, use ``.po``
+
 .. code-block:: jsx
 
    export async function dynamicActivate(locale: string) {
-      const { messages } = await import(`@lingui/loader!./locales/${locale}/messages.js`)
+      const { messages } = await import(`@lingui/loader!./locales/${locale}/messages.json`)
       i18n.load(locale, messages)
       i18n.activate(locale)
    }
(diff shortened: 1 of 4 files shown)


docs/ref/loader.rst
```diff
@@ -25,10 +25,12 @@ Usage
 Simply prepend ``@lingui/loader:`` in front of path to message catalog you want to
 import. Here's an example of dynamic import:
 
+Extension is mandatory, if you use minimal format, you should use ``.json``, in case of using po format, use ``.po``
```
tricoder42, line 28:
```suggestion
Extension is mandatory. If you use minimal or lingui file format, use ``.json``. In case of using po format, use ``.po``.
```

tricoder42 (commented):
Just a small suggestion. Native english speaker would probably write it differently, but I found it's easier to split long sentences into smaller ones.

tricoder42:
Good catch! Thanks

PR #832, opened 2020-11-07
docs: improve @lingui/cli installation description in README.md

MartinCerny-awin:


packages/cli/README.md
```diff
@@ -37,9 +42,29 @@ npm run extract
 npm run compile
 ```
 
+### 2) Use NPX
+
+You can run the scripts directly using a tool for executing Node packages `NPX`. `NPX` is included in `NPM` version 5.2 and higher.
+
+```
+npx lingui add-locale -- en cs
```
semoal, line 50: Add locale doesnt exist anymore

MartinCerny-awin: I am also looking into https://github.com/lingui/js-lingui/blob/main/packages/cli/src/api/types.ts and it looks to me that most types are not used. Should I remove them?

MartinCerny-awin: For example what is a purpose of this `CatalogApi` export https://github.com/lingui/js-lingui/blob/main/packages/cli/src/api/types.ts#L58 Isn't it better to define it directly in catalog.ts file?

packages/cli/README.md
```diff
@@ -4,20 +4,25 @@
...
+> lingui command line library for manipulating message catalogues
 
 `@lingui/cli` is part of [LinguiJS][linguijs]. See the [documentation][documentation] for all information, tutorials and examples.
 
 ## Installation
 
-```sh
-npm install --global @lingui/cli
-# yarn global add @lingui/cli
+The library can be installed globally or locally using `yarn` or `nmp`. The recommended way is installing the package locally. This ensures that everyone who uses the project has the same version and does not need to install additional packages.
```
MartinCerny-awin, line 13: npm

MartinCerny-awin:
I have fixed the doc based on comments. Let me know if there is anything else

packages/cli/README.md
```diff
@@ -4,25 +4,29 @@
...
+> lingui command line library for manipulating message catalogues
 
 `@lingui/cli` is part of [LinguiJS][linguijs]. See the [documentation][documentation] for all information, tutorials and examples.
 
 ## Installation
 
-```sh
-npm install --global @lingui/cli
-# yarn global add @lingui/cli
+The library can be installed globally or locally using `yarn` or `np`. The recommended way is installing the package locally. This ensures that everyone who uses the project has the same version and does not need to install additional packages.
```
tricoder42, line 13: Just small typo

```suggestion
The library can be installed globally or locally using `yarn` or `npm`. The recommended way is installing the package locally. This ensures that everyone who uses the project has the same version and does not need to install additional packages.
```

tricoder42:
Looks good! Thank you 👍

PR #835, opened 2020-11-07
Remove types file

MartinCerny-awin:
I have removed unused types and also migrated type definitions to the files that I think should be responsible for them


tricoder42 (approved):
Cool! I trust the green build, so I'm gonna assume it's all good :) Thanks for simplifying this 👍

PR #836, opened 2020-11-07
Enable pseudolocalization from pseudolocale folder

### LINKED ISSUE #744: [PseudoLocalization] Pseudolocalize on Extract
Text is pseudolocalized on compile rather than extract. This makes it hard to modify/fix any issues, for example #743 

`lingui extract` should run `pseudoLocalize`

MartinCerny-awin:
This PR utilises standard getTranslation eg. we first try to get translation from locale folder, then from fallback, and then keys, etc...

This will allow us to set text that is going to be pseudolocalized and it will close #744 

I do not think there is a reason to change to pseudolocalizing on extract, but it will solve the issue @nathanforce is having.

### DIFF
--- a/docs/ref/conf.rst
+++ b/docs/ref/conf.rst
@@ -372,7 +372,8 @@ pseudoLocale
 Default: ``""``
 
 Locale used for pseudolocalization. For example when you set ``pseudoLocale: "en"``
-then all messages in ``en`` catalog will be pseudo localized.
+then all messages in ``en`` catalog will be pseudo localized. The locale has to be included
+in `locales <https://lingui.js.org/ref/conf.html#locales>`_.
 
 rootDir
 -------
--- a/docs/ref/conf.rst
+++ b/docs/ref/conf.rst
@@ -373,7 +373,7 @@ Default: ``""``
 
 Locale used for pseudolocalization. For example when you set ``pseudoLocale: "en"``
 then all messages in ``en`` catalog will be pseudo localized. The locale has to be included
-in `locales <https://lingui.js.org/ref/conf.html#locales>`_.
+in `locales <https://lingui.js.org/ref/conf.html#locales>`_ config.
 
 rootDir
 -------
--- a/docs/ref/conf.rst
+++ b/docs/ref/conf.rst
@@ -373,7 +373,7 @@ Default: ``""``
 
 Locale used for pseudolocalization. For example when you set ``pseudoLocale: "en"``
 then all messages in ``en`` catalog will be pseudo localized. The locale has to be included
-in `locales <https://lingui.js.org/ref/conf.html#locales>`_ config.
+in :conf:`locales` config.
 
 rootDir
 -------
--- a/docs/tutorials/cli.rst
+++ b/docs/tutorials/cli.rst
@@ -207,8 +207,7 @@ examples: :conf:`en-PL`, :conf:`pseudo-LOCALE`, :conf:`pseudolocalization` or :c
 
 PseudoLocale string have to be in :conf:`locales` config as well. Otherwise no folder is going to be created.
 Pseudolocalized text is created on  ``yarn compile`` command.
-The pseudolocalization is automatically created from messages in order specified in :ref:`Preparing catalogs for production`.
-It can also be changed by setting translation in :conf:`message.json` into non-pseudolocalized text.
+The pseudolocalization is automatically created on ``yarn extract`` from messages in order specified in :ref:`Preparing catalogs for production`.
 
 How to switch your browser into specified pseudoLocale
 ------------------------------------------------------
--- a/docs/tutorials/cli.rst
+++ b/docs/tutorials/cli.rst
@@ -198,7 +198,7 @@ To setup pseudolocalization add :conf:`pseudoLocale` in ``package.json``::
 :conf:`pseudoLocale` option can be any string 
 examples: :conf:`en-PL`, :conf:`pseudo-LOCALE`, :conf:`pseudolocalization` or :conf:`en-UK`
 
-PseudoLocale string have to be in `locale` config as well. Otherwise no folder is created.
+PseudoLocale string have to be in `locale` config as well. Otherwise no folder is going to be created.
 Pseudolocalized text is created on  ``yarn compile`` command.
 The pseudolocalization is automatically created from TODOTODOTODO.
 It can also be changed by setting translation in :conf:`message.json` into non-pseudolocalized text.
--- a/docs/tutorials/cli.rst
+++ b/docs/tutorials/cli.rst
@@ -103,6 +103,13 @@ plural rules to a production ready catalog::
 The ``locale`` directory now contains the source catalogs (``messages.json``) and
 the compiled ones (``messages.js``).
 
+Messages added to compiled file are collected in specific order:
+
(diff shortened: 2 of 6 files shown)


docs/tutorials/cli.rst
```diff
@@ -198,9 +198,9 @@ To setup pseudolocalization add :conf:`pseudoLocale` in ``package.json``::
 :conf:`pseudoLocale` option can be any string 
 examples: :conf:`en-PL`, :conf:`pseudo-LOCALE`, :conf:`pseudolocalization` or :conf:`en-UK`
 
-PseudoLocale folder is automatically created based on configuration when running 
-``yarn extract`` command. Pseudolocalized text is created on  ``yarn compile`` command.
-The pseudolocalization is automatically created from default messages. 
+PseudoLocale string have to be in `locale` config as well. Otherwise no folder is going to be created.
+Pseudolocalized text is created on  ``yarn compile`` command.
+The pseudolocalization is automatically created from TODOTODOTODO.
```
MartinCerny-awin, line 203: Do we have already in documentation what is the prioritisation of compiling messages?

tricoder42: I don't think so. It's:

1. translation
2. fallback (using the new `fallbackLocales`)
3. message key

MartinCerny-awin: Wouldn't it be worth adding it to documentation? I just can't figure out where it could fit

Bertg:
Would these changes resolve #838 ?

docs/tutorials/cli.rst
tricoder42: Yeah, absolutely. Somewhere inside `compile` docs? Because it ultimately happens at compile step.

tricoder42 (commented):
What's the status of this PR? I see you've removed some code, but I don't see anything that replaces it. That code was just obsolete?

packages/cli/src/lingui-compile.ts
```diff
@@ -52,7 +49,7 @@ function command(config, options) {
 
     catalogs.forEach((catalog) => {
       const messages = catalog.getTranslations(
-        locale === config.pseudoLocale ? config.sourceLocale : locale,
```
MartinCerny-awin, line 55: this change means that we will pseudolocalize messages in the order as you were describing

1. translation
2. fallback (using the new fallbackLocales)
3. message key

before we were automatically using sourceLocale

MartinCerny-awin (commented):
I will improve documentation and then it can be merged

packages/cli/src/lingui-compile.ts
tricoder42: Nice, thank you!

docs/tutorials/cli.rst
```diff
@@ -198,9 +205,9 @@ To setup pseudolocalization add :conf:`pseudoLocale` in ``package.json``::
 :conf:`pseudoLocale` option can be any string 
 examples: :conf:`en-PL`, :conf:`pseudo-LOCALE`, :conf:`pseudolocalization` or :conf:`en-UK`
 
-PseudoLocale folder is automatically created based on configuration when running 
-``yarn extract`` command. Pseudolocalized text is created on  ``yarn compile`` command.
-The pseudolocalization is automatically created from default messages. 
+PseudoLocale string have to be in `locale` config as well. Otherwise no folder is going to be created.
+Pseudolocalized text is created on  ``yarn compile`` command.
+The pseudolocalization is automatically created from messages in order specified in "LINK TO Preparing catalogs for production".
```
MartinCerny-awin, line 210: How can I link to added compile section above?

semoal: https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html

tricoder42: Every heading is an anchor that you can reference with `:ref:`:

```suggestion
The pseudolocalization is automatically created from messages in order specified in :ref:`Preparing catalogs for production`.
```

You could also customize the link if you want to spell it differently:

```
The pseudolocalization is automatically created from messages in order specified in :ref:`section above<Preparing catalogs for production>`.
```

docs/ref/conf.rst
```diff
@@ -372,7 +372,8 @@ pseudoLocale
 Default: ``""``
 
 Locale used for pseudolocalization. For example when you set ``pseudoLocale: "en"``
-then all messages in ``en`` catalog will be pseudo localized.
+then all messages in ``en`` catalog will be pseudo localized. The locale has to be included
+in `locales <https://lingui.js.org/ref/conf.html#locales>`_ config.
```
MartinCerny-awin, line 376: is this the correct linking?

semoal: You can test it doing:

```
cd docs/
pip install pipenv
pipenv install
pipenv run make html
```

You shouldn't include the domain because we have different domains for each deploy.

tricoder42: @MartinCerny-awin You can use `:conf:` role to link specific configuration options:

```suggestion
in :conf:`locales` config.
```

MartinCerny-awin: THanks guys

docs/tutorials/cli.rst
```diff
@@ -198,9 +205,9 @@ To setup pseudolocalization add :conf:`pseudoLocale` in ``package.json``::
 :conf:`pseudoLocale` option can be any string 
 examples: :conf:`en-PL`, :conf:`pseudo-LOCALE`, :conf:`pseudolocalization` or :conf:`en-UK`
 
-PseudoLocale folder is automatically created based on configuration when running 
-``yarn extract`` command. Pseudolocalized text is created on  ``yarn compile`` command.
-The pseudolocalization is automatically created from default messages. 
+PseudoLocale string have to be in `locale` config as well. Otherwise no folder is going to be created.
```
tricoder42, line 208:
```suggestion
PseudoLocale string have to be in :conf:`locales` config as well. Otherwise no folder is going to be created.
```

docs/tutorials/cli.rst
```diff
@@ -198,9 +205,9 @@ To setup pseudolocalization add :conf:`pseudoLocale` in ``package.json``::
...
+Pseudolocalized text is created on  ``yarn compile`` command.
+The pseudolocalization is automatically created from messages in order specified in :ref:`Preparing catalogs for production`.
 It can also be changed by setting translation in :conf:`message.json` into non-pseudolocalized text.
```
tricoder42, line 211: Not sure what link was supposed to be here but I guess it should be sth like this:

```suggestion
It can also be changed by setting translation in catalog into non-pseudolocalized text.
```

tricoder42 (commented):
Looking good! I've just did a last check and found two small typos. Could you please check?

docs/tutorials/cli.rst
MartinCerny-awin: I think we should remove this sentence. It was supposed to show that users are able to edit message.json file manually. Considering that it would get removed after running compile then I do not find it useful.

MartinCerny-awin: I have removed the sentence

MartinCerny-awin:
I am happy with the changes. I have tested them and it looks to be working for me.

tricoder42 (approved):

Issue #834, opened 2020-11-07
Edge case may leave I18nProvider un-rendered


Bertg:
**Describe the bug**

When translation are loaded async and out of react render loop it is possible that ```I18nProvider``` never renders.

**To Reproduce**

```tsx
import { i18n } from "@lingui/core";
import { detect, fromStorage, fromNavigator } from "@lingui/detect-locale";
import { I18nProvider } from "@lingui/react";
import { t } from "@lingui/macro";

const getDefaultLocale = () => {
  return detect(fromStorage("lang"), fromNavigator(), () => "en") as string;
};


const buildDynamicActivate = (i18n: any) => async (locale: string) => {
  const { messages } = await import(`../../locales/${locale}/messages.po`);
  i18n.load(locale, messages);
  i18n.activate(locale);
};

const activateLocale = buildDynamicActivate(i18n);
i18n.loadLocaleData("en", { plurals: en });
activateLocale(getDefaultLocale());

export App = () => {
  render (
    <I18nProvider>
      <span>{t`A translation`}</span>
    <I18nProvider/>
  )
}
```

In our code base this, more often than not yield a blank page.

This is due to this piece of code in https://github.com/lingui/js-lingui/blob/8979aaf81e5f839a8406d3ac7516205113944c39/packages/react/src/I18nProvider.tsx

```tsx
React.useEffect(() => {
  const unsubscribe = i18n.on("change", () => setContext(makeContext()))
  return () => unsubscribe()
}, [])

const renderKey = forceRenderOnLocaleChange && i18n.locale
if (forceRenderOnLocaleChange && !renderKey) return null
```

It is perfectly possible that the ```i18n.activate``` call occurs after the render key is calculated, but before React invokes the function defined in the ```useEffect``` hook. 

It took quite some time to figure this out, as it is very unexpected for a Provider Node like this to refuse to render. Further more this is no warning or default placeholder text at all.

**Expected behavior**

* The ```I18nProvider``` should never get stuck in a not rendering state
* There should be a warning or error (console or on screen) for the developer to show what is happening

**Additional context**

- jsLingui version 3.0.3

I fixed this in our source code in a very ugly but reliable way: 

```tsx
import React, { useEffect, useState } from "react";
import { I18nProvider as LinguiI18nProvider } from "@lingui/react";

export const I18nProvider = ({ children }: { children: React.ReactNode }) => {
  const [f, setRefreshValue] = useState(0),
[...]

PR #839, opened 2020-11-09
Ensure render of I18nProvider in async scenarios

Bertg:
This is related to #834

This PR solves 2 issues:

* Making sure the component gets re-rendered when needed
* Provide a warning to the developer

Some more thoughts about this PR and problem:

* It is perfectly possible for the warning to appear even if the component will render. 
* It would be better to always render. We can add a warning to the developer that only defaults will be rendered. This requires the developer to make sure a language is activated before they render the component. We could even choose to completely throw if i18n is not activated yet, unless the developer set a prop indicating they are OK with defaults.
* ```forceRenderOnLocaleChange``` should be separated from ```dontRenderIfNoLocaleIsActivated```
* ideally there should be a ```i18n.on("activated" ``` which also triggers when first observed (if already triggered before). That way the extra call to ```setRenderKey``` in the effect is never needed.

### DIFF
--- a/packages/react/src/I18nProvider.tsx
+++ b/packages/react/src/I18nProvider.tsx
@@ -74,9 +74,15 @@ export const I18nProvider: FunctionComponent<I18nProviderProps> = ({
    * I18n object from `@lingui/core` is the single source of truth for all i18n related
    * data (active locale, catalogs). When new messages are loaded or locale is changed
    * we need to trigger re-rendering of LinguiContext.Consumers.
+   *
+   * We call `setContext(makeContext())` after adding the observer in case the `change`
+   * event would already have fired between the inital renderKey calculation and the
+   * `useEffect` hook being called. This can happen if locales are loaded/activated
+   * async.
    */
   React.useEffect(() => {
     const unsubscribe = i18n.on("change", () => setContext(makeContext()))
+    setContext(makeContext())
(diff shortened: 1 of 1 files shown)


tricoder42:
Perfect, thanks!

> ideally there should be a i18n.on("activated" 

Let's try it as it is and figure out how to improve it later. I guess we could add a signal `activated` or `loaded` which would be triggered after message catalog for active locale is loaded and available.

Bertg:
Quick question... may I be added to the contributor list? :)

tricoder42:
@Bertg I don't think we have an explicit list of contributors 🤔  The [contributors](https://github.com/lingui/js-lingui#contributors) section in README is generated automatically from [repository contributors](https://github.com/lingui/js-lingui/graphs/contributors). I guess it's just cached and you should appear there in few days.

And if I haven't said it explicitly, thank you for your help! I really appreaciate everyone who's willing to spend their free time contributing to this project 🙏

Bertg:
You're welcome :)

Actually felt kind of embarrassed asking, but this is a project I would be proud of having my name attached to.

tricoder42:
Yeah, this is a valid point 👍  We could use GitHub Action to automatically generate list of contributors after each pull request. I'm gonna check it later. GitHub Actions are really powerfull 🚀

PR #840, opened 2020-11-09
ci: remove old ci

semoal:


semoal:
I'm downloading on Parallels Desktop a windows image to see why that test is crashing on Windows :/

I'll fix that, and we'll be ready to merge this

tricoder42:
Yeah, I'm not very wise from the output... Looks like different line endings? Dunno...

semoal:
> Yeah, I'm not very wise from the output... Looks like different line endings? Dunno...

Were two issues, on Windows TZ=UTC doesn't work, you need to run some commands on the console to get it working, i've added a helper to add UTC automatically on each test. 

About the .po write test on Windows, was failing because mockFs or Jest dunno, adds like middle points to multiline strings

I only found this issue https://github.com/facebook/jest/issues/9559, we test the content by the moment and I'll investigate further this weeks

tricoder42:
So weird this isn't hapenning in Appveyor... We could keep the appveyor for now and resume working on this when you have time. It's not critical issue

semoal:
> So weird this isn't hapenning in Appveyor... We could keep the appveyor for now and resume working on this when you have time. It's not critical issue

I downloaded Parallels and I'm trying directly yarn test inside a real Windows 10, looks like appveyor is not as real as we tough or some windows distro where this doesnt happen

semoal:
We're 🟢 , let's :shipit:

tricoder42:
And the ship has sailed 🚢

PR #843, opened 2020-11-10
chore: added token to deploy

semoal:
We'll try with secrets.GH_TOKEN that is the LinguiBot Account, if that doesn't work we'll change to GITHUB_TOKEN.. that should work properly.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -66,6 +66,8 @@ jobs:
           token: '${{ secrets.GH_TOKEN }}'
 
       - name: Release packages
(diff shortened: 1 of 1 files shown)


tricoder42:
As I understand it, `GITHUB_TOKEN` is for the GitHub Actions account/user. `GH_TOKEN` is our Lingui Bot

semoal:
> As I understand it, `GITHUB_TOKEN` is for the GitHub Actions account/user. `GH_TOKEN` is our Lingui Bot

Yes, exactly that.

Issue #844, opened 2020-11-10
Function `t` macros do not get extracted


Bertg:
**Describe the bug**

When extracting translations, macros like ```t({id: "Hello"})``` do not get correctly extracted. 
Instead a ```msgid "undefined"``` is added.

**To Reproduce**

Using lingui 3.1.0 add a translation like this in source: 

```ts
t({id: "Hello"})
```

run ```lingui extract```

Find the wrong translation in the file. Also note that the line number is missing.

```po
#: src/pages/Page.tsx:
msgid "undefined"
msgstr ""
```

**Expected behavior**

Add the correct string to the PO file.

**Additional context**

- jsLingui version 3.1.0

PR #846, opened 2020-11-10
fix: t macro as function not extracting

semoal:
Will close https://github.com/lingui/js-lingui/issues/844

### DIFF
--- a/packages/macro/test/js-t.ts
+++ b/packages/macro/test/js-t.ts
@@ -86,22 +86,19 @@ export default [
     name: "Support id and comment in t macro as callExpression",
     input: `
         import { t } from '@lingui/macro'
-        t({
-          id: 'msgId_2',
-          message: 'text',
-          comment: 'description for translators'
-        })
-        t({ id: 'msgId', comment: 'description for translators', message: plural(val, { one: '...', other: '...' }) })
+        const msg = t({ id: 'msgId', comment: 'description for translators', message: plural(val, { one: '...', other: '...' }) })
       `,
-    expected: `
-      import { i18n } from "@lingui/core"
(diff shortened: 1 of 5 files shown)


packages/macro/test/js-t.ts
```diff
@@ -86,22 +86,19 @@ export default [
...
-      /*i18n*/
-      i18n._({ id: "msgId_2", message: 'text', comment: 'description for translators' })
-      
+    expected: `import { i18n } from "@lingui/core";
+    const msg =
       /*i18n*/
-      i18n._({ id: "msgId", comment: 'description for translators', message: '{val, plural, one {...} other {...}}', values: {
-        val: val,
-      } })
+      {
```
tricoder42, line 94: Actually, the descriptor must be wrapped in `i18n._` call. Otherwise the macro would be the same as `defineMessage`. I think the issue is that the `/* i18n */` comment must be *inside*:

```diff
- /*i18n*/i18n._({ id: "msgId_2", message: 'text', comment: 'description for translators' })
+ i18n._(/*i18n*/{ id: "msgId_2", message: 'text', comment: 'description for translators' })
```

Let me check how to fix it

tricoder42:
@semoal I've pushed one commit which seems to fix the issue. Please take a look and merge at your will 👍

packages/macro/test/js-t.ts
semoal: Yes, that was the issue.

tricoder42:
Do you want to release the patch or should we wait for more fixes?

semoal:
Let me take a look to that windows issue. This morning i could get it fixed, if not we just deploy the patch this afternoon

PR #850, opened 2020-11-11
docs: fix location of date/number formatters

Snaptags:
This remark is misleading. Core does not export date or number


tricoder42:
Good catch! Thank you 👍

PR #853, opened 2020-11-11
fix: absolute rootDir on Windows

semoal:
It's a bit workaround, I' have to look deeper to analyze the issue, but meanwhile it works on Windows and Mac as usual 👍🏻

### DIFF
--- a/examples/next-js/locale/cs/messages.po
+++ b/examples/next-js/locale/cs/messages.po
@@ -13,47 +13,47 @@ msgstr ""
 "Language-Team: \n"
 "Plural-Forms: \n"
 
-#: pages/index.tsx:38
-#~ msgid "Checkout LinguiJS examples and usecases."
-#~ msgstr "Podívej se na ukázky a příklady LinguiJS"
+#: pages/index.tsx:36
+msgid "Checkout LinguiJS examples and usecases."
+msgstr "Podívej se na ukázky a příklady LinguiJS"
 
-#: pages/index.tsx:23
-#~ msgid "Documentation →"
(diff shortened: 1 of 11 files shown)


examples/next-js/locale/cs/messages.po
```diff
@@ -14,46 +14,46 @@ msgstr ""
 "Plural-Forms: \n"
 
 #: pages/index.tsx:38
-msgid "Checkout LinguiJS examples and usecases."
-msgstr "Podívej se na ukázky a příklady LinguiJS"
+#~ msgid "Checkout LinguiJS examples and usecases."
```
tricoder42, line 17: It seems like it doesn't find any messages as they're all marked as obsolete.

tricoder42 (commented):
Thank you for working on this @semoal, but the fix doesn't seem to work. No messages were extracted, so it looks like the path isn't correct.

semoal:
> Thank you for working on this @semoal, but the fix doesn't seem to work. No messages were extracted, so it looks like the path isn't correct.

Ugh i take a further look then

semoal:
Finally fixed, was an issue with glob library that says:
`
Results from absolute patterns such as /foo/* are mounted onto the root setting using path.join. On windows, this will by default result in /foo/* matching C:\foo\bar.txt.
`

So on Mac was working properly but on windows was resolving to root of the system..

semoal:
With this fixed, we'ready to ship :shipit:

tricoder42 (approved):
Amazing work! Thanks for that 👍

PR #854, opened 2020-11-11
feat: extract multiple comments per translation ID

### LINKED ISSUE #845: Multiple comments get ignored for same ID
**Describe the bug**

A given translation ID may occur multiple times in the source code. 
It is reasonable that some of these invocations have comments and other not, some comments may even be different.
Lingui only uses the comment of the first instance it find. If that instance does not have a comment, no comment is added to the PO file.

**To Reproduce**

Given the following file

```tsx
<Trans id="long_value" comment="First comment" />
<Trans id="long_value" comment="Second comment, does not work yet" />
```

when running ```lingui extract```

We get 

```po
#. First comment
#: src/pages/Lingui.tsx:62
#: src/pages/Lingui.tsx:69
#: src/pages/Lingui.tsx:74
msgid "long_value"
msgstr ""
```

**Expected behavior**

The PO file should contain all comments:

```po
#. First comment
#: src/pages/Lingui.tsx:62
# Second comment, does not work yet
#: src/pages/Lingui.tsx:69
#: src/pages/Lingui.tsx:74
msgid "long_value"
msgstr ""
```

**Additional context**

- jsLingui version 3.1.0

Bertg:
This PR resolves #845 

On par file level is collects multiple ```extractedComments``` (using the same name as ```pofile``` package)
This - just like origins - gets merged together to build the catalogue.

Some refactoring was done to use the ```extractedComments``` name throughout.

**Possible issues**

This change should just work for PO file users.
However if users are using lingui file type thing might break. Namely the ```description (string)``` is replaced with ```extractedComments (string[])```. Either some backwards compatibility should be built on top of these changes, or sufficient change information should be provided.

**What can be improved***

Currently the PO file gets built like with ```extractedComments``` and ```references``` grouped: 

```po
#. Comment in code
#. Other comment
#: App.js:12
#: Screen.js:56
# msgid "Nice"
```

It would be cool to output them with comment associated with the reference file.

```po
#: App.js:12
#. Comment in code
#: Screen.js:56
#. Other comment
# msgid "Nice"
```

However, it seems that PO file here is major blocker and this change would be way to much work to achieve.

### DIFF
--- a/packages/babel-plugin-extract-messages/src/index.ts
+++ b/packages/babel-plugin-extract-messages/src/index.ts
@@ -56,7 +56,7 @@ export default function ({ types: t }) {
     t.isIdentifier(node.object, { name: "i18n" }) &&
     t.isIdentifier(node.property, { name: "_" })
 
-  function collectMessage(path, file, props) {
+  function collectMessage(path, file, {comment, ...props}) {
     const messages = file.get(MESSAGES)
 
     const rootDir = file.get(CONFIG).rootDir
@@ -64,7 +64,7 @@ export default function ({ types: t }) {
       .relative(rootDir, file.opts.filename)
       .replace(/\\/g, "/")
     const line = path.node.loc ? path.node.loc.start.line : null
-    props.origin = [[filename, line]]
+    props.origin = [[filename, line, comment]]
 
     addMessage(path, messages, props)
   }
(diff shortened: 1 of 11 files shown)


tricoder42:
> However if users are using lingui file type thing might break.

Well, `lingui` format was originally just a json dump of the whole catalog, so we could leave there comments as an array.

tricoder42 (approved):
Looks great! Thank you for working on this! 👍

PR #855, opened 2020-11-12
docs: add loading plurals to testing.rst

Snaptags:
We should add plural loading, testing library will log warnings about that to the console otherwise.
P.S. Thanks for adding a page about testing to the documentation!


tricoder42 (approved):

tricoder42:
Nice, thank you for reviewing and improving docs! There's definitely lot of room for improvement

PR #858, opened 2020-11-12
fix: @lingui/cli not compiling if locale contains empty strings

rhnorskov:
Should a locale contain a key which has not been translated yet, and is therefor an empty string. The CLI will now wrongfully fallback to the fallback locale.

It is caused by `getTranslation(locale)` being evaluated falsy due to it being an empty string, I've therefore replaced the logical OR operator with the null-coalescing operator, to only fallback if `getTranslation(locale)` is nullish.

### DIFF
--- a/packages/cli/src/api/catalog.ts
+++ b/packages/cli/src/api/catalog.ts
@@ -273,19 +273,19 @@ export class Catalog {
 
     return (
       // Get translation in target locale
-      getTranslation(locale) ||
+      getTranslation(locale) ??
       // We search in fallbackLocales as dependent of each locale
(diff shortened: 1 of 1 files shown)


semoal (approved):

tricoder42:
Hey @rhnorskov, could you please post a few examples? I don't quite understand the problem.

> Should a locale contain a key which has not been translated yet, and is therefor an empty string. The CLI will now wrongfully fallback to the fallback locale.

Does it mean that when the translation is missing (i.e. it's a empty string) it doesn't load translation from fallback locale?

PR #860, opened 2020-11-13
chore: deploy from build inside each package

semoal:
I don't why but I get used to see the build/ folder on rootDir and didn't though that lerna deploys dist folder of each package. `

I changed the scripts to build the pack inside each package, and adjusted integration testing to provide from there.

### DIFF
--- a/scripts/release.js
+++ b/scripts/release.js
@@ -10,7 +10,7 @@ const semver = require("semver")
 const R = require("ramda")
 
 const BUILD_DIR = "build/"
-const PACKAGES_DIR = "build/packages"
+const PACKAGES_DIR = "packages/build"
 
 const npmTagForBranch = {
   main: "latest",
--- a/.github/workflows/deploy-main-docs.yml
+++ b/.github/workflows/deploy-main-docs.yml
@@ -41,4 +41,4 @@ jobs:
         run: |
           cd docs/
           ./vercel-build.sh
-          npx vercel --public  --prod --token ${{secrets.NOW_TOKEN}} --name lingui-docs _build/html
+          npx vercel --public --yes --prod --token ${{secrets.NOW_TOKEN}} --name lingui-docs _build/html
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ node_modules/
 build/packages/
 .yalc/
 yalc.lock
+packages/**/build
 docs/_build/
 
 npm-debug.log
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-  "version": "3.2.0",
+  "version": "3.2.1",
   "packages": [
     "packages/*"
   ],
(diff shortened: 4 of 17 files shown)


semoal:
We could merge this, and try with your fix to patch 3.2.2

scripts/release.js
```diff
@@ -10,7 +10,7 @@ const semver = require("semver")
 const R = require("ramda")
 
 const BUILD_DIR = "build/"
-const PACKAGES_DIR = "build/packages"
+const PACKAGES_DIR = "packages/build"
```
tricoder42, line 13: Isn't the build folder in each package now? e.g. `packages/core/build`? I guess we'll need to refactor this script to use lerna for publishing... Or leave it as it is when shit hits the fan.

tricoder42 (commented):
Looks good. Did you have a chance to test it locally?

tricoder42:
> We could merge this, and try with your fix to patch 3.2.2

I don't want to rush the release. The bug isn't severe, there's a simple workaround, so I would rather double check the release process before we continue. I don't have much time to work on it now, I'll be available tomorrow

scripts/release.js
semoal: Yes, release script is outdated, with the ci process is enough. This file changed because i've did a full text replace

```
yarn release:build
yarn version:latest
yarn publish:latest
```

tricoder42: Cool. If it's possible to run these commands locally to publish new version manually, then we can remove the `release.js` file 👍

semoal: Yes, I'll work on another branch about what we spoke this morning. Verdaccio + testing examples with verdaccio version, I'll rename release.js, to local-release.js, and i'll write there the logic.

semoal:
> Looks good. Did you have a chance to test it locally?

Yes, i've tested to build i've chcked that build folders were correctly "npm packed and unpacked" for package.json version, and all should work properly after that.

I'll start this weekend, to create that local-release.js to verdaccio for testing locally, shouldn't be complex.

semoal:
```bash

lerna notice === Tarball Contents ===
lerna notice 1.1kB LICENSE
lerna notice 3.8kB index.js
lerna notice 951B  package.json
lerna notice 1.0kB README.md
lerna notice === Tarball Details ===
lerna notice name:          @lingui/loader
lerna notice version:       3.2.2
lerna notice filename:      lingui-loader-3.2.2.tgz
lerna notice package size:  3.1 kB
lerna notice unpacked size: 6.8 kB
lerna notice shasum:        0da06c5f45db022152c699fd47da0fb90539e30d
lerna notice integrity:     sha512-IwK1rR1s6Z4nD[...]oKgrgPKOgOcOw==
lerna notice total files:   4
lerna notice
lerna http fetch PUT 201 http://0.0.0.0:4873/@lingui%2floader 99ms
Successfully published:
```

With the new strategy the publish action works perfect Tom, I'll create anyways that little script to test the packages with Verdaccio in a automated way.

PR #862, opened 2020-11-13
fix: parse template strings in t and defineMessage macros

### LINKED ISSUE #861: Function `t` macros do not extract template strings
**Describe the bug**
You cannot use template strings with the t function:
```
t({message: `template`})
```
results in
```
msgid "undefined"
msgstr "undefined"
```

**Expected behavior**
```
msgid "template"
msgstr "template"
```

**Additional context**

- jsLingui version 3.2.1
- Babel version 7.12.3
- Your Babel config: Create React App

tricoder42:
Closes #861

### DIFF
--- a/packages/macro/test/js-defineMessage.ts
+++ b/packages/macro/test/js-defineMessage.ts
@@ -36,11 +36,11 @@ export default [
     `,
   },
   {
-    name: "should left string message intact - template literal",
+    name: "should transform template literals",
     input: `
         import { defineMessage } from '@lingui/macro';
         const message = defineMessage({
-          message: \`Message\`
+          message: \`Message \${name}\`
         })
     `,
     expected: `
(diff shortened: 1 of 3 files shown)


semoal (approved):

packages/macro/test/js-defineMessage.ts
```diff
@@ -36,19 +36,22 @@ export default [
     `,
   },
   {
-    name: "should left string message intact - template literal",
+    name: "should transform template literals",
```
tricoder42, line 39: Normally thiw would be a breaking change, but since the code didn't work as described in documentation, I believe it's just a fix ;)

semoal: I hope so 🤪

tricoder42:
Also we should improve the extract plugin so it throws a warning/error when it tries to extract message with `undefine` id.

PR #864, opened 2020-11-13
docs: suggest removal of types in migration docs

Snaptags:
I just noticed that TypeScript definition files are included now! Thanks for that.
We should encourage people to keep their dependency list short, so I suggest adding that to the "new features" section.


tricoder42:
Good point! Thank you for mentioning this!

PR #867, opened 2020-11-15
chore: Verdaccio testing

semoal:
Here it is 🍰 

- An example with a forced failing test: https://github.com/lingui/js-lingui/runs/1401346525?check_suite_focus=true (changed the test to fail) 🔴 
- I fix the test: [`639b632` (#867)](https://github.com/lingui/js-lingui/pull/867/commits/639b6328dc75f3489158d550530ae9ef7abddc48) 🛌🏻 
- And.. the release workflow test...:  https://github.com/lingui/js-lingui/runs/1401351821 🟢 

We can run it on each pull request because it's pretty fast and we ensure any user breaks the release step. 👍🏻


scripts/verdaccio-integration.js
```diff
@@ -0,0 +1,52 @@
...
+  cwd: path.resolve("examples/create-react-app")
+}
+
+async function main() {
+  const spinner = ora()
+
+  spinner.start("Linking create-react-app")
+  await exec('npm i -g update-by-scope --registry https://registry.npmjs.org', OPTS)
+  await exec('npm config set registry http://0.0.0.0:4873/', OPTS)
+  await exec('update-by-scope @lingui', OPTS)
```
tricoder42, line 15: 👀 Very useful!

tricoder42 (approved):
This is amazing! Beatiful and elegant. Especially all those git mumbo jambo commands :)

tricoder42:
:shipit:

PR #868, opened 2020-11-16
chore: fix workflow versioning

semoal:
The problem was that running build and "npm pack", before versioning ocurrs that the package.json of build/ folder is not the new version.

So when releasing fails because /build/package.json is 3.2.1, and not the new version

Also changed the command to update all packages, instead of just upgrading the packages that has changed.


tricoder42:
> Also changed the command to update all packages, instead of just upgrading the packages that has changed.

Yes! Thank you! 👍

tricoder42 (approved):

PR #871, opened 2020-11-16
fix: locale not present in catalogs warn

### LINKED ISSUE #869: Compile not working with unstranslated strings
I am having an issue compiling messages, when some strings are untranslated. I am not able report one specific bug, as I am unsure what is intended behaviour and what is not. So to explain the issue I will walk through what I have tried, and hopefully it will make sense what the issue it.

So I created a fresh create-react-app following the instructions in the documentation. I then changed the locales, shown in this configuration. 
```JSON
{
  "locales": [
    "en-US",
    "en-CA",
    "fr-CA"
  ],
  "sourceLocale": "en-US",
  "catalogs": [
    {
      "path": "src/locales/{locale}/messages",
      "include": [
        "src"
      ]
    }
  ],
  "format": "po"
}
```
I then ran `lingui extract` and `lingui compile` and that is when the first error appeared:
```
/.../node_modules/@lingui/cli/api/catalog.js:218
        return catalogs[locale][key].translation;
                               ^

TypeError: Cannot read property 'hello.world' of undefined
```
I then looked into the cause of this error, and it is caused by `getTranslation(locale)` being evaluated as falsy due to the empty string, it is then attempted to get the translation from the fallback locale `getMultipleFallbacks(locale)` however, default value for `fallbackLocales` with my locales is contains: 

```JSON
{ 
  "en-US": "en",
  "en-CA": "en"
}
```
It therefore fails, since locale `en` does not exist.
Which from the documentation is to be expected. I then attempted to set `fallbackLocales` to `false`, as also mentioned in the documentation. However that just provoked a new error
```
/.../node_modules/@lingui/cli/api/catalog.js:223
        var fL = fallbackLocales[locale]; // some probably the fallback will be undefined, so just search by locale

TypeError: Cannot read property 'en-CA' of null
```
This error looks to be caused by lack of check if `fallbackLocales` is defined or not.

I then tried setting the default fallback locale.
```JSON
"fallbackLocales": {
  "default": "en-US"
},
```
But that had no effect as for some reason the `fallbackLocales` object still contained the default values and now looked like this
```JSON
{ 
  "default": "en-US",
  "en-US": "en",
  "en-CA": "en"
}
```
I then tried to explicitly override the default values:
```JSON
"fallbackLocales": {
  "default": "en-US",
  "en-US": "en-US",
  "en-CA": "en-US"
},
```
However `fallbackLocales` remained unchanged.

So now I am at a loss. I am more than happy to correct some of these errors myself and submit a PR, but I can not do so before I know how it should work.

semoal:
Will close #869

### DIFF
--- a/packages/conf/src/index.ts
+++ b/packages/conf/src/index.ts
@@ -278,7 +278,7 @@ export function fallbackLanguageMigration(
 
   config.locales.forEach((locale) => {
     const fl = getCldrParentLocale(locale.toLowerCase())
-    if (fl) {
+    if (fl && !config.fallbackLocales[locale]) {
       config.fallbackLocales = {
         ...config.fallbackLocales,
         [locale]: fl
--- a/packages/cli/src/api/catalog.ts
+++ b/packages/cli/src/api/catalog.ts
@@ -252,7 +252,21 @@ export class Catalog {
       console.error(`Message with key ${key} is missing in locale ${locale}`)
     }
 
-    const getTranslation = (locale) => catalogs[locale][key].translation
+    const getTranslation = (locale) => {
+      const configLocales = this.config.locales.join('", "')
+      if (catalogs[locale]) {
+        return catalogs[locale][key].translation
+      }
+
+      console.warn(`
+        Catalog "${locale}" isn't present in locales config parameter
+        Add "${locale}" to your lingui.config.js:
+        {
+          locales: ["${configLocales}", "${locale}"]
+        }
+      `)
+      return null
+    }
 
     const getMultipleFallbacks = (locale) => {
       const fL = fallbackLocales[locale]
--- a/packages/conf/src/index.test.ts
+++ b/packages/conf/src/index.test.ts
@@ -173,7 +173,7 @@ describe("@lingui/conf", function () {
         ),
       })
       expect(config.fallbackLocales).toEqual({
-        "en-US": "en",
+        "en-US": ["en"],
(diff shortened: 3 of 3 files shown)


tricoder42 (approved):
Looking good 👍

packages/conf/src/index.ts
```diff
@@ -278,7 +278,7 @@ export function fallbackLanguageMigration(
 
   config.locales.forEach((locale) => {
     const fl = getCldrParentLocale(locale.toLowerCase())
-    if (fl) {
+    if (fl && !config.fallbackLocales[locale]) {
```
telpochyaotl, line 281: What if one simply defines the default fallback locale and not each individual language's fallback?

semoal: We tought that was a good idea to always build the cldr fallback and co-exist with default except if a user passes false to config. 

Anyways was a fast implementation and can be re-thinked

telpochyaotl: Appreciated, the issue is that with 3.2.1, if you set `defaultLocales` to false and at the same time you add a local that is in the cldr list and the parent is not the sourceLocale, you still get errors down the road. So it would be nice if instead, if config has defaultLocales defined in my case with { default: 'en-US' } that this config not be overridden nor enhanced ... Anyways, it's my suggestion ... I'm grateful for all the hard work nonetheless, awesome library! Thanks again!

PR #872, opened 2020-11-16
docs: remove add-locale from documentation

MartinCerny-awin:
Removes information about running deprecated add-locale command from React documentation


semoal:
Looks good to me 🥇  thxx

PR #875, opened 2020-11-17
docs: testing guide improved

semoal:
I've tried to use the testing guide and didn't worked properly, so I ended up using our integration tests


tricoder42:
Nice 👍

PR #881, opened 2020-11-23
feat: extract accepts files as cli option

### LINKED ISSUE #510: Run lingui extract for specified files
I am trying to run `lingui extract` in CI steps to warn developer if they forget to run `lingui extract`. However running `lingui extract` over the entire project takes ~10mins

`lingui extract` should take an optional glob that narrow down the list of files the command run on

lingui v3 seems to include a watch option. Maybe this is related?

**Additional context**

semoal:
Will close #510

I've adjusted also the help.ts function because when running `yarn extract src/ --verbose` was leading to confusion because params were passed also to `yarn compile` command.

On create-react-app example of Lingui projects results are really good:

### New version with files param:
`yarn extract src/ 1.47s user 0.26s system 111% cpu 1.709 total`

###  Actual code branch without files filter:
`yarn extract 4.19s user 0.33s system 135% cpu 3.345 total`

### DIFF
--- a/packages/cli/src/api/catalog.test.ts
+++ b/packages/cli/src/api/catalog.test.ts
@@ -152,7 +152,7 @@ describe("Catalog", function () {
         {
           name: "messages",
           path: "locales/{locale}",
-          include: [fixture("collect/componentA")],
+          include: [fixture("collect/componentA"), fixture("collect/componentB.js")],
           exclude: [],
         },
         mockConfig()
--- a/packages/cli/src/api/catalog.test.ts
+++ b/packages/cli/src/api/catalog.test.ts
@@ -147,6 +147,24 @@ describe("Catalog", function () {
       expect(messages).toMatchSnapshot()
     })
 
+    it("should extract only files passed on options", function () {
+      const catalog = new Catalog(
+        {
+          name: "messages",
+          path: "locales/{locale}",
+          include: [fixture("collect/componentA")],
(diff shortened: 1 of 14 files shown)


packages/cli/src/api/catalog.test.ts
```diff
@@ -147,6 +147,24 @@ describe("Catalog", function () {
       expect(messages).toMatchSnapshot()
     })
 
+    it("should extract only files passed on options", function () {
+      const catalog = new Catalog(
+        {
+          name: "messages",
+          path: "locales/{locale}",
+          include: [fixture("collect/componentA")],
```
tricoder42, line 155: Shouldn't we also add another directory to test that the `files` actually filter available files? When I comment out line `163`, the test passes anyway. Adding an extra file or directory, the snapshot should stay the same.

```suggestion
          include: [fixture("collect/componentA"), fixture("collect/componentB.js")],
```

tricoder42 (commented):
Looks great! Just one minor point, but the code works well.

tricoder42 (approved):

tricoder42:
Just another awesome feature! Good job! 👍

semoal:
> Just another awesome feature! Good job! 👍

We ship this tomorrow?

tricoder42:
> We ship this tomorrow?

Let's wait a week or so and collect other fixes/features on the way.

PR #884, opened 2020-11-23
fix: macro underscore type

### LINKED ISSUE #883: Plural numbers TypeScript error
```
  Property '_0' does not exist on type 'IntrinsicAttributes & { value?: string | number | undefined; } & TransProps & ChoiceOptions<ReactNode> & { children?: ReactNode; }'.

52             _0="There're no messages"
               ~~
```

Typescript does not allow to specify number on Plural component. 

```
          <Plural
            value={messagesCount}
            offset={1}
            _0="There're no messages"
            one="There's # message in your inbox"
            other="There're # messages in your inbox"
          />
```

I tried to have a look how to fix it, but couldn't figure it out. 

TypeScript version: 4.1.2
Lingui: 3.2.3

semoal:
Will close #883

### DIFF
--- a/packages/macro/index.d.ts
+++ b/packages/macro/index.d.ts
@@ -7,15 +7,16 @@ export function t(
   ...placeholders: any[]
 ): string
 
+export type UnderscoreDigit<T = string> = { [digit: string]: T }
 export type ChoiceOptions<T = string> = {
-  [digit: number]: T
   offset?: number
   zero?: T
   one?: T
   few?: T
   many?: T
   other?: T
-}
+} & UnderscoreDigit<T>
+
 export function plural(arg: number | string, options: ChoiceOptions): string
 export function selectOrdinal(
   arg: number | string,


packages/macro/index.d.ts
```diff
@@ -7,15 +7,16 @@ export function t(
   ...placeholders: any[]
 ): string
 
+export type UnderscoreDigit<T = string> = { [digit: string]: T }
```
tricoder42, line 10: Isn't the problem after all that previously the `digit` was a `number`, now it's a `string`? That would make probably sense, because all object keys are strings (or normalized to strings).

semoal: We can't do this, because T as string is not assignable to offset as number

```
export type ChoiceOptions<T = string> = {
	offset?: number;
	zero?: T;
	one?: T;
	few?: T;
	many?: T;
	other?: T;
	[digit: string]: T;
};
```

semoal: and with this: 
```
export type ChoiceOptions<T = string> = {
	offset?: number;
	zero?: T;
	one?: T;
	few?: T;
	many?: T;
	other?: T;
	[digit: number]: T;
};

const Plural: React.ComponentType<ChoiceProps> = (props) => {
	return <div>{props["_0"]}</div>;
};

class App extends React.Component {
	render() {
		return <Plural _0="one" />;
	}
}
```

_0 fails because typescript interprets it's a string

semoal: https://codesandbox.io/s/react-typescript-playground-forked-becf6?file=/src/index.tsx:463-645

tricoder42: Got it, thank you!

Issue #886, opened 2020-11-24
Using development version in Contributing file is outdated


MartinCerny-awin:
Recently we have removed `scripts/release.js` file, but CONTRIBUTING.md file with description how to use development version in your project did not get updated.

https://github.com/lingui/js-lingui/blob/main/CONTRIBUTING.md#using-development-version-in-your-project

When I try to run `yarn release --dev` I get
```
$ node ./scripts/release.js --dev
internal/modules/cjs/loader.js:883
  throw err;
  ^

Error: Cannot find module '/Volumes/sites/lingui/scripts/release.js'
```

semoal:
Yess, good point Martin I'll create a pr later

semoal:
Done! Thanks Martin for notifying

Issue #743, opened 2020-09-25
[Pseudo Localization] Syntax tokens shouldn't be pseudo localized


nathanforce:
**Describe the bug**

Pseudolocalization mutates syntax tokens, for example in the follow `select` syntax the key "screener" is mutated as well as the `select` keyword.

```
Error: Can't parse message. Please check correct syntax: "Ŷōũŕ ƥŕōvĩďēŕ śēńţ à {0, śēĺēćţ, śćŕēēńēŕ {ŵēĺĺńēśś àśśēśśmēńţ} other {{assessmentType}}} ƒōŕ ŷōũ ţō ţàķē. Ƥĺēàśē ţàķē à ƒēŵ mĩńũţēś ţō ćōmƥĺēţē ţĥĩś àśśēśśmēńţ ƀēƒōŕē ŷōũŕ ńēxţ àƥƥōĩńţmēńţ."
```

**To Reproduce**
Steps to reproduce the behavior, possibly with minimal code sample, e.g:

```
import { Trans } from "@lingui/react"

export default function App() {
   return (
    <div>
      Your provider sent a{' '}
            <Select
              value={isScreenerAssessment ? 'screener' : 'other'}
              screener="wellness assessment"
              other={assessmentType}
            />{' '}
            for you to take. Please take a few minutes to complete this
            assessment before your next appointment.
    </div>
}
```

**Expected behavior**
Syntax and tokens are unchanged, only actual language.

**Additional context**

- jsLingui version: 2.9.1

tricoder42:
This definitely shouldn't happen, thanks for report.

Could you please checkout `next` branch and test, if the problem persists in v3? If so, feel free to send PR, but please use `next` branch as a base (it should be relesed early November). Let me know if I can help somehow

MartinCerny-awin:
Hello @tricoder42 I can have a look at this issue

tricoder42:
@MartinCerny-awin Perfect, thanks a lot!

pauloborges:
Should this be really closed?

semoal:
Probably this issue is already fixed mate.

pauloborges:
Unfortunately no, the issue still happens with 3.4.0.

semoal:
oh, alright i'll take a look then

MartinCerny-awin:
I think Select is the last one that is missing. I am waiting for this issue to get fixed https://github.com/lingui/js-lingui/issues/919

@pauloborges is it Select that you are having issues with?

[...]

PR #887, opened 2020-11-24
fix: plural pseudolocalization with offset

MartinCerny-awin:
I have improved RegEx for plurals. It now correctly escapes plural with offset. This is fix for #743, but it does not close it as there is need to escape other React components like Select, SelectOrdinal & DateFormat

I was not able to test the code because of this #886

### DIFF
--- a/packages/cli/src/api/pseudoLocalize.test.ts
+++ b/packages/cli/src/api/pseudoLocalize.test.ts
@@ -20,30 +20,55 @@ describe("PseudoLocalization", () => {
     )
   })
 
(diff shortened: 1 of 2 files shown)


semoal:
Give it a try to the new contributing.md and let us know if all works properly 👍🏻

MartinCerny-awin:
@semoal it works well, I happy for the pr to be merged

tricoder42 (approved):

tricoder42:
Perfect, thanks for testing it! 👍

PR #889, opened 2020-11-25
fix: formatters exceptions throw error

### LINKED ISSUE #888: Lingui Extract Erases All Messages when JSON is Invalid
**Describe the bug**

Lingui overwrites messages.json if the messages.json is invalid on lingui extract.

**To Reproduce**

Run `lingui extract` with minimal output. Fill in values for one of the locales, and then make the json invalid (e.g.. drop a comma). Run `lingui extract` again. All of the messages will now be blank.

**Expected behavior**

Lingui should crash with an error that says the messages file is invalid json.

semoal:
Will close #888

### DIFF
--- a/packages/cli/src/api/formats/csv.ts
+++ b/packages/cli/src/api/formats/csv.ts
@@ -45,8 +45,7 @@ const csv: CatalogFormatter = {
     try {
       return deserialize(raw)
     } catch (e) {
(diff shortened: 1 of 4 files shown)


tricoder42 (approved):
Interesting! Good job hunting down this bug 👍

PR #890, opened 2020-11-26
fix: select props typescript

MartinCerny-awin:
Getting error if using example from documentation

```
 <Select value={gender} male="His book" female="Her book" other="Their book" />
```

```
Type '{ value: "male"; male: string; female: string; other: string; }' is not assignable to type 'IntrinsicAttributes & { value?: string | undefined; other?: ReactNode; } & TransProps & { children?: ReactNode; }'.
  Property 'male' does not exist on type 'IntrinsicAttributes & { value?: string | undefined; other?: ReactNode; } & TransProps & { children?: ReactNode; }'
```

### DIFF
--- a/packages/macro/index.d.ts
+++ b/packages/macro/index.d.ts
@@ -40,10 +40,26 @@ export type ChoiceProps = {
 } & TransProps &
   ChoiceOptions<ReactNode>
 
+/**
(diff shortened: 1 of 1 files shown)


tricoder42 (approved):
Looks good! 👍

PR #891, opened 2020-11-27
chore: add information how to restore default registry

MartinCerny-awin:
Added restore default registry description to CONTRIBUTING.md


semoal:
Yay, you're right, thanks mate

Issue #822, opened 2020-11-05
Configure the @lingui/core name in the macro


arcanis:
**Is your feature request related to a problem? Please describe.**

I'm abstracting Lingui behind an intermediary layer, to eases maintenance (should we need to switch to a later version, we can make sure we can do this incrementally by simply updating our own layer). Unfortunately, `@lingui/macro` seems to hardcode imports to the `@lingui/core` package, meaning that Lingui still "leaks" into the consumer code. It works on typical installs due to incorrect hoisting, but fails on stricter environment like Plug'n'Play.

**Possible solution 1 (preferred)**

Use `require.resolve('@lingui/core')` instead of hardcoding `@lingui/core`.

**Possible solution 2**

Allow to configure somehow the import names. For instance I currently do something like this:

*@mypackage/i18n/macro.js*
```ts
module.exports = require(`@lingui/macro`);
```

I could do this instead:

*@mypackage/i18n/macro.js*
```ts
module.exports = require(`@lingui/macro`).custom({
  coreModule: `@mypackage/lingui/core`),
});
```

tricoder42:
Hey @arcanis, you can configure them using [`runtimeConfigModule`](https://lingui.js.org/ref/conf.html#runtimeconfigmodule). Does it solve your problem?

tricoder42:
Feel free to reopen if [`runtimeConfigModule`](https://lingui.js.org/ref/conf.html#runtimeconfigmodule) doesn't help

arcanis:
Hey @tricoder42! This was exactly what I was looking for, thanks 👍

PR #895, opened 2020-11-29
feat: add support for runtimeConfigModule w/ Trans

arcanis:
The `runtimeConfigModule` setting cannot be used to transform the hardcoded `import {Trans} from '@lingui/react'` statements, which is problematic when the `Trans` symbol is expected to be provided by an intermediary layer (#822).

This diff addresses that by adding support for `runtimeConfigModule` being an object, in which case the key is the symbol name (either `i18n` or `Trans`), and the value is the symbol name and from where to import it. This change is completely backward compatible, only adding syntax.

The test framework doesn't seem to allow for testing `runtimeConfigModule` (its current implementation doesn't seem tested) so I couldn't add relevant tests.

### DIFF
--- a/packages/macro/src/index.ts
+++ b/packages/macro/src/index.ts
@@ -5,7 +5,27 @@ import MacroJS from "./macroJs"
 import MacroJSX from "./macroJsx"
 
 const config = getConfig({ configPath: process.env.LINGUI_CONFIG })
-const [i18nImportModule, i18nImportName = "i18n"] = config.runtimeConfigModule
+
+const getSymbolSource = (name: string) => {
+  if (Array.isArray(config.runtimeConfigModule)) {
+    if (name === "i18n") {
+      return config.runtimeConfigModule
+    } else {
+      return ["@lingui/react", name]
+    }
+  } else {
+    if (
+      Object.prototype.hasOwnProperty.call(config.runtimeConfigModule, name)
+    ) {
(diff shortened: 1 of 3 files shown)


packages/macro/src/index.ts
```diff
@@ -5,7 +5,27 @@ import MacroJS from "./macroJs"
...
+      Object.prototype.hasOwnProperty.call(config.runtimeConfigModule, name)
+    ) {
+      return config.runtimeConfigModule[name]
+    } else {
+      return ["@lingui/react", name]
+    }
+  }
+}
+
+const [i18nImportModule, i18nImportName = `i18n`] = getSymbolSource(`i18n`)
```
semoal, line 27: can't this just be simple quotes? ''

semoal:
Looks good to me, except one question that is confusing to me, thanks for your contribution Maël

packages/macro/src/index.ts
arcanis: Oh yes it can, sorry - I have the habit of using backquotes everywhere, and it seems prettier doesn't change those. Will update.

semoal (approved):

tricoder42 (approved):
Looks good, thank you!

semoal:
😱  Released 3.3.0 with this fix/feature introduced!

```
npx update-by-scope @lingui
```

PR #898, opened 2020-11-30
fix: scaped literals double backslash formatting

### LINKED ISSUE #897: Escape bracket gets escaped in messages
**Describe the bug**
In linguijs v2, I was able to escape back ticks using `\`, however now in linguijs v3 the escape brackets are being escaped and thus breaking the translations when it tries to go and find the doubly escaped `\\`.

**To Reproduce**
```js
i18n._(t`Passing \`${argSet}\` in \`${fieldSet}\` is not supported.`)
```
Produces:
```
msgid "Passing \\`${argSet}\\` in \\`${fieldSet}\\` is not supported. "
msgstr "Passing \\`${argSet}\\` in \\`${fieldSet}\\` is not supported. "
```

**Expected behavior**
Previously in v2 running extract on the original string would produce
```
msgid "Passing `${argSet}` in `${fieldSet}` is not supported. "
msgstr "Passing `${argSet}` in `${fieldSet}` is not supported. "
```

- jsLingui version ``@lingui/cli@3.2.3``
- Babel version ``@babel/core@7.12.9``
- Your Babel config: Marcos Plugin, and ``@babel/preset-env, targets: node``

semoal:
Will close #897

### DIFF
--- a/packages/macro/src/macroJs.ts
+++ b/packages/macro/src/macroJs.ts
@@ -8,6 +8,7 @@ import { COMMENT, ID, MESSAGE, EXTRACT_MARK } from "./constants"
 
 const keepSpaceRe = /(?:\\(?:\r\n|\r|\n))+\s+/g
 const keepNewLineRe = /(?:\r\n|\r|\n)+\s+/g
+const removeExtraScapedLiterals = /(?:\\(.))/
 
 function normalizeWhitespace(text) {
   return text.replace(keepSpaceRe, " ").replace(keepNewLineRe, "\n").trim()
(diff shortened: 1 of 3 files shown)


tricoder42 (approved):
The PR as it is looks good, but see my comment in related issue. If you feel the regexp is good enough solution, feel free to merge it 👍

semoal:
Looks better this solution than cooked value, probably the cooked value will break other things that we don't expect, so we can go with this solution and I'll write in some "tech debt" file this issue to investigate more about

PR #903, opened 2020-12-06
fix: accept pseudolocalization in SelectOrdinal

MartinCerny-awin:
It correctly pseudolocalize SelectOrdinal. I have also tried to fix Select
```
<Select value={gender} male="His book" female="Her book" other="Their book" />
```

After extraction the message is:

```
{gender, select, other {Their book}}
```

It looks to me like a bug. I would expect `male` and `female` to be in the message as well.

### DIFF
--- a/packages/cli/src/api/pseudoLocalize.test.ts
+++ b/packages/cli/src/api/pseudoLocalize.test.ts
@@ -68,6 +68,16 @@ describe("PseudoLocalization", () => {
     })
   })
 
+  it("SelectOrdinal", () => {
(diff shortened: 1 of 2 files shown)


semoal:
Hmm has sense to me after the clarification of Tom in your last issues! Thanks

semoal (approved):

MartinCerny-awin:
Should I open bug for Select?

semoal:
Na, not required, we already spoke about this in other issues.

PR #906, opened 2020-12-07
feat: config accepts compilerBabelOptions

### LINKED ISSUE #902: Bug with utf chars
**Describe the bug**
My compiled messages files contains encoded UTF characters instead of simple utf

image

**To Reproduce**
Compile `po` file with Cyrillic characters

**Expected behavior**
Simple not encoded utf chars

- jsLingui version
```
@lingui/cli": "3.2.3",
"@lingui/macro": "3.2.3",
"@lingui/react": "3.2.3",
```
- Babel version ``@babel/core@7.12.9``

semoal:
Will close #902 

We won't break babel 6 users, but babel 7 could use this to modify that ASCII configuration :)

### DIFF
--- a/docs/ref/conf.rst
+++ b/docs/ref/conf.rst
@@ -20,6 +20,7 @@ Default config:
     }],
     "compileNamespace": "cjs",
     "extractBabelOptions": {},
+    "compilerBabelOptions": {},
     "fallbackLocales": {},
     "format": "po",
     "locales": [],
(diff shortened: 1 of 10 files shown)


tricoder42 (approved):
Looks great and it'll definitely save lot of bundle size. We just officially no longer support Babel 6 since v3.

Anyway, let's keep it for other options as well 👍

semoal:
@tricoder42 this is ready, this fixes also the catalogsMergePath warning, was just that optional `?` type.

tricoder42 (approved):
Looks good to me. Let's ship it 🚢

Issue #921, opened 2020-12-29
i18n._ returns any in typescript


gitschwifty:
**Describe the bug**
The type definitions for Lingui @ 3.3.0 have i18n._() returning an any value, rather than a string.

**To Reproduce**
Caught with [@typescript-eslint/no-unsafe-assignment](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-unsafe-assignment.md) turned on. Any element expecting a string throws a lint error if passed the return value from i18n._(t`Some string`)

```
import { withI18n } from "@lingui/react"
import { t } from '@lingui/macro';

export default function App() {
   return <img alt={i18n._(t`An image`)} />
}
```

**Expected behavior**
For i18n._ to return a string, matching up with the external type definitions for 2.x. If it's supposed to return something other than a string sometimes, a union type rather than any.

**Additional context**
The type definition files, @lingui/core/cjs/i18n.d.ts
```
export declare class I18n extends EventEmitter<Events> {
...
    _(id: MessageDescriptor | string, values?: Object | undefined, { message, formats }?: MessageOptions | undefined): any;
...
}
```

@lingui/core/cjs/context.d.ts (after investigating and determining that i18n._ returns the result from interpolate in context)
```
export declare function interpolate(translation: CompiledMessage, locale: string, locales: Locales, localeData: Object): (values: Object, formats?: Object) => any;
```

Here's a patch I made using patch-package to keep the string return behavior for now as well.
```
diff --git a/node_modules/@lingui/core/cjs/i18n.d.ts b/node_modules/@lingui/core/cjs/i18n.d.ts
index 06a4bfe..bd83e16 100644
--- a/node_modules/@lingui/core/cjs/i18n.d.ts
+++ b/node_modules/@lingui/core/cjs/i18n.d.ts
@@ -46,7 +46,7 @@ export declare class I18n extends EventEmitter<Events> {
     load(allMessages: AllMessages): void;
     load(locale: Locale, messages: Messages): void;
     activate(locale: Locale, locales?: Locales): void;
-    _(id: MessageDescriptor | string, values?: Object | undefined, { message, formats }?: MessageOptions | undefined): any;
+    _(id: MessageDescriptor | string, values?: Object | undefined, { message, formats }?: MessageOptions | undefined): string;
     date(value: string | Date, format?: Intl.DateTimeFormatOptions): string;
     number(value: number, format?: Intl.NumberFormatOptions): string;
 }
[...]

PR #922, opened 2020-12-29
fix: added return types to functions returned by interpolate and context in core/cjs/context.ts

gitschwifty:
Should fix issue #921 i18n._ now shows a string return type according to typescript tooltips in vscode


semoal (approved):

semoal:
in v3, it's recommeded to just use t`` macro, you don't need i18n._ for practically anything 👍🏻  anyways returning string i think it's better than any

semoal:
Released in 3.4.0 🥳

PR #931, opened 2021-01-06
docs: update core javascript docs for setupI18n

teebszet:
option catalogs was replaced with option messages
also clearer example to use compiled messages rather than the inconsistent use of catalogs/messages


tricoder42 (approved):
Good point, thank you!

PR #932, opened 2021-01-07
fix: Use LINGUI_CONFIG env as fallback for extract

FredyC:
Tiny fix where `macro` package is using the `LINGUI_CONFIG` env while extract command isn't. In the case of different config locations, this is causing problems. I know there is `--config` for extract, but that means having a path specified in two different ways.

Only thing I am not sure of if the env variable should take precedence over CLI arg. I chose not, but let me know.

### DIFF
--- a/packages/cli/src/lingui-extract.ts
+++ b/packages/cli/src/lingui-extract.ts
@@ -83,7 +83,9 @@ if (require.main === module) {
     .option("--format <format>", "Format of message catalogs")
     .parse(process.argv)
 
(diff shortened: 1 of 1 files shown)


semoal:
AWS says this precedence:

> If you specify an option by using a parameter on the CLI command line, it overrides any value from either the corresponding environment variable or a profile in the configuration file.

We should go with that =)

semoal (approved):

semoal:
Released in 3.4.0 🥳

PR #938, opened 2021-01-13
fix: compilerBabelOptions jest-validate

### LINKED ISSUE #935: Warnings for compilerBabelOptions other than empty.
**Describe the bug**
When using anything other than `compilerBabelOptions: {}` in my lingui configuration, I am issued a warning.

**To Reproduce**

create a lingui configuration file with

```json
{
    "compilerBabelOptions": {
      "minified": false
    },
    "locales": ["en"]
}
```

and run `npm run extract`, to be served with:

```none
● Validation Warning:

  Unknown option "compilerBabelOptions.minified" with value false was found.

Documentation: https://lingui.js.org/ref/conf.html
● Validation Warning:

  Unknown option "compilerBabelOptions.minified" with value false was found.

Documentation: https://lingui.js.org/ref/conf.html
```

`npm run compile` produces the same message, albeit just once.

**Expected behavior**

The configuration is correct, and should not result in a warning.

**Additional context**

- jsLingui version ``lingui --version``: 3.3.0

I suspect the `defaultConfig` definition in `@lingui/conf` needs updating, it currently sets `compilerBabelOptions` to an empty object. If I set it to the default from https://lingui.js.org/ref/conf.html#compilerbabeloptions, the warnings are gone.

semoal:
Will close #935 

unfortunately jest-validate doesn't work if in the example config we set an empty object, we must strict declare the values we accept. I'll setup the default values that we're using and probably the only ones that the final user should be able to change.. Dunno

### DIFF
--- a/packages/cli/src/api/compile.ts
+++ b/packages/cli/src/api/compile.ts
@@ -1,5 +1,5 @@
 import * as t from "@babel/types"
-import generate from "@babel/generator"
+import generate, { GeneratorOptions } from "@babel/generator"
 import { parse } from "messageformat-parser"
 import * as R from "ramda"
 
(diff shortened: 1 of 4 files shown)


semoal:
Released in 3.4.0 🥳

PR #941, opened 2021-01-13
Support for TS based Lingui configuration

### LINKED ISSUE #940: Enable TypeScript based config
I want to extract configured locales and default locales to a separate file (`i18n.config.ts`) so it can be consumed by other parts of the app without duplicating configuration with full type support. But such a file cannot be imported inside the `lingui.config.js` file.

Furthermore, Typescript based config allows using ESM easily to have everything nicely typed.

The `cosmiconfig` allows using custom loaders and to specify custom search paths. Specifically, there is _official_ Typescript loader where you can see how the configuration looks like. It's pretty straightforward.

https://github.com/EndemolShineGroup/cosmiconfig-typescript-loader

An alternative is to keep using untyped `lingui.config.js` and suffer from it :)

**Additional context**
There is a slight issue in the linked loader plugin which makes it work only with async code. Since the Lingui is using `cosmiconfigSync()` it doesn't work together. I already created [the PR](https://github.com/EndemolShineGroup/cosmiconfig-typescript-loader/pull/132) to address that, but the loader seems unmaintained.

However, if you look at [its source code](https://github.com/EndemolShineGroup/cosmiconfig-typescript-loader/blob/develop/src/index.ts), it's like 15 LoC so I wouldn't hesitate much to just copy & paste it inside Lingui. Given it's for the CLI, it doesn't matter that much.

**I am willing to make a PR if you would accept this approach.**

FredyC:
Fixes #940 

In the end, I've included `@endemolshinegroup/cosmiconfig-typescript-loader` as a dependency to consume the Error class from it so it doesn't need to be copied too. In case the https://github.com/EndemolShineGroup/cosmiconfig-typescript-loader/pull/132 gets merged & published, the code here can be simplified.

### DIFF
--- a/packages/conf/package.json
+++ b/packages/conf/package.json
@@ -19,6 +19,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.11.2",
+    "@endemolshinegroup/cosmiconfig-typescript-loader": "^3.0.2",
     "chalk": "^4.1.0",
     "cosmiconfig": "^7.0.0",
     "jest-validate": "^26.5.2"
--- a/packages/conf/src/index.ts
+++ b/packages/conf/src/index.ts
@@ -1,8 +1,9 @@
 import path from "path"
 import fs from "fs"
 import chalk from "chalk"
-import { cosmiconfigSync } from "cosmiconfig"
+import { Loader, cosmiconfigSync } from "cosmiconfig"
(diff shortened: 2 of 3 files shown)


semoal:
Great Daniel, thanks for your contribution, i'm polishing some pending issues and we'll release a new version with this feature introduced :)

semoal:
Released in 3.4.0 🥳

PR #947, opened 2021-01-18
feat: Introduced @lingui/snowpack-plugin

semoal:
Introduced @lingui/snowpack-loader. Let's snowpack users compile on the fly his catalogs (the same as @lingui/loader)

Alright is almost ready!

Kapture 2021-01-18 at 14 49 38

- [x] Pending docs


packages/snowpack-loader/src/index.ts
```diff
@@ -0,0 +1,65 @@
...
+  filePath: string
+}
+function extractLinguiMessages(_?, linguiConfig: LinguiConfigOpts = {}) {
+  const strict = process.env.NODE_ENV !== 'production'
+  const config = getConfig(linguiConfig)
+
+  return {
+    name: '@lingui/snowpack-loader',
+    resolve: {
+      input: ['.po'],
```
semoal, line 20: Probably here we should accept also .json files as webpack loader does.

amankkg: Sure

amankkg:
Regarding package name. Snowpack and its community are using `snowpack-plugin-xyz` or `@xyz/snowpack-plugin` naming template.  
Maybe we can also align to it? To me, it seems like `xyz-loader` is something specific to webpack

docs/index.rst
```diff
@@ -113,6 +113,7 @@ Documentation contents
    Configuration <ref/conf>
    Locale Detection <ref/locale-detector>
    Webpack Loader <ref/loader>
+   Snowpack Loader <ref/snowpack-plugin>
```
amankkg, line 116:
```suggestion
   Snowpack Plugin <ref/snowpack-plugin>
```

docs/ref/snowpack-plugin.rst
```diff
@@ -0,0 +1,46 @@
+***********************************************
+API Reference - Snowpack Loader (@lingui/snowpack-plugin)
```
amankkg, line 2:
```suggestion
API Reference - Snowpack Plugin (@lingui/snowpack-plugin)
```

packages/snowpack-plugin/src/index.ts
```diff
@@ -0,0 +1,65 @@
...
+
+type LinguiConfigOpts = {
+  cwd?: string;
+  configPath?: string;
+  skipValidation?: boolean;
+}
+type SnowpackLoadOpts = {
+  filePath: string
+}
+function extractLinguiMessages(_?, linguiConfig: LinguiConfigOpts = {}) {
```
amankkg, line 13:
```suggestion
function extractLinguiMessages(snowpackConfig?, linguiConfig: LinguiConfigOpts = {}) {
```
this would be more readable unless there are some linter warnings for unused variables

amankkg (approved):
LGTM

Just left some suggestions

packages/snowpack-plugin/src/index.ts
```diff
@@ -0,0 +1,65 @@
...
+  const strict = process.env.NODE_ENV !== 'production'
+  const config = getConfig(linguiConfig)
```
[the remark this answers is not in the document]
semoal: For me has sense outside load because if config is wrong Will fail with all files. If we use load we get the config on each file (slower and could be desynchronization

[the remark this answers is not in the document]
amankkg: OK, I was thinking the same 👌

PR #956, opened 2021-01-23
docs(setup-react): correct links to react and macro pages

SleepWalker:
React and macro links was mixed up


semoal:
uh, thanks mate!

Issue #919, opened 2020-12-23
<Select /> choices are not extracted correctly


chuyik:
**Describe the bug**
Some choices on `<Select />` macro are missing after `lingui extract` called.

**To Reproduce**
```jsx
import { Select } from "@lingui/macro"

export default function App() {
  const gender = "female";
  return (
    <Select
      value={gender}
      male="His book"
      female="Her book"
      other="Their books"
    />
  );
}
```

After extraction, `male` and `female` missing:

```json
{
  "{gender, select, other {Their books}}": {
    "translation": "{gender, select, other {Their books}}",
    "extractedComments": []
  }
}
```

**Expected behavior**
```json
{
  "{gender, select, male {His book} female {Her book} other {Their books}}": {
    "translation": "{gender, select, male {His book} female {Her book} other {Their books}}",
    "extractedComments": []
  }
}
```

**Additional context**

- jsLingui version `3.3.0`
- Babel version
```bash
├─ @babel/core@7.12.10
├─ @jest/transform@26.6.2
│  └─ @babel/core@7.9.0
├─ babel-preset-react-app@10.0.0
│  └─ @babel/core@7.12.3
├─ jest-config@26.6.3
│  └─ @babel/core@7.9.0
└─ react-scripts@4.0.1
   └─ @babel/core@7.12.3
```
- Framework: Create React App

chuyik:
However, if I use the `select` macro(not `Select`), it outputs as expected.

Dig into source code, I found that `Select` will detect if option matches plural rules, while `select` does not.

`select`: https://github.com/lingui/js-lingui/blob/23b06b5dfb/packages/macro/src/macroJs.ts#L318-L329
`Select`: https://github.com/lingui/js-lingui/blob/23b06b5dfb/packages/macro/src/macroJsx.ts#L296-L304

semoal:
Released 3.5.0

PR #958, opened 2021-01-26
fix: <Select /> macro not extracted properly

### LINKED ISSUE #959: Macro `Plural` are not extracted correctly with `component` prop
**Describe the bug**
If marco `<Plural>` is passed with `component` prop, this prop will also be extracted to message catalog.

**To Reproduce**

```tsx
import { Plural } from "@lingui/macro"

  const count = 1
  const Div = ({ children }) => <div>{children}</div>

  return (
    <Plural
      id="message"
      value={count}
      _0="No messages"
      one="# message"
      other="# messages"
      component={Div}
    />
  )
}
```

This code will extracts incorrect message like follows:

```json
{
  "message": {
    "translation": "{count, plural, =0 {No messages} one {# message} other {# messages} component {{Div}}}",
    "message": "{count, plural, =0 {No messages} one {# message} other {# messages} component {{Div}}}",
    "extractedComments": []
  }
}
```

Please notice that ``component {{Div}}`` is appended to `message.translation` and `message.message`.

**Expected behavior**
The `component` prop should be ignored.

**Additional context**

- jsLingui version `3.4.0`
- Babel version `@babel/core@7.12.10`

semoal:
#919 Will close this issue

Also will close #959

### DIFF
--- a/packages/cli/src/api/pseudoLocalize.test.ts
+++ b/packages/cli/src/api/pseudoLocalize.test.ts
@@ -43,7 +43,7 @@ describe("PseudoLocalization", () => {
           "{count, plural, offset:1 zero {There are no messages} other {There are # messages in your inbox}}"
         )
       ).toEqual(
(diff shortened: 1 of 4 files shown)


semoal:
@MartinCerny-awin

PR #963, opened 2021-01-27
docs: Fix prop naming in react pattern example

vladshcherbin:


semoal:
Goood! tthanks for your contribution

PR #964, opened 2021-01-27
No message when translations are missing

HamAndRock:
Hello, 
I stumble upon this issue quite frequently, sometimes someone at team accidentally deletes a string in one of translations files and it throws a generic error. 

Error:
image

I saw your if check few lines before the crashing point, hovewer that if won't catch it so I moved it to correct scope. 

My fix results into this:

image


landsman:
I thought that this was resolved here: https://github.com/lingui/js-lingui/pull/545 but it's not I see.

semoal (approved):

PR #971, opened 2021-02-08
fix(@lingui/loader): accept webpack 5.x as a peer dependency

jesec:
npm 7 mandates matching peer dependencies. See RFC 25.

Fixes "Could not resolve dependency: npm ERR! peer webpack@"^4.0.0" from @lingui/loader"

@lingui/loader works perfectly fine with Webpack 5.x and a medium-size project.


semoal:
Great mate, we'll release it as soon as possible

PR #972, opened 2021-02-08
fix: runtimeConfigModule config validation

arcanis:
It seems like the validation got stricter, and `runtimeConfigModule` alternative syntax (#895) wasn't covered.


semoal:
Thanks you so much Maël! Much appreciated

PR #980, opened 2021-02-21
docs: Update react patterns

orangain:
[React common patterns](https://lingui.js.org/tutorials/react-patterns.html) is very helpful but its content seems to be a little old. This PR update the content to use `useLingui` hook instead of `I18n` render-prop component.

This PR also add link from react tutorial to common i18n patterns as a further reading.

### DIFF
--- a/docs/tutorials/react-patterns.rst
+++ b/docs/tutorials/react-patterns.rst
@@ -169,10 +169,11 @@ passing it to :jsxmacro:`Trans` as its ``id`` prop:
    }
 
 Or to render the message descriptor as a string-only translation, just pass it to
-the :js:meth:`I18n._` method as usual:
+the :js:meth:`I18n._` method:
 
 .. code-block:: jsx
 
+   import { i18n } from '@lingui/core'
    import { t } from "@lingui/macro"
 
    const favoriteColors = [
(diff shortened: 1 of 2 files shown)


semoal:
On v3 i18n._(t``) isn't required anymore you can use directly t``

orangain:
@semoal Thank you for the comment! I've fixed several examples to use `t` macro directly. I've also fixed usage of `t` macro with custom ID.

From my understanding, `i18n._` call is still required in the example of lazy translation https://github.com/lingui/js-lingui/pull/980/commits/0eec517ba7697890c1d739b775807ee764d0befb even in v3. Is it true?

semoal:
> @semoal Thank you for the comment! I've fixed several examples to use `t` macro directly. I've also fixed usage of `t` macro with custom ID.
> 
> From my understanding, `i18n._` call is still required in the example of lazy translation [0eec517](https://github.com/lingui/js-lingui/commit/0eec517ba7697890c1d739b775807ee764d0befb) even in v3. Is it true?

Tecnically for lazy translation the best way of doing it, is `defineMessage` but you can use also the i18n._(), so keep it like that. We have tons of work with the documentation :( so your help is really helpful

PR #995, opened 2021-03-04
docs: explain locales config in the Setup tutorial

piettes:
A new user has no idea how the configuration file should be, so writing "After fixing configuration" in the tutorial is not helping anything.
And actually just putting the locales in the config will make lingui scan the /node_modules/ directory, which will throw a lot of errors. 
So a minimal working .linguirc/config example would be great.


semoal:
LGTM, we have tons of works pending with documentation, so feel free to collaborate when you want mate =)
Thanks for the contribution

PR #997, opened 2021-03-07
fix: exports attribute on package.json, expo compat

renchap:
React Native's Metro bundler expects all imports to be declared in the `exports` attribute when there is one defined. 
Lingui imports it (to get the version I guess?) and Metro refuses it as it was not defined as a valid export.
See https://github.com/eemeli/make-plural/issues/15 for a similar issue on the `make-plural` package

The `main` field for `@babel/macro` was also incorrect as there is no `index.ts` in this project.

These 2 changes allowed me to successfully use Lingui on a React Native project.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -29,8 +29,11 @@
     "node": ">=10.0.0"
   },
   "exports": {
-    "require": "./index.js",
-    "import": "./esm/index.js"
+    ".": {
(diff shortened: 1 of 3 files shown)


semoal:
Ah super Renaud! thanks for the contribution :)

semoal:
Released 3.7.1 with this fix =)

PR #1002, opened 2021-03-11
chore: include conf build folder in npm

### LINKED ISSUE #1001: @lingui/conf on npm has no main file
**Describe the bug**
image
these are all the files that get installed from npm, so the main
```
  "main": "index.js",
```
in package.json is pointing to a file that does not exist.

**To Reproduce**

install from npm

**Expected behavior**

main file is present

**Additional context**
This is causing me issues getting lingui work with vite

capaj:
this fixes #1001

### DIFF
--- a/packages/conf/.npmignore
+++ b/packages/conf/.npmignore
@@ -1,0 +1,3 @@
+build/package.json
+build/LICENSE
+build/README.md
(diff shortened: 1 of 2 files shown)


capaj:
oh weird, it does seem to work fine now. This was probably never needed. I have no idea why npm installed `conf` package without the index file for me. It is in the 3.7.1 tarball so it might have been some NPM glitch.

PR #1024, opened 2021-04-02
[Feature] Locked Catalogs

nathanforce:
This PR implements a first pass at locking a catalog to opt out of future runs of the `extract` command as discussed in #1015.  Locked catalogs have no impact on the `compile` command as we still want to consider locked catalogs as valid and loadable, just frozen in time. Releasing the lock (changing to `"active"` or deleting the property and running `extract` would update/regenerate the catalog). 

### Open Question

1. How should I test this? The CLI`extract` command is currently untested.
2. There is some confusing overlap between `status: "locked"` and `include: []`.  Maybe it makes more sense to keep this use case of "locking" a catalog in user-land by decoupling `include/exclude` from `getCatalogs` and allowing `compile` to compile catalogs regardless of their `include/exclude` patterns. What do you think? It seems like `include/exclude` is specific to `extract`?

### DIFF
--- a/packages/cli/src/api/catalog.ts
+++ b/packages/cli/src/api/catalog.ts
@@ -7,7 +7,12 @@ import glob from "glob"
 import micromatch from "micromatch"
 import normalize from "normalize-path"
 
-import { LinguiConfig, OrderBy, FallbackLocales } from "@lingui/conf"
+import {
+  LinguiConfig,
+  OrderBy,
+  FallbackLocales,
+  CatalogStatus,
+} from "@lingui/conf"
 
 import getFormat, { CatalogFormatter } from "./formats"
 import extract from "./extractors"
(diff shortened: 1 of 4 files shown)


semoal:
Hello @nathanforce , really appreciated your pull request.
About your questions:
1. I have pending to test all the cli commands, so don't worry about testing it on the suite. If you can give it a try to this functionality following this tutorial: https://github.com/lingui/js-lingui/blob/main/CONTRIBUTING.md#using-development-version-in-your-project, basically runs a verdaccio when you deploy a version of lingui with your changes and you can test it in a real project. If you do this and you ensure that everything is fine, we'll be okay since this feature won't broke anything to another users.
2. Right now when compiling doesnt not check the extract/include keys? In my humble opinion should act in the same way compile & extract when checking the config keys (in the same way for locked status). If a user wants to compile something especial we should allow it by cli args.

semoal (approved):

nathanforce:
@semoal I think I am going to close this PR.  After playing around with the changes locally I don't know if a frozen catalog is what we want.  Freezing works fine to opt out of future translations but does not work well with Lingui's approach to message ID generation. Changing the content of an unlocked (active) translation would change the ID and thus break the frozen catalog.

Issue #957, opened 2021-01-26
Macro `t` with template string id could not be extracted


chuyik:
**Describe the bug**
Macro `t` will NOT be extracted to message catalog, if we pass id with a template string like ``t({id: `id`})``.

**To Reproduce**

```jsx
import { t } from '@lingui/macro'

export default function App() {
  return (
    <div>
      <p>
        {/* id with string */}
        {t({
          id: 'today1',
          message: `Today is ${new Date()}`,
        })}
      </p>

      <p>
        {/* id with template literals */}
        {t({
          id: `today2`,
          message: `Today is ${new Date()}`,
        })}
      </p>
    </div>
  )
}

```

Only `today1` is extracted, but `today2` is missing:
```json
{
  "today1": {
    "translation": "Today is {0}",
    "message": "Today is {0}",
    "extractedComments": []
  }
}
```

**Expected behavior**
`today 2` should also be extracted.

**Additional context**

- jsLingui version `3.4.0`
- Babel version `@babel/core@7.12.10`

dmn-bernard:
Yup this is pity I work on a project where backticks are enforced via eslint. Now I have to copy paste `// eslint-disable-line` everytime I use  the `t` macro with an id 😓

semoal:
Just for curisosity, do you enforce the backticks always? The default rule by eslint is to use singlequote or doublequote, and if you're using + "" + "", replace it with backticks. I don't get the utility of forcing backticks to every string.

Anyways, this behaviour could be a good addition that's why triage accepted, but couldn't find time to work on it since isn't urgent at all.

doc-sultan:
I think developers should be free to use any eslint rules on their own choice. 
I would like to try to fix that issue, but before start doing it I would like to know, is this fix not against the philosophy of lingui maintainers or maybe some other reasons were to implement it the way it is?

semoal:
> I think developers should be free to use any eslint rules on their own choice.
> I would like to try to fix that issue, but before start doing it I would like to know, is this fix not against the philosophy of lingui maintainers or maybe some other reasons were to implement it the way it is?

The "common" behaviour is to define id like a simple string with double/single quotes, that's why the standard implementation works. When Tom refactored the v2 codebase to v3, probably didn't though about this use case, but is not against the philosophy of Lingui, it's just okay, (no free time on my side, and still less on Tom side)

Feel free to take it and ping me if you need some guidance =)

semoal:
[...]

PR #1027, opened 2021-04-04
fix: extract works with template string id's

sultan99:
This PR closes issue #957 macro `t` with template string id.

### DIFF
--- a/packages/babel-plugin-extract-messages/src/index.ts
+++ b/packages/babel-plugin-extract-messages/src/index.ts
@@ -237,8 +237,8 @@ export default function ({ types: t }) {
                 .get(`properties.${i}.value`)
                 .buildCodeFrameError("Only strings are supported as comments.")
             }
-
-            props[key.name] = value.value
+            const isIdLiteral = !value.value && key.name === "id" && t.isTemplateLiteral(value)
+            props[key.name] = isIdLiteral ? value?.quasis[0]?.value?.cooked : value.value
           })
 
         collectMessage(path, file, props)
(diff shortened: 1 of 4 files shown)


semoal:
Looks good mate, but could you me explain what does it mean?
`value?.quasis[0]?.value?.cooked`, just for giving context to the pull request

sultan99:
@semoal `?` - It is an [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), so with `?` we don't get ReferenceError error if any value in the object will be undefined. I did it to cover all cases for the test coverage 😅.

semoal:
> @semoal `?` - It is an [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining), so with `?` we don't get ReferenceError error if any value in the object will be undefined. I did it to cover all cases for the test coverage 😅.
> 
> If you are asking about this object `value.quasis[0].value.cooked`, so the value in the template string is stored in that path and I had to get it in order to make it all work.

Yes, i was asking about the object quasis and the cooked value.  Also, another question, shouldn't be like this?:

```js
props[key.name] = (isIdLiteral && value?.quasis[0]?.value?.cooked) || value.value || undefined
```
When is a template literal always value.value it's undefined?
Just being sure that we don't break anything ^^

sultan99:
Well, there is no big difference, but anyway I have optimized it.
Please check it & let's publish it! 🤘

Issue #1022, opened 2021-04-02
Cannot extract using i18n.plural : undefined is not iterable!


kopax-polyconseil:
Hello, we use lingui v2.9.2 and we got issue extracting using `i18n.plural`:

This works:

```tsx
export function _(
  id: MessageDescriptor | string,
  values?: Record<string, unknown>,
  messageOptions?: MessageOptions
): string {
  if (typeof id === 'string') {
    return i18n._(id, values, messageOptions)
  }
  return i18n._(id)
}
```

This fail:

```tsx
export function _(id: MessageDescriptor | PluralProps): string {
  return !Object.prototype.hasOwnProperty.call(id, 'one')
    ? i18n._(id as MessageDescriptor)
    : i18n.plural(id as PluralProps)
}
```

It fails when we run `lingui extract --clean`, with the following error:

```ruby
TypeError: src/libs/i18n.ts: undefined is not iterable!
    at module.exports.require.getIterator (/home/dka/workspace/github.com/pass-culture/pass-culture-app-native/node_modules/babel-runtime/node_modules/core-js/library/modules/core.get-iterator.js:5:42)
    at Transformer.transformChoiceMethod (/home/dka/workspace/github.com/pass-culture/pass-culture-app-native/node_modules/@lingui/babel-plugin-transform-js/transformer.js:199:56)
    at Transformer.transformMethod (/home/dka/workspace/github.com/pass-culture/pass-culture-app-native/node_modules/@lingui/babel-plugin-transform-js/transformer.js:393:21)
    at PluginPass.Transformer.transform (/home/dka/workspace/github.com/pass-culture/pass-culture-app-native/node_modules/@lingui/babel-plugin-transform-js/transformer.js:69:25)
    at newFn (/home/dka/workspace/github.com/pass-culture/pass-culture-app-native/node_modules/babel-traverse/lib/visitors.js:276:21)
    at NodePath._call (/home/dka/workspace/github.com/pass-culture/pass-culture-app-native/node_modules/babel-traverse/lib/path/context.js:76:18)
    at NodePath.call (/home/dka/workspace/github.com/pass-culture/pass-culture-app-native/node_modules/babel-traverse/lib/path/context.js:48:17)
    at NodePath.visit (/home/dka/workspace/github.com/pass-culture/pass-culture-app-native/node_modules/babel-traverse/lib/path/context.js:105:12)
[...]

PR #1035, opened 2021-04-08
fix: unicode chars were extracting with double slashes

semoal:
Re-fixes the unicode error we found migrating the passCulture application to v3.

I went for a walk and start to think about this and saw that we weren't extracting correctly the unicodes.

`\\unicode` is not correct => `\unicode`this is correct 

**probably with this adjustment we fix the wrong behaviour when we built the application to production**

About the string failing in development about not being processed as a string, we'll see tomorrow debuggin the node_modules which is the root cause 

#1022

### DIFF
--- a/packages/macro/src/macroJs.test.ts
+++ b/packages/macro/src/macroJs.test.ts
@@ -87,7 +87,7 @@ describe("js macro", () => {
       expect(tokens).toEqual([
         {
           type: "text",
-          value: 'Message \\u0020',
+          value: 'Message \/u0020',
         },
       ])
     })
--- a/packages/macro/src/macroJs.ts
+++ b/packages/macro/src/macroJs.ts
@@ -361,7 +361,7 @@ export default class MacroJs {
    */
   clearBackslashes(value: string) {
     // it's an unicode char so we should keep them
-    if (value.includes('\\u')) return value
+    if (value.includes('\\u')) return value.replace(removeExtraScapedLiterals, "\/u")
     // if not we replace the extra scaped literals
     return value.replace(removeExtraScapedLiterals, "`")
   }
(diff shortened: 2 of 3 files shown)


kopax-polyconseil:
I've just tested it and it seems NOK, read more here https://github.com/lingui/js-lingui/issues/1022#issuecomment-815957967

semoal:
Absolutely right. Drunk some coffees and took the playground i've built yesterday of React Native and start trying out things.

What i've found:
- 1st the GENIOUS fix i've merged here, it's not ok and the default behaviour is correct.

Cool things I've think I found a fix:

Screenshot 2021-04-09 at 11 34 53
Screenshot 2021-04-09 at 11 34 42

What extracts?:

```ts
/*eslint-disable*/ export const messages = {
  'Autoriser l’envoi d’e‑mails': 'Autoriser l’envoi d’e‑mails',
  EAN: 'EAN',
  'Profite de 300 €': 'Profite de 300 €',
  'Version {0}': ['Version ', ['0']],
};
```

```po
#: App.tsx:41
#: App.tsx:43
msgid "Autoriser l’envoi d’e‑mails"
msgstr "Autoriser l’envoi d’e‑mails"

#: App.tsx:33
#: App.tsx:35
msgid "EAN"
msgstr "EAN"

#: App.tsx:37
#: App.tsx:39
msgid "Profite de 300 €"
msgstr "Profite de 300 €"

#: App.tsx:45
msgid "Version 300 €  {0}"
msgstr "Version 300 €  {0}"
```

In the case you want strings with unicode chars in the compiled and the .po file, you can use this option in the lingui.config.js:

```js
  compilerBabelOptions: {
    jsescOption: {
      minimal: true,
    },
  },

More info about this option here: https://lingui.js.org/ref/conf.html#compilerbabeloptions
```

I think we're close of fixing this, the pending one is to make sure that this fixes are fixed also in the prod build

semoal:
Also i saw that we use EventEmitter class for our internal events inside Lingui but React Native doesn't have EventEmitter, I'm looking already on it

PR #1038, opened 2021-04-09
fix: event emitter refactor (reverted)

semoal:
This reverts commit a7aa6a6cae302c9ff595abe4f60b67e930a11b89.

### DIFF
--- a/packages/core/src/eventEmitter.test.ts
+++ b/packages/core/src/eventEmitter.test.ts
@@ -19,12 +19,12 @@ describe("@lingui/core/eventEmitter", () => {
     const listener = jest.fn()
     const emitter = new EventEmitter()
 
-    emitter.on("test", listener)
+    const unsubscribe = emitter.on("test", listener)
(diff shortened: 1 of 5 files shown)


kopax-polyconseil:
I am receiving a bunch of testeurs and developer complaining about variable interpolation and unicodes chars. 

Can we conclude the PP session? Otherwise, I must downgrade to v2 and won't be able to support further in those investigations.

semoal:
> I am receiving a bunch of testeurs and developer complaining about variable interpolation and unicodes chars.
> 
> Can we conclude the PP session? Otherwise, I must downgrade to v2 and won't be able to support further in those investigations.
> 
> That's just another example which involve dev an testeurs: [pass-culture/pass-culture-app-native#861 (files)](https://github.com/pass-culture/pass-culture-app-native/pull/861/files)

Just did an update in the other thread, i'm going to lunch and we can pp other time and we try to resolve everything.

Issue #1058, opened 2021-05-11
Extracting fails when next/dynamic is used


datagutt:
**Describe the bug**
When using Next.js dynamic imports, Lingui is unable to extract messages from the pages/components using dynamic imports.

**To Reproduce**
Steps to reproduce the behavior, possibly with minimal code sample, e.g:

```
import { Trans } from "@lingui/react"
import dynamic from "next/dynamic"
const TranslatedComponent = dynamic(() => import("../components/TranslatedComponent"))

export default function Index() {
   return (<div><TranslatedComponent /></div>
}
```

When extracting pages/index.js, i get this error:
```Cannot process file app/pages/index.js: Cannot read property 'pagesDir' of undefined```

**Expected behavior**
Messages are extracted from pages using dynamic components.

**Additional context**

This used to work fine in Next.js 10.1, but broke in 10.2. Likely related to this: 
PR: https://github.com/vercel/next.js/pull/24281
Issue: https://github.com/vercel/next.js/issues/24566

- jsLingui version ``3.8.10``
- Babel version ``7.14.0``
- Your Babel config (e.g. `.babelrc`) or framework you use (Create React App, Meteor, etc.)
Next.JS 10.2

semoal:
Actually crashes internally Next but extracts correctly the value to the messages file. (Tried with @canary version of Next and the latest)

Actually the fix from our side is to remove the Typescript extractor located here: https://github.com/lingui/js-lingui/blob/main/packages/cli/src/api/extractors/index.ts#L6, we already discussed this on https://github.com/lingui/js-lingui/discussions/1047 and probably we I can find some free hours I'll work hard on a next version of Lingui will all these bugs fixed.

On my head I want to add a similar API of Jest to add any extractors :)

semoal:
Released 3.9.0 give it a try when you can, but i'm pretty sure that will fixed :)

PR #1065, opened 2021-05-18
feat: configurable and extendable extractors with Lingui config

semoal:
Now with this pull request users can create his own extractors:

```
type ExtractorType = {
  match(filename: string): boolean
  extract(filename: string, targetDir: string, options?: any): void
}
```

### Little breaking change
Before we had two extractors `[babel, typescript]`, the typescript step it's useless since babel support typescript practically out of the box, since this compiling it's done twice first by babel and then by typescript was causing a lot of failures on Typescript projects with bad origin's and strange errors about NODE_ENV.

Related issues: 
https://github.com/lingui/js-lingui/issues/1058
https://github.com/lingui/js-lingui/discussions/1047

Now if your package it's pure Typescript and doesn't contain Babel, you could just use typescript extractor instead of using the babel one.

```js
  extractors: [
    require.resolve("@lingui/cli/api/extractors/typescript")
  ],
```

- [x] Document this feature

### DIFF
--- a/docs/ref/conf.rst
+++ b/docs/ref/conf.rst
@@ -29,6 +29,7 @@ Default config:
     "fallbackLocales": {},
     "format": "po",
     "locales": [],
+    "extractors": ["babel"],
     "orderBy": "messageId",
     "pseudoLocale": "",
     "rootDir": ".",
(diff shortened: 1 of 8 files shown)


semoal:
@nathanforce

PR #1073, opened 2021-05-25
fix: po-gettext format issues with CLI

nakkamarra:
I noticed two issues with the `po-gettext` format option:

1. The `disableSelectWarning` flag is not part of the `CatalogFormatOptions` type, and so it is reported as being invalid by the CLI:

Screen Shot 2021-05-25 at 1 32 13 PM

2. The `po-gettext` format does not properly support the use of `lineNumbers` format option that the standard `po` format option supports.

### DIFF
--- a/packages/cli/src/api/formats/index.ts
+++ b/packages/cli/src/api/formats/index.ts
@@ -17,6 +17,7 @@ const formats: Record<CatalogFormat, CatalogFormatter> = {
 
 type CatalogFormatOptionsInternal = {
   locale: string
+  disableSelectWarning?: boolean
 } & CatalogFormatOptions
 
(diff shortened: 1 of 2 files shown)


nakkamarra:
@semoal hey, can I get a quick review on this when you have the chance? Thanks.

nakkamarra:
@semoal It looks like @iStefo authored the feature, so perhaps he would be able to give a quick review if you are too busy?

semoal:
Hello Nick, sorry for the delay, had been busy lately with work and Rematch's book. 
I'll take a review to everything this weekend and on Tuesday we'll release a version =)

semoal (approved):

PR #1082, opened 2021-06-09
fix: Move property def into `conf` package

nakkamarra:
@semoal I believe I placed this in the wrong file, the validation of the config object actually happens within the `conf` package, so that was a bit of misreading on my part. I tested this on locally and it suppresses the issue I was describing in #1073


semoal:
No worries, I have to adjust some details of the remoteLoader implementation so I'll release a new version at the end of today :)

PR #1088, opened 2021-06-16
fix: 'compile --watch' not watching for correct file based on format

luludan:
Currently the `--watch` option for `compile` only works for `po` format. This is because the file extension matches the format.

For other format options like `minimal` then currently the watcher is looking for files with a `.minimal` extension rather than a `.json` extension.


semoal:
Looks good, nice catch! I'll release it tomorrow probably. Thanks for the contribution mate

PR #1089, opened 2021-06-19
fix: add extract-template for po-gettext format

truong-hua:
The previous implemented of `lingui extract-template` command does not work with po-gettext format because of hard-coded `po`. This change support any further `po-xyz` format in the future.


semoal:
LGTM @truong-hua, really appreciated the contribution. Going to merge it right now, once the ci passes. And I'll release a new version tomorrow, I want to improve some usability with the new tool remote-loader.

Issue #856, opened 2020-11-12
Support `msgctx`


tricoder42:
- [ ] Add `context` prop to message descriptor (used in `defineMessage` and `t` macros)
- [ ] Add `context` prop to `Trans` macro and also `Trans` component (from `@lingui/macro`)
- [ ] Collect `context` in extract plugin. In `cli` package, throw an error when file format is anything else than `po`
- [ ] Refactor `@lingui/core` to support lookup by context. Right now you get a translation using a simple object lookup `catalog[messageId] === translation`
- [ ] Add documentation

Related discussion: https://github.com/lingui/js-lingui/issues/845#issuecomment-725728398
Related issues: #598 #797 #298

nakkamarra:
@tricoder42 Awesome, let me know if you'd like some help, I can maybe spend some time this weekend to make a PR.

semoal:
Feel free to take it a provide a pull request Nick, I assign this to you, if you dont mind

tricoder42:
@nakkamarra Definitely go ahead. I've created the issue only to outline what's required to do

Bertg:
I'll probably have some time to work on this tomorrow (Wed 17th Central European timezone).
Is someone wants to pair up, let me know.

nakkamarra:
@Bertg Yeah that sounds good. I've gotten started on this and am happy to chat about where I've gotten. However, I think there are some implications for the changes, particularly around how we will handle `missing` when message context is part of the equation, that I have not yet figured out.

Bertg:
I worked a bit on this tonight and wanted to collect some feedback before finalising and creating a PR. (Especially @tricoder42)

https://github.com/lingui/js-lingui/compare/main...Bertg:msgctxt_support

What works: 

- Extract messages and remain uniq based on context
- Lookup based on context

When extracting messages & during lookup at runtime an Object (hash) is used. Before these changes the key was the ID. A simple approach would have been to change the key to an array of strings, however this leads to several issues. The extract process wants to serialise its internal structure to JSON which does not support arrays as keys, and I assume lookup of object with arrays as keys is not that performant. So I chose to replace the ID as key with a this: `[id, context].reject(s => !s).join('\u001F')`. This yield the same key for existing transactions and creates a uniq key for messages with a context. 

[...]

PR #1094, opened 2021-06-30
feat: msgctxt support 

nakkamarra:
As described as part of: https://github.com/lingui/js-lingui/issues/856

My decision here was to use `context` / `msgctxt` as a way to "sub-key" a given key, which I feel lines up more with the [original gettext implmentation/intention of context](https://www.gnu.org/software/gettext/manual/html_node/Contexts.html). So if we had a catalog like so:

```json
{
    "keyA": "translationA",
    "keyB": "translationB",
}
```

the addition of context could be modeled like so:

```json
{
    "keyA": "translationA",
    "keyB": "translationB",
    "contextA": {
         "keyA": "translationAA",
         "keyB": "translationAB"
     },
     "contextB": {
         "keyA": "translationBA"
     }
}
```

where a given context string acts as the key of another sub-catalog of translations. This allows one to add multiple translations for a given ID / string by using different contexts. To pull out translations, one would reference `this.messages[context][id]` instead of `this.messages[id]` in the case that context was defined for a given ID.

On top of that,  I think this is actually pretty trivial to support for many of lingui's current formats (`po`, `po-gettext`, `minimal`, and `raw`), since we're just creating sub-objects and serializing / deserializing them. As far as CSV is concerned, I think there'd need to be a bit of extra work, but it could be possible to apply something similar to what we're doing with JSON (imagine instead of each row being `key, translation` you have `context, key, translation` and work with that)

There's the caveat that this does not abstract `msgid`s and solve the shrinking of IDs in catalogs, one of the upsides to @Bertg's approach to this problem.

I'd say this is about 90% of the way there, but I'd love to get some feedback and thoughts around this approach, particularly around this chunk which could be refactored to be much cleaner: https://github.com/lingui/js-lingui/compare/main...nakkamarra:msgctx-support?expand=1#diff-9317c05f72203559bc3788fb80d76d19f38b457e50c35c430463d964574eec90R27-R80

@semoal @tricoder42 @Bertg


nakkamarra:
Hey @semoal think you could take a look when you have the chance?

semoal:
After so much time, finally this is released guys.. Sorry for the delay I've been lately so busy ..
3.13.0

PR #1101, opened 2021-07-22
feat: Add debounce option to compile and extract CLI

jeroenvisser101:
This adds support for debounce, avoiding multiple subsequent file saves from also resulting in subsequent extraction/compilation.

I noticed that the files, according to the `.prettierrc` in the repo, were formatted even though they weren't touched to add this feature. If this is undesirable, I can revert those changes. Let me know.

Thanks!

### DIFF
--- a/docs/ref/cli.rst
+++ b/docs/ref/cli.rst
@@ -44,7 +44,7 @@ Commands
 ``extract``
 -----------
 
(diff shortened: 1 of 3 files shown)


semoal:
Thanks for the contribution, really appreciated!

PR #1108, opened 2021-08-05
fix: no files being found when catalogs.include has more than one entry

gwilymhumphreys:
When catalogs.include contains an array of > 1 globs they are joined together into one expression. This expression is missing an end curly brace } and this fix adds it.


semoal:
Thanks for the contribution, really appreciated!

PR #1119, opened 2021-08-18
fix: wrong detect whether a translation is plural

truong-hua:
The previous logic is wrong because a language like Vietnamese, Indonesian, Malaysian... does not support plural form, so gettext tools does not allow these kind of to have multiple msgstr. Therefore detecting a plural translation through number of msgstr is not true. As consequence, that causes plural macro can not be replaced with proper translation.

Ex:
```
#. js-lingui:icu=%7BmonthNumber%2C+plural%2C+one+%7B%2F+%23+th%C3%A1ng%7D+other+%7B%2F+%23+th%C3%A1ng%7D%7D&pluralize_on=monthNumber
msgid "/ # tháng"
msgid_plural "/ # tháng"
msgstr[0] "/ # bulan"
```


semoal:
LGTM, thanks mate

PR #1123, opened 2021-08-30
Fix: handle the case fallbackLocales is false

Shinyaigeek:
Fix: 1121

lingui fails to compile when `fallbackLocales: false` is specified in lingui configuration and `msgStr` is missing.
I fixed this issue.


semoal:
LGTM, thanks for contributing mate. I'll keep merging triage: accepted bugs/features along the week and I'll release a new version next Tuesday.

PR #1138, opened 2021-09-23
docs: 2 new pages: "Sync & Collaboration Tools" and "Translation.io"

MichaelHoste:
I finally made some documentation for the "Sync & Collaboration Tools" (see #1107).

There is a small introduction that explains how to translate an app in a regular workflow (`.po` files and text editor or PoEdit), and that also explains how to scale that up with a sync & collaboration tool if the application becomes too big (source text that doesn't stop evolving and lots of target languages). The diagrams are mine so there are no copyright issues.

I also explained the process to integrate new tools into Lingui if other services want to add theirs in the future.

There is a second page that describes Translation.io (features, installation, usage). This page contains many animated gifs that are hosted directly on Translation.io. I didn't want to push them on this repository because they are quite big. I hope it's ok with you.

This PR is obviously not completely selfless and we hope that there will be some positive synergies between Lingui and [Translation.io](https://translation.io/lingui), but we tried to create a structure so that other competing services could easily be added in the future, providing more value for Lingui users.

------

Many small Sphinx warnings were also fixed in this PR.


semoal:
Looks perfect Michael, really happy to introduce this merge request to main :) Hope you get more users since we collaborate together =)

Issue #901, opened 2020-12-06
Documentation/clarifications for server side rendering


max-mykhailenko:
Can't understand the correct way to use Lingui with SSR after updating to v3.

First of all I know that a lot of requests can be received in the same time, but you remove `setup18n` and now i18n is a singleton for all simultaneously requests. Is it correct (possible some macros magic fix this, but from docs can't understand)?

Next question is keeping `import { i18n } from "@lingui/core"` but not using it. Do we really need it?
<img width="738" alt="image" src="https://user-images.githubusercontent.com/1213198/101279693-a1a91180-37cc-11eb-98b1-b01a12cce51f.png">

semoal:
I start from the end. You're right you don't need that import internally will use that instance, so just use macros. (If you want to create a pr to fix the docs are welcome =) )

About the first question, `setupI18n` is not removed, you can still use it if you need some extra config, that returns a i18n instance that can be used anywhere. But in the 99% cases you can use the default `{ i18n } from '@lingui/core'`

max-mykhailenko:
@semoal thank you for your fast response. I'll create fix for this and for next section
<img width="717" alt="image" src="https://user-images.githubusercontent.com/1213198/101288144-9b7f5900-37fd-11eb-8ca4-eb779cbfcc18.png">
As I saw in types we should pass messages, but not catalogs
<img width="636" alt="image" src="https://user-images.githubusercontent.com/1213198/101288155-acc86580-37fd-11eb-9e81-fd009bee0c8e.png">

semoal:
Perfect, thanks you very much!

khmm12:
Hey @semoal 👋

> About the first question, setupI18n is not removed, you can still use it if you need some extra config, that returns a i18n instance that can be used anywhere. But in the 99% cases you can use the default { i18n } from '@lingui/core'.

How does it work on concurrent environments? 

Before it was possible to create as many instances as you want and use them isolated from each other. Now it became impossible without CLS on server-side. 

Yeah it's possible to get rid of macroses and live without them, but loose a syntax sugar and extracting of messages via CLI.

seeden:
@semoal can you answer https://github.com/lingui/js-lingui/issues/901#issuecomment-773242945 I would like to know this answer too and I have no idea how to fix it without my own "t``" macro

max-mykhailenko:
[...]

Issue #578, opened 2019-10-16
[v3] `t` macro uses global i18n instance


janicduplessis:
**Describe the bug**
I'm using this library on the server and create a new i18n instance for each request so they don't conflict with concurrent requests. The issue is that when using the `t` macro it seems to do more than what I expect. Instead of converting the string to a message object like mentioned in the doc it gets transformed to a call to `i18n._(message)` which returns a translated string.

Not sure if this is expected behaviour in v3 but it would make the `t` macro unusable on the server.

**To Reproduce**
Steps to reproduce the behavior, possibly with minimal code sample, e.g:

```
ctx.intl._(t`Hello ${area.name}`)
```
->
```
ctx.intl._(
/*i18n*/
core["i18n"]._("Hello {0}", {
  0: area.name
})),
```

This results in 2 calls of `_` function and breaks variables in translations.

**Expected behavior**
The `t` macro simply converts the message to an object that needs to get passed to an instance of i18n.

**Additional context**

- jsLingui version `3.0.0-7`
- Babel version `7.6.3`
- Your Babel config
```
module.exports = {
  presets: [
    [
      '@babel/preset-env',
      {
        targets: {
          node: 'current',
        },
        loose: false,
        useBuiltIns: false,
        shippedProposals: true,
      },
    ],
    '@babel/preset-react',
    '@babel/preset-flow',
    '@babel/preset-typescript',
  ],
  plugins: [
    'macros',
    ['@babel/plugin-proposal-optional-chaining', { loose: true }],
    ['@babel/plugin-proposal-nullish-coalescing-operator', { loose: true }],
    '@babel/plugin-syntax-export-default-from',
    '@babel/plugin-proposal-export-namespace-from',
    [
      '@babel/plugin-transform-runtime',
      {
        helpers: true,
        regenerator: false,
        useESModules: false,
      },
    ],
  ],
};
```

janicduplessis:
I went through the commits and this change seems to be done on purpose. I think this might need to be reconsidered as it will break regular server usage as well as react server rendering as soon as there are multiple concurrent requests.

tricoder42:
@janicduplessis Yeah, you're right. This is intentional. In the near future, I would like to generate per-locale bundles as described in #508 and for that I need to wrap message and i18n._ call into macro.

I understand that in server apps the bundle size doesn't matter and per-locale bundles are useful only in browser. So what about adding a config `macroStandalone` which would be true by default:

```
/*
 * macroStandalone: true
[...]

PR #1139, opened 2021-09-27
feat(macro): Add support for passing custom i18n instance

jeroenvisser101:
This allows developers that work in concurrent rendering environments
to use a Lingui instance that is locale-specific, without risking global
Lingui to be switched to another locale.

In React, `<I18nProvider>` may be used to provide a locale-specific
Lingui instance, and it can be retrieved by using `useLingui`. After
that, the instance can be passed as argument to the `t` macro and
it will be used instead of the global instance, ensuring deterministic
behaviour.

This addresses concerns raised in https://github.com/lingui/js-lingui/issues/901, https://github.com/lingui/js-lingui/issues/578

Documentation, TypeScript definition changes and (more) tests are 
left to do,  but the main goal of this PR is to receive feedback on the 
approach and to see if this is on the right track.

## Example

<details><summary>Click to expand usage example</summary>

```tsx
import { I18n } from "@lingui/core";
import React, { FC, useCallback } from "react";
import * as Yup from "yup";
import { Field, Form, Formik } from "formik";

function buildYupValidationSchema(i18n: I18n) {
  return Yup.object().shape({
    // 3. Use the `t` macro to translate with the custom instance
    name: Yup.string().required(t(i18n)`Name is required`),
    description: Yup.string().required(t(i18n)`Description is required`),
  });
}

export const ExampleForm: FC = () => {
  // 1. Retrieve the i18n instance from the context
  const { i18n } = useLingui();

  // 2. Pass the i18n instance to functions that need it
  const validationSchema = buildYupValidationSchema(i18n);

  const handleSubmit = useCallback((values) => alert(JSON.stringify(values, null, 2)), []);

  return (
    <Formik
      initialValues={{ name: "", description: "" }}
      validationSchema={validationSchema}
      onSubmit={handleSubmit}
    >
      {({ isSubmitting }) => (
        <Form>
          <Field name="name" type="text" />
          <Field name="description" type="text" />
          <button type="submit" disabled={isSubmitting}>
            Submit
          </button>
        </Form>
      )}
    </Formik>
  );
};
```

</details>

### DIFF
--- a/docs/ref/macro.rst
+++ b/docs/ref/macro.rst
@@ -77,7 +77,7 @@ transformed into ``i18n._`` call.
 
    By default, the ``i18n`` object is imported from ``@lingui/core``.
    If you use a custom instance of ``i18n`` object, you need to set
-   :conf:`runtimeConfigModule`
+   :conf:`runtimeConfigModule` or pass a custom instance to :jsmacro:`t`.
 
 The only exception is :jsmacro:`defineMessage` which is transformed into
 message descriptor. In other words, the message isn't translated directly
@@ -119,6 +119,16 @@ Examples of JS macros
 |                                                             |                                                                    |
 |    t`Attachment ${name} saved`                              |    /*i18n*/                                                        |
 |                                                             |    i18n._("Attachment {name} saved", { name })                     |
++-------------------------------------------------------------+--------------------------------------------------------------------+
+| .. code-block:: js                                          | .. code-block:: js                                                 |
+|                                                             |                                                                    |
+|    t(customI18n)`Refresh inbox`                             |    /*i18n*/                                                        |
+|                                                             |    customI18n._("Refresh inbox")                                   |
++-------------------------------------------------------------+--------------------------------------------------------------------+
+| .. code-block:: js                                          | .. code-block:: js                                                 |
+|                                                             |                                                                    |
+|    t(customI18n)`Attachment ${name} saved`                  |    /*i18n*/                                                        |
+|                                                             |    customI18n._("Attachment {name} saved", { name })               |
 +-------------------------------------------------------------+--------------------------------------------------------------------+
 | .. code-block:: js                                          | .. code-block:: js                                                 |
 |                                                             |                                                                    |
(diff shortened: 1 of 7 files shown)


semoal:
Looks incredibly good Jeroen, not much more to say from my part that we should adjust TypeScript definitions and document this nicely and we're done for shipping this asap :)

jeroenvisser101:
@semoal thanks! I was looking at the types, and found some time to add more documentation to them, along with usage examples that show up when using them in VSCode or other editors with language server integrations:

image

Let me know if there are parts where the wording may differ from the wording used in other places in Lingui, because that could cause confusion so I'd prefer to match them :)

I'll work on the documentation tonight 👍 

PS. I was also still thinking about an additional improvement, to allow a `t`-like macro that works like `defineMessage`, which would enable you to do: ```i18n._(dt`Hello ${world}`)```. It wouldn't differ much from ```defineMessage({message: `Hello ${world}`})```, and so I'll think about it a bit more, but if you have thoughts/ideas about it, I'm happy to hear them :)

jeroenvisser101:
I think the docs are ready to go, they are short, but I also don't think a lot more is to be said about it. Going into it with examples may be useful, but I worried that they would be too specific to one implementation (e.g. React SSR) and not useful for other use-cases, maybe even misleading. Wdyt?

semoal (approved):

semoal:
> I think the docs are ready to go, they are short, but I also don't think a lot more is to be said about it. Going into it with examples may be useful, but I'm worried that they would be too specific to one implementation (e.g. React SSR) and not useful for other use-cases, maybe even misleading. Wdyt?

We're good to go with the current documentation, I think creating an SSR example will too specific.

semoal:
> PS. I was also still thinking about an additional improvement, to allow a t-like macro that works like defineMessage, which would enable you to do: i18n._(dt`Hello ${world}`). It wouldn't differ much from defineMessage({message: `Hello ${world}`}), and so I'll think about it a bit more, but if you have thoughts/ideas about it, I'm happy to hear them :)

Is not a bad idea, but probably making it that similar makes complex to understand our users what's the principal problem of t`` and why does defineMessage exist. Not sure about that, but if you think it's a good addition you can open an issue and we track if people is interested on this 👍 

Thanks for the hard-work mate, really appreciated

jeroenvisser101:
Thanks! I'll consider the use case, maybe the `t(i18n)` will suffice for this use-case :)

jeroenvisser101:
@semoal I think I may have messed up the typescript annotations, I'll look into it and open a follow up PR tonight, sorry for the hassle!

PR #1143, opened 2021-10-02
fix: stripping origin if line numbers already strippped

luludan:
Given config with `lineNumbers: false`, if you run extract, and then subsequently run extract with the watcher on the entire origin is stripped.

Line 30 in `po.ts` was slicing regardless of the length of the array.

Modified this to destructure the array and remove the unnecessary joinOrigin call.

Added a new test as I wasn't quite sure how to modify the existing one to include this case 🤔


semoal:
Thx for the contribution!!

garcia206:
@semoal Any update on when this is going to be released?

semoal:

PR #1163, opened 2021-11-08
docs: the mentioned message id should be "msg.docs"

panlina:
In `Comparison with react-intl / Macros for component-based message syntax`, it shows a sample where the message id is `msg.docs`, then adds a note that says it's optional. In the explanation the previous message id is mentioned after the generated message id. The previous message id is mistakenly called out as `msg.id`.


panlina:
Why the deployment failed? I clicked the detail link and get 404.

PR #1165, opened 2021-11-11
fix: Allow pseudoLocalize to work in @lingui/loader

ggascoigne:
@lingui/loader uses strict mode when it's not running in production.  This is turn means that `compileSingleKey` always gets passed an empty string as `translation` for the pseudo locale to translate.  This seems to defeat the purpose of using pseudo localization.

With this change, you can use @lingui/loader and also get dynamically generated pseudo locale string.


semoal:
Sorry for the late release, has been really busy lately.

PR #1166, opened 2021-11-15
fix(cli): extract with --overwrite should fallback to key

### LINKED ISSUE #1162: Extract `--overwrite` command not using IDs as default message
### Describe the bug

I noticed an issue with `lingui extract --overwrite` command. If I understand well this command is supposed to force the `msgstr` in catalog files to the value set in the source code, right?

This works well when you have custom ids, but if your project has some translations (using `@lingui/macro`) without custom ids, `--overwrite` command will not use the default value as `msgstr`

Maybe there is a configuration that I'm not aware or I'm doing something wrong?

### To Reproduce

```typescript
import { t, Trans } from '@lingui/macro';
import { i18n } from '@lingui/core';
import { I18nProvider } from '@lingui/react'

function App() {
    const translationD = t`Translation test D`;
    const translationE = t({
        id: 'translation_e',
        message: 'Translation test E'
    });
    return (
        <I18nProvider i18n={i18n}>
            {/* Not working ❌ */}
            <p><Trans>Translation test A</Trans></p>
            {/* Not working ❌ */}
            <p><Trans>{`Translation test B`}</Trans></p>
            {/* Working ✅ */}
            <p><Trans id="translation_c">Translation test C</Trans></p>
            {/* Not working ❌ */}
            <p>{translationD}</p>
            {/* Working ✅ */}
            <p>{translationE}</p>
        </I18nProvider>
    );
}

export default App;
```

Exported catalog (without headers)
```po
#: src/App.tsx:23
msgid "Translation test A"
msgstr "" <--  should not be empty

#: src/App.tsx:25
msgid "Translation test B"
msgstr "" <--  should not be empty

#: src/App.tsx:13
msgid "Translation test D"
msgstr "" <--  should not be empty

#: src/App.tsx:26
msgid "translation_c"
msgstr "Translation test C"

#: src/App.tsx:14
msgid "translation_e"
msgstr "Translation test E"
```

To reproduce I created [this repository](https://github.com/ardeois/lingui-extract-overwrite-bug) where you can see multiple variations of `@lingui/macro` usages. Only translations with a defined `message` or custom id will work.

Note also, if you remove all `msgid` and `msgstr` from the source catalog, then run `yarn extract` it works the first time (`msgstr` are populated with the default id)
However, if you rerun the command, it removes them

### Expected behavior
`--overwrite` option should reset `msgstr` to:
- custom `message` if defined
- default `id` otherwise

### Additional context
Check this video for steps to reproduce

https://user-images.githubusercontent.com/1867939/140570251-b38b5565-1e21-4893-bbef-daa3ae8f8564.mp4

- jsLingui version ``3.12.1``
- Babel version ``7.12.17``
- Using Create React App

ardeois:
When we run `lingui extract --overwrite` only translations with a custom id + message were working properly.
However, if there is no default `message` but only the `id`, `--overwrite` option was resetting the translation to empty string instead of using `key` as default

This was working properly when a "new" key was added, but not when we update the key (on merge)

Also update snapshot `go-gettext.test.ts.snap` because it fails on master

Fixes #1162 
See [this repository](https://github.com/ardeois/lingui-extract-overwrite-bug) to reproduce the issue

### DIFF
--- a/packages/cli/src/api/catalog.test.ts
+++ b/packages/cli/src/api/catalog.test.ts
@@ -330,11 +330,17 @@ describe("Catalog", function () {
             message: "",
             translation: "Message with custom ID",
           }),
+          "Message with <0>auto-generated</0> ID": makePrevMessage({
+            translation: "Source of message with <0>auto-generated</0> ID",
+          }),
         },
         cs: {
           "custom.id": makePrevMessage({
             message: "",
             translation: "Translation of message with custom ID",
+          }),
+          "Message with <0>auto-generated</0> ID": makePrevMessage({
+            translation: "Translation of message with auto-generated ID",
           }),
         },
       }
@@ -343,10 +349,11 @@ describe("Catalog", function () {
         "custom.id": makeNextMessage({
           message: "Message with custom ID, possibly changed",
         }),
+        "Message with <0>auto-generated</0> ID": makeNextMessage(),
       }
 
       // Without `overwrite`:
-      // The translation of `custom.id` message for `sourceLocale` is kept intact
+      // The translations of all IDs for `sourceLocale` are kept intact
       expect(
         makeCatalog({ sourceLocale: "en" }).merge(
           prevCatalogs,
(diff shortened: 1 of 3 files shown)


ardeois:
@semoal whenever you have some time to check this out, let me know if the fix I suggest for issue #1162 is correct

Also the Vercel deployment [failed](https://vercel.com/lingui-js/js-lingui/8ZX6E4x3uiUS9ZY3UbPnkgRPnDMm), but I don't have access to the log, so I'm not sure if I can do something about it?

Thanks !!

semoal:
Super! Thanks mate

semoal:
Sorry for the late release, has been really busy lately.

PR #1170, opened 2021-11-22
feat(cli): make extract cmd asynchronous execution

SergeyPoluektov:
Some [custom extractors](https://lingui.js.org/ref/conf.html#extractors) are required to have async execution. For instance, in order to extract messages from Svelte+Typescript you need to preprocess svelte files before compilation. Preprocessing is an async operation in Svelte (see [doc](https://svelte.dev/docs#svelte_preprocess)). But lingui-cli extract command doesn't support async extract method from custom extractor – see `catalog.make` method in `packages/cli/src/lingui-extract.ts` and all function calls down to `extract` method of a particular extractor.

This PR adds support for async extractors by awaiting for `extract` is finished before calling `traverse` function in `Catalog.collect` method

### DIFF
--- a/packages/cli/src/api/catalog.test.ts
+++ b/packages/cli/src/api/catalog.test.ts
@@ -35,7 +35,7 @@ describe("Catalog", function () {
   })
 
   describe("make", function () {
-    it("should collect and write catalogs", function () {
+    it("should collect and write catalogs", async function () {
       const localeDir = copyFixture(fixture("locales", "initial"))
       const catalog = new Catalog(
         {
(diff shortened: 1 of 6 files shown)


semoal:
I would love to receive a pull request with a real example of Lingui + Svelte in our monorepo mate, I'm not experienced with Svelte and would love to get Lingui compatible with him.

SergeyPoluektov:
> I would love to receive a pull request with a real example of Lingui + Svelte in our monorepo mate, I'm not experienced with Svelte and would love to get Lingui compatible with him.

Yes, that would be great to have in there. I'll prepare an example with Svelte and a custom extractor.

PR #1174, opened 2021-12-07
fix: cloud service `import()` on case-sensitive filesystems

MichaelHoste:
We found an issue on case sensitive systems where the cloud service (if one was present) was not loading correctly.

We made sure to lowercase the first letter so that it works on both case-sensitive and case-insensitive systems.


semoal:
Hi 👋 Michaël!  LGTM, will release a new version today

MichaelHoste:
Thanks a lot for your quick response @semoal ! 🥳

PR #1176, opened 2021-12-10
docs: Update showroom.rst

petrpisek:
We're using lingui for the adventure booking portal Ascendero.com. Great tool, helps a lot, thanks.


semoal:
Thanks for the update, consider donating/sponsorship for helping us maintaining the library =)

PR #1220, opened 2022-03-24
docs: Fix missing closing tag in jsx example

marcvangend:
I noticed two places where example JSX code is invalid because of a missing `</span>` tag. This PR fixes it.


skovhus:
Looks good to me. :)

PR #1222, opened 2022-03-29
docs: Use placeholder in selectOrdinal's examples and test cases

orangain:
Currently, the following example is used in the [selectOrdinal's documentation](https://lingui.js.org/ref/macro.html#jsmacro-selectOrdinal). However, this code may return a wrong result for larger numbers. For example, it returns `"1st"` when `count` is `21` . Expected behavior is to return `"21st"`. 

```ts
   const message = selectOrdinal(count, {
      one: "1st",
      two: "2nd",
      few: "3rd",
      other: "#th",
   })
```

This is not a problem of the Lingui itself, but a problem of the documentation. I think the document should suggest proper use of placeholders as follows:

```ts
   const message = selectOrdinal(count, {
      one: "#st",
      two: "#nd",
      few: "#rd",
      other: "#th",
   })
```

This code returns `"21st"` when `count` is `21` as expected.

This PR fixes selectOrdinal's examples and test cases to use placeholders. Note that these placeholders are already used in some test cases like [js-selectOrdinal.ts](https://github.com/lingui/js-lingui/blob/v3.13.2/packages/macro/test/js-selectOrdinal.ts) and [jsx-selectOrdinal.ts](https://github.com/lingui/js-lingui/blob/main/packages/macro/test/jsx-selectOrdinal.ts).

### DIFF
--- a/docs/ref/macro.rst
+++ b/docs/ref/macro.rst
@@ -456,9 +456,9 @@ cardinal plural forms it uses ordinal forms:
 
    import { selectOrdinal } from "@lingui/macro"
    const message = selectOrdinal(count, {
-      one: "1st",
-      two: "2nd",
-      few: "3rd",
+      one: "#st",
+      two: "#nd",
+      few: "#rd",
       other: "#th",
    })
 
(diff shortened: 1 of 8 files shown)


semoal:
Thanks @orangain , nice work as always

Issue #1225, opened 2022-04-05
Quoted JSX attributes use XML-style escapes


skovhus:
**Describe the bug**
Currently the macro transforms children with quotes `"` into invalid JSX.

**To Reproduce**
```typescript
import { Trans } from "@lingui/react"

export default function App() {
   return <Trans>This "should" be quoted correctly!</Trans>
}
```

Currently this is transformed to `<Trans id="This \"should\" be quoted correctly!" />` which is invalid JSX due to the escaping. esbuild, Babel and TypeScript treat this as a syntax error.

Repl examples:
- [esbuild repl example](https://hyrious.me/esbuild-repl/?version=0.14.31&mode=transform&input=%3Cdiv+id%3D%22Say+%5C%22hi%5C%22%21%22+%2F%3E&options=--loader%3Djsx)
- [Babel repl example](https://babeljs.io/repl/#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=DwEwlgbgBGILwCIAOAbApgQwM5qgHWQCc0ssCoAXAC1wCMBXCigewDsEA-YAenAg6A&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2&prettier=false&targets=&version=7.17.8&externalPlugins=&assumptions=%7B%7D)

**Expected behavior**
It should be transformed to valid JSX: `<Trans id={"This \"should\" be quoted correctly!"} />`

**Additional context**

- jsLingui version ``latest``
- Babel version ``latest``

skovhus:
Turns out this is not only an issue for qoutes in the auto generated `id`. Example:
```jsx
<Trans id="custom-id">This should also be "quoted"</Trans>
```

skovhus:
Turns out there are other cases related to this where the `jsxAttribute` string literal value should be wrapped in an `jsxExpressionContainer`:
```jsx
<Trans>
  <Text>This should work &nbsp;</Text>
</Trans>
```

Removing the conditional introduced in #1226 solves this.

PR #1226, opened 2022-04-05
fix: JSX attribute parsing issue when using double quotes

skovhus:
- Fixes https://github.com/lingui/js-lingui/issues/1225 

This was the least intrusive way I could come up with by wrapping the relevant `jsxAttribute` string literal value in a `jsxExpressionContainer`. 

This is currently done conditionally in case the message includes a double quote. Note that I believe this could safely be done unconditionally, but it requires updating a lot of tests. I happily do this if we like the suggested solution.

This has been tested out on a fairly big codebase.

### DIFF
--- a/packages/macro/src/macroJsx.ts
+++ b/packages/macro/src/macroJsx.ts
@@ -51,6 +51,21 @@ export default class MacroJSX {
     this.elementIndex = makeCounter()
   }
 
+  safeJsxAttribute = (name: string, value: string) => {
+    // Quoted JSX attributes use XML-style escapes instead of JavaScript-style escapes.
+    // This means that <Trans id="Say \"hi\"!" /> is invalid, but <Trans id={"Say \"hi\"!"} /> is valid.
+
+    // We could consider removing this condition and always wrap in a jsxExpressionContainer.
+    const attributeValue = value.includes('"')
+      ? this.types.jsxExpressionContainer(this.types.stringLiteral(value))
+      : this.types.stringLiteral(value)
+
+    return this.types.jsxAttribute(
+      this.types.jsxIdentifier(name),
+      attributeValue
+    )
+  }
+
   replacePath = (path: NodePath) => {
     const tokens = this.tokenizeNode(path.node)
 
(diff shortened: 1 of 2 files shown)


skovhus:
@semoal Note that I've been testing this out on a fairly big project and haven't identified any issues.

skovhus:
Let me know if I can help fixing the Vercel CI error.

packages/macro/src/macroJsx.ts
```diff
@@ -87,11 +87,15 @@ export default class MacroJSX {
         }
       }
     } else {
+      // Quoted JSX attributes use XML-style escapes instead of JavaScript-style escapes.
+      // This means that <Trans id="Say \"hi\"!" /> is invalid, but <Trans id={"Say \"hi\"!"} /> works.
+      // We could consider removing the condition here and always wrap in a jsxExpressionContainer.
```
skovhus, line 92: I tested out removing this conditional and didn’t find any regressions. It might fix other cases that the double quoting issue. 

Let me know if we should remove the conditional here.

packages/macro/src/macroJsx.ts
```diff
@@ -51,6 +51,21 @@ export default class MacroJSX {
     this.elementIndex = makeCounter()
   }
 
+  safeJsxAttribute = (name: string, value: string) => {
+    // Quoted JSX attributes use XML-style escapes instead of JavaScript-style escapes.
+    // This means that <Trans id="Say \"hi\"!" /> is invalid, but <Trans id={"Say \"hi\"!"} /> is valid.
+
+    // We could consider removing this condition and always wrap in a jsxExpressionContainer.
```
skovhus, line 58: Removing the conditional here seems safe – I've done this in a larger codebase. 

Let me know if we should remove the conditional here (a lot of the tests would need to be updated).

semoal:
Super cool, thanks for the contribution mate

packages/macro/src/macroJsx.ts
skovhus: @semoal Let me know if we should remove the conditional here (a lot of the tests would need to be updated).

semoal: Will that imply a breaking change? Sorry but I don't get why we should update the tests?

skovhus: It will not introduce a breaking change, but as all id and and message attribute will now be wrapped in an expression, then all the tests would need to be updated.

So instead of `<Trans id="test" />` it would be `<Trans id={"test"} />`. Again not breaking change.

Keeping the conditional might be a fine tradeoff, then we only wrap the attributes in an expression in case there is a `"` in them.

skovhus: We should remove the conditional here to handle additional corner cases. This was done in https://github.com/lingui/js-lingui/pull/1234

PR #1230, opened 2022-04-11
fix: specify children for react v18 types

christopherklint97:
## Short summary

After some changes in react v18, the `FunctionComponent` type no longer implicitly passes `children` as part of the props. This means that the `children` prop needs to be declared explicitly.

### DIFF
--- a/packages/macro/index.d.ts
+++ b/packages/macro/index.d.ts
@@ -172,6 +172,7 @@ export type TransProps = {
   comment?: string
   values?: Record<string, unknown>
   context?: string
+  children?: string | any[] | React.ReactNode
   component?: React.ComponentType<TransRenderProps>
   render?: (props: TransRenderProps) => ReactElement<any, any> | null
 }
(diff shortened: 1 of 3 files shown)


packages/macro/index.d.ts
```diff
@@ -172,6 +172,7 @@ export type TransProps = {
   comment?: string
   values?: Record<string, unknown>
   context?: string
+  children?: string | any[] | React.ReactNode
```
ggascoigne, line 175: Shouldn't that just be  

```ts
children?: ReactNode | undefined
```

That was the previous definition.  The `string` part is already part of `ReactNode`, and the `any[]` seems too wide.

christopherklint97: That is right, have fixed it now 😄

ggascoigne (commented):
Well I'm not part of this project so I don't feel that I can approve this, but I can say that I think that this is the correct change.

semoal:
Thanks guys for the contribution, will take care of the new pull requests and new issues in a few days! Been doing some home renovations and couldn’t find time to oss 🤟🏻

semoal:
Thanks guys, will release a new version at the end of today

PR #1234, opened 2022-04-26
fix: JSX attribute parsing issue when using html entities

skovhus:
Issue: https://github.com/lingui/js-lingui/issues/1225

https://github.com/lingui/js-lingui/pull/1226 added support for handling quotes in JSX attributes. Turns out there are other cases related to this where the `jsxAttribute` string literal value should be wrapped in an `jsxExpressionContainer`:
```jsx
<Trans>
  <Text>This should work &nbsp;</Text>
</Trans>
```

The case here fails in compilers such as ESBuild. This requires updated a lot of test expectations, but the change is **not a breaking change**.


skovhus:
@semoal the size testing check failure seems unrelated.

skovhus:
@semoal let me know if there is anything I can do to land this.

skovhus:
> <Trans id="Hello World" /> with this pr, will be mandatory use brackets?

No you can still do that. Note that it is just all the snapshots that are updated – not the input. 

There should be a `jsxExpressionContainer` (i.e. brackets) in order to support cases like: 
```jsx
<Trans>
  <Text>This should work &nbsp;</Text>
</Trans>
```

skovhus:
Let me know if there is anything needed before releasing this. Thanks!

semoal:
> Let me know if there is anything needed before releasing this. Thanks!

Just released 3.14.0 :)

dahei:
I stumbled over that problem while migrating our project from Webpack to Vite (which uses esbuild). Upgrade to 3.17.2 make it work 👍

PR #1245, opened 2022-05-20
add Fider to showroom

goenning:


semoal:
Glad that you use Lingui with that amazing product you built.
How was your experience, something you felt that needs an improvement?

Thanks!!

Issue #1254, opened 2022-06-16
lingui should ship ESM files in `.mjs` extension


ocavue:
**Describe the bug**

Although lingui has `esm` build under the `dist/`, but it cannot be import from a ESM file when using Node.js.

I found this issue when I try to run some unit tests on the Node.js side. 

**To Reproduce**

```
bash-3.2$ npm add @lingui/core
bash-3.2$
bash-3.2$ echo 'import {i18n} from "@lingui/core"; console.log(i18n);' > ./file.mjs
bash-3.2$ node ./file.mjs
file:///private/tmp/npm-t/file.mjs:1
import {i18n} from "@lingui/core"; console.log(i18n);
        ^^^^
SyntaxError: Named export 'i18n' not found. The requested module '@lingui/core' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@lingui/core';
const {i18n} = pkg;

    at ModuleJob._instantiate (node:internal/modules/esm/module_job:128:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:194:5)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:385:24)
    at async loadESM (node:internal/process/esm_loader:88:5)
    at async handleMainPromise (node:internal/modules/run_main:61:12)
bash-3.2$
```

**Expected behavior**

Lingui should be able to be imported in a `.mjs` file in Node.js environment. 

**Additional context**

- Here is an explanation of these kinds of errors: https://github.com/sheremet-va/dual-packaging
- Two simplest solutions are 
  - Rename all file under `dist/esm/` from `.js` to `.mjs`, or 
  - Rename all file under `dist/cjs/` from `.js` to `.cjs` and add `"type": "module"` in `package.json`. 

- jsLingui version `"@lingui/core": "3.13.3"`
- Babel version: N/A
- Your Babel config (e.g. `.babelrc`) or framework you use (Create React App, Meteor, etc.): Node.js v16.15

PR #1255, opened 2022-06-18
fix: output .mjs files while bundling esm

ocavue:
# Description

Rename the bundled files from `.js` to `.mjs` under `esm/`. This allows lingui to be imported in Node.js with the ESM syntax. 

Fixes https://github.com/lingui/js-lingui/issues/1254

# Checklist

- [ ] Includes tests
- [ ] Includes documentation

### DIFF
--- a/packages/core/npm/esm/dev.mjs
+++ b/packages/core/npm/esm/dev.mjs
@@ -3,14 +3,14 @@ import {
   setupI18n as setupI18nProd,
   formats as formatsProd,
   I18n as I18nProd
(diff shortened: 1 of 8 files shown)


semoal:
Will this imply a breaking change release? Someone who their system is not resolving .mjs, will crash right?

ocavue:
You have a point there. Instead of changing existed files, maybe adding new files is better for compatibility. I'll close this RP and open another one https://github.com/lingui/js-lingui/pull/1258

PR #1267, opened 2022-07-04
docs: removed the extra ` before :js:func

hehex9:
https://lingui.js.org/ref/react.html#providers

image

it makes the `useLingui` unclickable


hehex9:
🤕

PR #1276, opened 2022-08-14
feat Add support for oct escaping

mordicool-medigate-io:
Hi,

Can we please add escaping for the # in the replaceOctothorpe function?

This will be very appreciated, and enable the use of # as part of the plural macro 👍🏻

### DIFF
--- a/packages/core/src/context.ts
+++ b/packages/core/src/context.ts
@@ -3,6 +3,9 @@ import { date, number } from "./formats"
 import { isString, isFunction } from "./essentials"
 
 export const UNICODE_REGEX = /\\u[a-fA-F0-9]{4}|\\x[a-fA-F0-9]{2}/g;
+const OCT_REGEX = /(.)(#)/g;
+const OCT_ESCAPE_CHARACTER = '\\';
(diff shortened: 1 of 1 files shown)


packages/core/src/context.ts
```diff
@@ -3,6 +3,9 @@ import { date, number } from "./formats"
 import { isString, isFunction } from "./essentials"
 
 export const UNICODE_REGEX = /\\u[a-fA-F0-9]{4}|\\x[a-fA-F0-9]{2}/g;
+const OCT_REGEX = /(.)(#)/g;
```
avshalomt2, line 6: Maybe `PLURAL_PLACEHOLDER_REGEX`?
This doesn't handle the case where hash is the first character (I've slacked you a solution that works in that case).

packages/core/src/context.ts
```diff
@@ -3,6 +3,9 @@ import { date, number } from "./formats"
...
+const OCT_ESCAPE_CHARACTER = '\\';
+const octReplacer = (m, valueStr) => m.replace(OCT_REGEX, (match, p1, p2) => p1 === OCT_ESCAPE_CHARACTER ? p2 : p1 + valueStr);
```
avshalomt2, line 8: Please give more meaningful var names

avshalomt2 (changes requested):

PR #1279, opened 2022-08-25
fix: extract-template command

thekip:
## Issue: 

If you don't have a *.pot file on the disk the command will fail with:

```
catalogStats[catalog.templateFile] = Object.keys(catalog.readTemplate()).length;
TypeError: Cannot convert undefined or null to object
```

If you create a file manually (just empty file), the command will complete successfully but catalog statistics will show outdated info. 

> Catalog statistics for /translates/messages.pot: 0 messages

Both problems caused by `catalog.makeTemplate` which is async function, but code don't await for its execution. 

This PR tackle this issue, and adds await for these async calls.

### DIFF
--- a/packages/cli/src/lingui-extract-template.ts
+++ b/packages/cli/src/lingui-extract-template.ts
@@ -14,10 +14,10 @@ export type CliExtractTemplateOptions = {
   files?: string[]
 }
 
-export default function command(
+export default async function command(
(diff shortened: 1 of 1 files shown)


thekip:
I also quite surprised that there are no even primitive smoke tests covering cli commands.

PR #1281, opened 2022-09-14
improvement: Context

### LINKED ISSUE #1198: Message context (msgctxt) not working
**Describe the bug**
When providing context to a message, the context is treated like an ID in the generated `en.po` file.

**To Reproduce**
With the version `3.13.1` with `create-react-app`, and not specifying `orderBy` in `lingui.config.js`:

```js

  return <div>
    <Trans context='ctxtA'>My string</Trans>
    <Trans context='ctxtB'>My string</Trans>
    <Trans>My string</Trans>
  </div>
}
```

**Expected behavior**
3 messages should be created in the generated `en.po` file, like this:
```po
#: src/App.tsx:11
msgid "My string"
msgstr "My string"

#: src/App.tsx:12
msgctxt "ctxtA"
msgid "My string"
msgstr "My string"

#: src/App.tsx:13
msgctxt "ctxtB"
msgid "My string"
msgstr "My string"
```

**Actual behavior**
This is created in the `en.po` file.
```po
#: src/App.tsx:11
msgid "My string"
msgstr "My string"

msgid "ctxtA"
msgstr "ctxtA"

msgid "ctxtB"
msgstr "ctxtB"
```

**Additional context**
If using `orderBy: 'origin'` in `lingui.config.js`, an exception is thrown in `getFirstOrigin()` when running `extract`.

- jsLingui version ``3.13.1``
- Using `create-react-app`

nmorel:
Hello 👋

@JSteunou and I needed the context to work for our company so we tried to fix the different issues we encountered with the current implementation.  

We tried to follow the same "schema" for the internal structure and limit the changes.  
It works (closes #1198) but it has some flaws.  
 
For an example of the structure, for this [file](https://github.com/nmorel/js-lingui/blob/context/packages/cli/src/api/fixtures/collect/componentA/componentA.js), it gives the following on extraction :  
```json
{
  "Component A": {
    "extractedComments": [],
    "origin": [["collect/componentA/componentA.js", 1]]
  },
  "Hello World": {
    "extractedComments": ["Comment A", "Comment A again"],
    "origin": [
      ["collect/componentA/componentA.js", 2],
      ["collect/componentA/componentA.js", 3]
    ]
  },
  "world ctxt": {
    "Hello World": {
      "extractedComments": [],
      "origin": [["collect/componentA/componentA.js", 4]]
    }
  }
}
```

With this, we no longer have a flat structure and context add another level.  
We now have at the same level a key that can represent a message id or a context.  
At various location, I added a [test](https://github.com/nmorel/js-lingui/blob/context/packages/cli/src/api/catalog.ts#L320) to know which one I'm manipulating.  
If a user needs to translate the words `origin` or `translation` with a context, it breaks.  
This structure also add lots of complexity.

Before we fix the remaining issues, what do you think ?  
Should we change something ?  

Some solutions we thought about but wanted your impressions first :  
- change the key to add the context like described [here](https://github.com/lingui/js-lingui/issues/856#issuecomment-729979886)
- add some default context in the structure : 
```json
{
  "_": {
    "Component A": {
      "extractedComments": [],
      "origin": [["collect/componentA/componentA.js", 1]]
    },
    "Hello World": {
      "extractedComments": ["Comment A", "Comment A again"],
      "origin": [
        ["collect/componentA/componentA.js", 2],
        ["collect/componentA/componentA.js", 3]
      ]
    }
  },
  "world ctxt": {
    "Hello World": {
      "extractedComments": [],
      "origin": [["collect/componentA/componentA.js", 4]]
    }
  }
}
```

For a complete example, I updated [one of the repository example](https://github.com/nmorel/js-lingui/commit/bd5e22c469a42d28504ab18bc33203f4a35438a4) (the ^3.14.1 version references a local version I built).


JSteunou:
As @nmorel said, it fixes the primary issue without changing the logic & structure already in place.

I think it would be better to focus on this here, and if some changes should be made, tailor this in another PR.

nmorel:
One thing I forgot to add, I'm not used to ramda 😅  
So there is a few places where I used vanilla JS api.

semoal:
Will review this afternoon guys, thanks for the hard work!

examples/create-react-app/package.json
```diff
@@ -3,8 +3,8 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@lingui/react": "^3.13.2",
-    "@lingui/core": "^3.13.2",
+    "@lingui/react": "^3.14.1",
```
andrii-bodnar, line 6: Maybe, it would be better to increment the minor version for this Context improvement? **v3.15.0**. 

@semoal what do you think about that? 🙂

semoal: Yep, it must be as a minor version; Technically we use conventional commits; If it's squashed and rebased to main with a commit title like: `feat(): add message context `
Will try to review the pull request on this morning, before merging :)

andrii-bodnar: Hi @semoal!
Have you had a chance to review this PR?

thekip:
@semoal is something in that PR what stops it from being merged?

andrii-bodnar:
@thekip could you please review this PR? 🙂 

@nmorel It would be great if you could rebase the branch as it currently has conflicts that need to be resolved, sorry for the inconvenience

nmorel:
@andrii-bodnar I prefer to get feedback on the PR first before putting any more work on current branch.  
The fix isn't perfect and I still have some unanswered questions.

thekip:
@andrii-bodnar @nmorel I reviewed the code in general it looks good to me. 

Here are my thoughts (doesn't mean you have to do this, just want to rise an opinion): 
1. It's probably would be better to add in the docs more real-life examples. Right now, docs for context would be good only for people how already know what context is and would look for that in the docs. 

Example from other i18n project docs which I've found helpful: 

> The same text elements with different meanings are extracted with different IDs. For example, if the word "right" uses the following two definitions in two different locations, the word is translated differently and merged back into the application as different translation entries.
>
> correct as in "you are right"
> direction as in "turn right"
> If the same text elements meet the following conditions, the text elements are extracted only once and use the same ID.

2. Relaying on the field `origin` and `translation` feels a bit dirty and fragile. I like option with default context in internal structures. So lookup would always be as `catalog[context || DEFAULT_CONTEXT][messageId]`.  This also would be nicer in the code as you will not have to copying these checks everywhere `(message.hasOwnProperty('origin') || message.hasOwnProperty('translation') || message.context) `

But actually i want to rise a bigger discussion here, read at the end of the comment. 

3. How context would be serialized in formats other than `.po`? I didn't check the code in action, unfortunately. Would it bring breaking changes in already existing catalogs? 

### Discussion: 

Implementation offered in this PR is ok just as quick workaround for such essential feature as context. And this is fine. 
However, I think better implementation is needed here, but it would require bigger changes in sourcecode and how the lib would be used. 

Part of this was already discussed [here](https://github.com/lingui/js-lingui/issues/856#issuecomment-729979886), I want to summarize. 


Current lingui implementation has a flaw:  you have your original message in the bundle at least 2 times + 1 translation. 
For a line "Hello world" you would have one in the source code as id in i18n call. Another one as key in the message catalog and then a translation itself. Strings could be very long, not just a couple lines, so this may bring more kb to the bundle. 

This PR would also add even more kilobytes because for each msg which should have a context you would have context explicitly written in the bundle + in the msg catalog. 

A much better option would be generating a "stable" ID based on the msg + context as hash with fixed length. 
So internal structure would look like 
```
{
 [<hash>]: {..}
}
```
And lookups in runtime would be much simpler. 

Hash would be calculated at build time by macros  + compile time for catalogs. So macros instead of

```ts
const message = t({
   context: 'My context',
   message: `Hello`
})

// ↓ ↓ ↓ ↓ ↓ ↓

const message = i18n._(/*i18n*/{
   context: 'My context',
   id: `Hello`
})
```
Would generate: 

```ts
const message = i18n._(/*i18n*/{
   id: "<hash(message + context)>",
   message: `Hello`, // only for development purpose, would be dropped in production
})
```

Compile catalog command would also calculate hashes from msg + context on the fly. 
So context would affect only how msg id generates and will not have it's place in the bundle.

However, there are amounts of users who use a lib without a macros. I don't know who they are and what were the reasons behind this decision. I think if you don't use a macro, you lose everything what making lingui awesome. Without a macro there are better alternatives with more stars and bigger community. So I think we should concentrate on one thing and do it well instead of supporting multiple different usages. 

This optional status of the macro and necessity of implementing all existing features in 2 modes adds additional complexity to the project and prevents the possibility of implementing some really cool features because we always should think about "runtime" users. 

Anyway, it would be great to know how many of users don't use a macro and how painful would be dropping support of the non-macro usage.

@andrii-bodnar what do you think? We can discuss dropping non-macro usage separately somewhere. Depending on this, we can later decide a what to do with this PR.

PS this proposal with build-time id generation still can be used without dropping non-macro users, but it would require 2 runtimes, 2 sets of tests and confusing setup options, that's why I want to avoid it. And avoid generating hashes at runtime as well.

thekip:
I read test snapshots carefully one more time and the following came up. If we bring "a default" context as was proposed, this brings "nested" structures to these catalog formats:

Lingui (Raw) and Simple JSON catalog formats 
https://lingui.js.org/ref/catalog-formats.html#json. 

If we do that, we also need to provide an upgrade path for the users. This should be relatively simple.
However, if nested structure would be consumed by some translation tools, such as crowdin, it's probably would be treated as one entry, so it may break a flow for some. 

So I would say context feature not really compatible with supported JSON formats. 

Also generating a "hashed" id as proposed by me would definitely break simple JSON format flow, because key would no longer be an original string.

That's another questions to discuss.

nmorel:
For the default context, it's possible I think to introduce it without breaking change.  
It will only be present in the internal code. When we write to a file (json/js/po..), the default context can be removed.

thekip:
@nmorel could you elaborate on this more?

nmorel:
If I take examples from [here](https://lingui.js.org/ref/catalog-formats.html), we will have this with the current PR :    

- po   
```po
#: src/App.js:3
#  Comment for translators
msgid "messageId"
msgstr "Translated Message"

#: src/App.js:4
#  Comment for translators
msgctxt "messageCtxt"
msgid "messageId"
msgstr "Translated Message with ctxt"
```  
- json   
```json
{
  "messageId": "Translated Message",
  "messageCtxt": {
    "messageId": "Translated Message with ctxt"
  }
}
```
- raw   
```json
{
  "messageId": {
    "translation": "Translated message",
    "defaults": "Default message",
    "description": "Comment for translators",
    "origin": [["src/App.js", 3]]
  },
  "messageCtxt": {
    "messageId": {
      "translation": "Translated Message with ctxt",
      "defaults": "Default message",
      "description": "Comment for translators",
      "origin": [["src/App.js", 4]]
    },
  }
}
```

We can keep this format and only introduce the default context internally with a structure like this :  
```json
{
  "<user context or default context>": {
    "<message id>": {
      "translation": "<message translation>",
      "defaults": "<default message>",
      "description": "<comments>",
      "origin": [<origin1>, <origin2>]
    }
  }
}
```

When parsing po files, if there is no context, we set the default context to the message.  
When parsing json files, if the value is a string, we set the default context to the message.   
Raw files is the hardest to parse. We could test origin/translation fields like the PR to assign the default context or introduce a special property to context to recognize them easily :   
```json 
{
  "messageCtxt": {
    "__isContext__": true,
    "messageId": {
      "translation": "Translated Message with ctxt",
      "defaults": "Default message",
      "description": "Comment for translators",
      "origin": [["src/App.js", 4]]
    },
  }
}
```  

When converting the internal structure to po/json/raw files, it won't be too hard to remove the default context.  

With this, we keep the format intact (except for message with context) and in lingui code, the structure is the same.  
We no longer need to test if the object we manipulate is a message or a context containing messages.  

But if you want to introduce the hash solution soon, is it really necessary ? 🙂

andrii-bodnar:
@nmorel @thekip thanks a lot for your involvement here!

I'm still learning the Lingui and it's a bit hard to make decisions about the concerns above. @Martin005 maybe you will have some thoughts?

Anyway, we'd like to introduce this feature in the nearest release. It's important to keep it backward compatible and not to provide any breaking changes for now.

Could you please guys briefly summarize what is currently blocking this PR?

And a small general question - what's the difference between `comment` and `context` here? For me, it has quite a similar purpose - providing some additional information for translation about the string

thekip:
@andrii-bodnar the problem here if we release current implementation we would be forced to maintain it as it is right now. But the current approach has many flaws and will limit us in the implementing of new features.

So instead of making a release right now and returning to this question again and again in future we propose to implement it properly from the begining and probably ship it as version 4. 

> And a small general question - what's the difference between comment and context here? For me, it has quite a similar purpose - providing some additional information for translation about the string

This is a common for i18n process to have both. The might be called diffrently, f ex. meaning / description or context / comment as in Lingui. The purpose of context is provide few diffrent translation for one source message:  

> correct as in "you are right"
> direction as in "turn right"

Unlike context, comment don't affect message id generation, and used just as an additional information for the translator. 
For example context could be "direction" where comment could be "used on the header in metablock, please keep it 20 symbols length max"

JSteunou:
Just to throw another stone here: it seems *better* to do things right, with a v4, *if* the v4 is really coming soon. But if it is coming soon, then there is no worries about supporting the v3, so the argument about being forced to maintain it turn void. So ship it quick and iterate would be better and non blocking.

On the contrary if the v4 is not coming soon, then it is also better to ship it now in v3 (maybe with the changes @nmorel explain above).

TLDR; in both cases I think it is better for users to ship it now.

thekip:
Suppressed by https://github.com/lingui/js-lingui/pull/1440

Issue #1213, opened 2022-02-25
Fallback locales not taken into account for `pseudoLocale`


um-synth:
**Describe the bug**
Hi. 

When setting up a pseudo local I have a problem that instead of using values of fallback `locale`, pseudo local messages are generated from the `ids`. In our case ids are not the same as default values, which makes pseudo local language a bit useless. 
Not sure if this is intentional or a bug? Following the documentation, I would consider that a bug. 

**To Reproduce**
Our `.linguirc` setup:

```
{
  "locales": [
    "en",
    "en-PL"
  ],
  "sourceLocale": "en",
  "pseudoLocale": "en-PL",
  "fallbackLocales": {
    "en-PL": "en",
    "default": "en"
  },
  "catalogs": [
    {
      "path": "src/locales/{locale}/messages",
      "include": [
        "src"
      ]
    }
  ],
  "format": "po"
}


```

Please let me know if you need any additional information.

sharonyogev:
Hi, this is because [`compileSingleKey`](https://github.com/lingui/js-lingui/blob/0e270a394e3317dd243519445feafae8b6e3214c/packages/cli/src/api/compile.ts#L28) uses the `key` and not the `translation`
But, when I tried to replace it with `translation` some of the translations failed the compilation process, for example:
* `{count, plural, one {boy} other {# boys}} {count, plural, one {girl} other {# girls}}`
* `{class} {count, plural, one {boy} other {# boys}}`

As I see it, there are 2 solutions:

1. try to fix the escaping in [pseudoLocalize.ts](https://github.com/lingui/js-lingui/blob/main/packages/cli/src/api/pseudoLocalize.ts) - I think it will be hard/impossible to be bulletproof since the language is not regular (we need to count the number of `{` for example), but more than possible to fix most common issues
2. do the pseolocalization in [compile.ts](https://github.com/lingui/js-lingui/blob/main/packages/cli/src/api/compile.ts) in the `processTokens` function, it's safer since it's after we parsed the message-format and we traverse the AST so we know the type of each token

What do you think?

um-synth:
Hi. Sorry - I moved fwd with `formatjs` for now, since that was quite painful not to have. Though - maybe I will find some time during next month to take a look how this could be done based on the suggestion from @sharonyogev.

PR #1293, opened 2022-10-19
fix: Make pseudolocal use the message AST instead of the key

sharonyogev:
This is fixing this issue: https://github.com/lingui/js-lingui/issues/1213

I'm not sure why in this PR (https://github.com/lingui/js-lingui/pull/1165) there was a change from the message to the key, but I hope I don't mess anything

Also, I wonder if we should remove the `addDelimitersVariables` & `addDelimitersMacro` and keep only the handling of HTML inside the `pseudoLocalize.ts` code

### DIFF
--- a/packages/cli/src/api/compile.test.ts
+++ b/packages/cli/src/api/compile.test.ts
@@ -2,8 +2,8 @@ import generate from "@babel/generator"
 import { compile, createCompiledCatalog } from "./compile"
 
 describe("compile", () => {
-  const getSource = (message: string, shouldPseudolocalize: boolean = false) =>
-    generate(compile(message, shouldPseudolocalize) as any, {
+  const getSource = (message: string) =>
+    generate(compile(message) as any, {
       compact: true,
       minified: true,
       jsescOption: { minimal: true },
@@ -85,7 +85,12 @@ describe("compile", () => {
   })
 
   describe("with pseudo-localization", () => {
-    const getPSource = (message: string) => getSource(message, true)
+    const getPSource = (message: string) =>
+      generate(compile(message, true) as any, {
+        compact: true,
+        minified: true,
+        jsescOption: { minimal: true },
+      }).code
 
     it("should pseudolocalize strings", () => {
       expect(getPSource("Martin Černý")).toEqual('"Màŕţĩń Čēŕńý"')
@@ -93,8 +98,8 @@ describe("compile", () => {
 
     it("should pseudolocalize escaping syntax characters", () => {
       // TODO: should this turn into pseudoLocale string?
-      expect(getSource("'{name}'", true)).toEqual('"{name}"')
-      // expect(getSource("'{name}'", true)).toEqual('"{ńàmē}"')
+      expect(getPSource("'{name}'")).toEqual('"{name}"')
+      // expect(getPSource("'{name}'")).toEqual('"{ńàmē}"')
     })
 
     it("should not pseudolocalize arguments", () => {
@@ -111,16 +116,18 @@ describe("compile", () => {
 
     it("should not pseudolocalize HTML tags", () => {
       expect(getPSource('Martin <span id="spanId">Černý</span>')).toEqual(
-        '"Màŕţĩń <span id=\\"spanId\\">Čēŕńý</span>"'
+        JSON.stringify('Màŕţĩń <span id="spanId">Čēŕńý</span>')
       )
       expect(
         getPSource("Martin Cerny  123a<span id='id'>Černý</span>")
-      ).toEqual('"Màŕţĩń Ćēŕńŷ  123à<span id=\'id\'>Čēŕńý</span>"')
+      ).toEqual(
+        JSON.stringify("Màŕţĩń Ćēŕńŷ  123à<span id='id'>Čēŕńý</span>")
+      )
       expect(getPSource("Martin <a title='>>a'>a</a>")).toEqual(
-        '"Màŕţĩń <a title=\'>>a\'>à</a>"'
+        JSON.stringify("Màŕţĩń <a title='>>a'>à</a>")
       )
       expect(getPSource("<a title=TITLE>text</a>")).toEqual(
-        '"<a title=TITLE>ţēxţ</a>"'
+        JSON.stringify("<a title=TITLE>ţēxţ</a>")
       )
     })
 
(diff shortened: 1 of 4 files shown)


packages/cli/src/api/compile.test.ts
```diff
@@ -67,14 +68,165 @@ describe("compile", () => {
...
+        )
+      ).toEqual(
+        '[["bcount","plural",{one:"ƀōŷ",other:["#"," ƀōŷś"]}]," ",["gcount","plural",{one:"ĝĩŕĺ",other:["#"," ĝĩŕĺś"]}]]'
+      )
+    })
+  })
 })
 
 describe("createCompiledCatalog", () => {
+  it("nested message", () => {
```
sharonyogev, line 213: just added this test for more coverage on this file

packages/cli/src/api/compile.test.ts
```diff
@@ -113,7 +265,7 @@ describe("createCompiledCatalog", () => {
         {
           Hello: "Ahoj",
           Missing: "",
-          Select: "{id, select, Gen {Genesis} 1John {1 John}  other {____}}"
+          Select: "{id, select, Gen {Genesis} 1John {1 John}  other {____}}",
```
sharonyogev, line 268: prettier changes (this and the following)

packages/cli/src/api/pseudoLocalize.test.ts
```diff
@@ -50,20 +50,20 @@ describe("PseudoLocalization", () => {
     it("with HTML tags", () => {
       expect(
         pseudoLocalize(
-          "{count, plural, zero {There's # <span>message</span>} other {There are # messages}"
+          "{count, plural, zero {There's # <span>message</span>} other {There are # messages}}"
```
sharonyogev, line 53: these weren't valid ICU messages (it does not make the tests failed, since these tests does not validate it's a valid ICU message), but still need to be correct (unless we'll remove the ICU handling from this file)

packages/cli/src/api/pseudoLocalize.test.ts
```diff
@@ -92,4 +92,14 @@ describe("PseudoLocalization", () => {
     expect(pseudoLocalize("replace {count}")).toEqual("ŕēƥĺàćē {count}")
     expect(pseudoLocalize("replace { count }")).toEqual("ŕēƥĺàćē { count }")
   })
+
+  it("multiple plurals is wrong45", () => {
```
sharonyogev, line 96: this is the test that shows why the currently implemented ICU support in ` pseudoLocalize.ts` is wrong

packages/cli/src/api/pseudoLocalize.test.ts
```diff
@@ -92,4 +92,14 @@ describe("PseudoLocalization", () => {
...
+    expect(
+      pseudoLocalize(
+        "{bcount, plural, one {boy} other {# boys}} {gcount, plural, one {girl} other {# girls}}"
+      )
+    ).not.toEqual(
+      "{bcount, plural, one {ƀōŷ} other {# ƀōŷś}} {gcount, plural, one {ĝĩŕĺ} other {# ĝĩŕĺś}}"
+    )
+  })
 })
```
avshalomt2, line 105: Consider adding a test for `compile` with `shouldPseudolocalize = true`, with ICU strings, and without ICU strings.

packages/cli/src/api/compile.ts
```diff
@@ -118,34 +113,40 @@ function buildExportStatement(expression, namespace: CompiledCatalogNamespace) {
...
   if (isString(ast)) return t.stringLiteral(ast)
 
   return ast
 }
 
-function processTokens(tokens) {
+function processTokens(tokens, shouldPseudolocalize: boolean) {
   // Shortcut - if the message doesn't include any formatting,
   // simply join all string chunks into one message
   if (!tokens.filter((token) => !isString(token)).length) {
```
avshalomt2, line 136: Not your code, but it can be optimized (so it stops iterating once the first non-string item was encountered).
```suggestion
  if (tokens.some((token) => !isString(token))) {
```

packages/cli/src/api/compile.test.ts
```diff
@@ -67,14 +68,165 @@ describe("compile", () => {
...
+      expect(getPSource("{name, number}")).toEqual('[["name","number"]]')
+      expect(getPSource("{name, number, percent}")).toEqual(
+        '[["name","number","percent"]]'
+      )
+    })
+
+    it("should not pseudolocalize HTML tags", () => {
+      expect(getPSource('Martin <span id="spanId">Černý</span>')).toEqual(
+        '"Màŕţĩń <span id=\\"spanId\\">Čēŕńý</span>"'
+      )
```
avshalomt2, line 115: 1. Maybe add a test case with both `'` and `"` in it.
2. Because there's escaping in the string you send to `toEqual`, but also in the JSON it represents, i'd use a template string instead to prevent the double escapes, and possibly anywhere there's any type of quote-mark in a string here

```suggestion
      expect(getPSource(`Martin <span id="spanId">Černý</span>`)).toEqual(
        `"Màŕţĩń <span id=\"spanId\">Čēŕńý</span>"`
      )
```

packages/cli/src/api/compile.test.ts
```diff
@@ -67,14 +68,165 @@ describe("compile", () => {
...
+      expect(
+        getPSource("Martin Cerny  123a<span id='id'>Černý</span>")
+      ).toEqual('"Màŕţĩń Ćēŕńŷ  123à<span id=\'id\'>Čēŕńý</span>"')
+      expect(getPSource("Martin <a title='>>a'>a</a>")).toEqual(
```
avshalomt2, line 119: Maybe add a test with correct HTML escaping?
```
<tag attr="hello&quot;world" attr2="hello&quot;&gt;">...</tag>
```

packages/cli/src/api/pseudoLocalize.test.ts
```diff
@@ -92,4 +92,14 @@ describe("PseudoLocalization", () => {
     expect(pseudoLocalize("replace {count}")).toEqual("ŕēƥĺàćē {count}")
     expect(pseudoLocalize("replace { count }")).toEqual("ŕēƥĺàćē { count }")
   })
+
+  it("multiple plurals pseudolocalize gives wrong ICU message", () => {
```
sharonyogev, line 96: this is the test that shows why the currently implemented ICU support in pseudoLocalize.ts is wrong

PR #1296, opened 2022-10-28
feat: Support GetText template files support during compile

thekip:
This PR brings full e2e support for gettext *.pot files into linguijs. 

## What is a template or *.pot file? 
Template is a file where all messages extracted from sourcecode is stored. It doesn't have any translations. It treated as build artifact and could be added to gitignore. 

## What's wrong with current support of *.pot files? 
Short version: 
They are not taken into account during compiling messages.

Long version:
Standard gettext flow looks like the following
- Change source code and potentially change messages
- Run extractor and produce *.pot file
- Send *.pot file together with translations (en.po, pl.po) to the translators. 
- Translators upload .pot and *.po files into theirs translation software which merges them
- Translators translate
- Translators return translated files. 

There are few things which important to notice. 
Merging new messages to the translations should happen outside of build process. It's the responsibility of translation software (such as PoEditor or modern cloud services like crowdin) to do so, not the linguijs extractor.

In current linguijs flow, extractor tries to merge new messages (and delete old) into translation files. 

What's wrong with this? Translations files are usually under version control. Running a build with NODE_ENV=production without fresh extracted messages will cause that messages where ICU is used would be corrupted and this force developers to always extract messages before build in CI process (for test/staging builds, for production builds we assume that 100% messages are translated).

Every time we run a development build on CI, our translation files marked as changed. 

This might not be a big deal for regular repositories, but could be a dealbreaker for monorepos where some strategies for incremental builds are used. Usually tools such as turborepo / rush / ultra-runner trying to be smart and detect which project inside monorepo was changed and rebuild only this one. With linguijs this doesn't work because every build touches translations and therefore bursting the cache. 

But lingui has a command to `extract-template`, you can use this, you may say.  Actually not.

What happend if we generate template file using `lingui extract-template`? 
- The template would be generated and translations files would not be touched. That's good. 
- You build your application and realize that all plurals expressions doesn't work because they was not compiled because they are not in the translation files - That sucks. 
- Your understand that in current stage, extracting *.pot files are useless in linguijs. 

## How this PR changes it? 
Instead of merging new messages into catalogs during extraction phase, it merges template with requested catalog during compile time. If you have actual template - your translations catalogs are also actual.

### DIFF
--- a/packages/cli/src/api/__snapshots__/catalog.test.ts.snap
+++ b/packages/cli/src/api/__snapshots__/catalog.test.ts.snap
@@ -1,5 +1,19 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Catalog POT Flow Should get translations from template if locale file not presented 1`] = `
+Object {
+  Hello World: Hello World,
+  Test String: Test String,
+}
+`;
+
+exports[`Catalog POT Flow Should merge source messages from template if provided 1`] = `
+Object {
+  Hello World: Cześć świat,
+  Test String: Test String,
+}
+`;
+
 exports[`Catalog collect should extract messages from source files 1`] = `
 Object {
   Component A: Object {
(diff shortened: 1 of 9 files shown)


semoal:
Looks good to me! Thanks for the contribution mate

Issue #1297, opened 2022-11-03
@lingui/macro 3.x slows down builds + tests and leaks memory


decafdennis:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `@lingui/macro@3.14.0` for the project I'm working on.

When we upgraded from Lingui 2.x to 3.x, we noticed our builds (babel-loader) were up to 2x slower and our tests (babel-jest) became flaky in our CI environment. We found this was caused by a piece of global state in `@lingui/macro` that is ever-growing and leaks path objects over time, resulting in longer build times and high memory pressure in big test suites. 

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@lingui/macro/index.js b/node_modules/@lingui/macro/index.js
index 1384958..bc98648 100644
--- a/node_modules/@lingui/macro/index.js
+++ b/node_modules/@lingui/macro/index.js
@@ -134,13 +134,14 @@ function isRootPath(allPath) {
   };
 }
 
-var alreadyVisitedCache = [];
+// Used to mark path objects as visited without relying on global state or creating memory leaks.
+var alreadyVisitedKey = Symbol('@lingui/macro#alreadyVisited');
 
 function alreadyVisited(path) {
-  if (alreadyVisitedCache.includes(path)) {
+  if (path.hasOwnProperty(alreadyVisitedKey)) {
     return true;
   } else {
-    alreadyVisitedCache.push(path);
+    path[alreadyVisitedKey] = true;
     return false;
   }
 }
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

semoal:
Could be better instead of using a Symbol, using a weakmap? Nice find mate, thanks for the contribution

decafdennis:
Yes. For some reason I assumed `Weak[Anything]` is not widely supported, but it is. CPU and memory usage appears to be equivalent.

If this looks good to you, I'd be happy to open a PR!

```diff
diff --git a/node_modules/@lingui/macro/index.js b/node_modules/@lingui/macro/index.js
index 1384958..0b834c5 100644
--- a/node_modules/@lingui/macro/index.js
+++ b/node_modules/@lingui/macro/index.js
@@ -134,13 +134,13 @@ function isRootPath(allPath) {
   };
 }
 
-var alreadyVisitedCache = [];
[...]

PR #1298, opened 2022-11-07
Improve @lingui/macro CPU and memory usage

decafdennis:
Fixes https://github.com/lingui/js-lingui/issues/1297


semoal:
Thanks for the contribution!! Really appreciated

PR #1306, opened 2022-11-30
feat: implement @lingui/vite-plugin

thekip:
Implementation of [Vite ](https://vitejs.dev/) Plugin based on Snowpack Plugin

1. Although webpack and probably snowpack loaders may work with extensions different from `.po` this plugin works only with `.po`. This is because we need a certain way to tell Vite to use proper loader for extension. Therefore files with `json` extension would be processed by Vite itself, and it's quite uncertain how to tell vite to use this loader for some subset of json files (we probably can play with query parameters in request, but I would prefer avoid it because `.po` is recommended way and everybody should use it. 
2. Naming, vite documentation https://vitejs.dev/guide/api-plugin.html#conventions clearly states that you should use `vite-plugin-` prefix, but in this project other plugins have own internal naming convention. I followed this convention instead of vite's. I left decision which one is right up to maintaners.


thekip:
Tests failing because vite is in dev deps which requires min node v16. Do we really need to support all this ancient node versions?

thekip:
by the way configuration of monorepo is also not well ergonomic, I would like to add "real" test for this plugin involving execution of vite itself, but it's not possible because ESM/CJS inconsistency and lack of ability to change jest config exactly for this subproject.

andrii-bodnar:
@thekip will it work on Node 14? 

Currently, I am upgrading the CI/CD (#1325) and Node 14 - now is the min version for tests (since this version LTS period going to end on [2023-04-30](https://nodejs.dev/en/about/releases/))

thekip:
I'm not sure actually. Vite probably requires node v16 so this will not work. 

I think the whole pipeline should be changed to be independent for each package.

andrii-bodnar:
It's quite a big change for the entire build and release process

andrii-bodnar:
@thekip could you please rebase the branch?

thekip:
@andrii-bodnar rebased

packages/vite-plugin/src/index.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export default function lingui(linguiConfig: LinguiConfigOpts = {}): Plugin {
+  return {
+    name: 'vite-plugin-lingui',
+
+    transform(src, id) {
+      const config = getConfig(linguiConfig);
+
+      if (fileRegex.test(id)) {
```
Martin005, line 21: We should throw error here, if the regex test fails. Similarly to how the Snowpack plugin does it.

thekip: If we throw here, it would throw in vite for every single file which is not a .po 

This statement is analog of webpack's `rule.test`, determine which modules should be processed by this transform. 

```js
         {
           test: /\.tsx?$/,
           loader: 'babel-loader',
           exclude: /node_modules/,
         },
```

Martin005:
> It's also should be registered in `scripts/build/bundles.` otherwise it would not be picked up by build system

Forgot to mention that, thanks!

andrii-bodnar:
@thekip is it possible to finalize and release this PR? 

Also, I would suggest adding the corresponding docs article within this PR so users would be able to easily find this info

thekip:
I would like to have [this PR](https://github.com/lingui/js-lingui/pull/1366) merged, because I want to use new `test` folder to add integration test with real Vite setup into that folder.

thekip:
I also noticed what could be improved in this plugin. This plugin can compile only `*.po` catalogs. 

The issue wouldn'be fixed by just changing a regexp to accepting `*.json` extension  because this will break applications in general. We could not distinguished catalogs from other json files in the project using only extension. 

So better solution would be take a catalog's regexp from lingui config and treat all files matching this regexp as catalogs no matter of extension. 

Probably snowpack plugin also has this issue, it treat all json files in bundle as catalogs and tries to compile all of them. 

I think for the first iteration we could merge it as-is (with `po` support only) and improve later on. Let me know what do you think.

packages/vite-plugin/README.md
```diff
@@ -0,0 +1,50 @@
...
+> https://vitejs.dev/guide/features.html#dynamic-import
+
+
+## License
+
+[MIT][license]
+
+[license]: https://github.com/lingui/js-lingui/blob/main/LICENSE
+[linguijs]: https://github.com/lingui/js-lingui
+[documentation]: https://lingui.js.org/
```
andrii-bodnar, line 46:
```suggestion
[documentation]: https://lingui.dev/
```

packages/vite-plugin/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@lingui/vite-plugin",
+  "version": "3.11.0",
```
andrii-bodnar, line 3: 3.17.0?

website/docs/ref/vite-plugin.md
```diff
@@ -0,0 +1,46 @@
+# Vite Plugin
+
+It's a good practice to use compiled message catalogs during development.
+However, running [`compile`](/docs/ref/cli.md#compile) everytime messages are changed soon becomes tedious.
```
andrii-bodnar, line 4: Please make it a single-line statement (it's a common style for all the documentation)

website/docs/ref/vite-plugin.md
```diff
@@ -0,0 +1,46 @@
...
+import { UserConfig } from 'vite';
+import lingui from '@lingui/vite-plugin'
+
+const config: UserConfig = {
+  plugins: [lingui()]
+}
+```
+
+Then in your code all you need is to use [dynamic imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#dynamic_imports)
+to load only necessary catalog. Extension is mandatory.
```
andrii-bodnar, line 30: also single-line

website/docs/ref/vite-plugin.md
```diff
@@ -0,0 +1,46 @@
...
+export async function dynamicActivate(locale: string) {
+   const { messages } = await import(`./locales/${locale}.po`);
+
+   i18n.load(locale, messages)
+   i18n.activate(locale)
+}
+```
+
+See the [guide about dynamic loading catalogs](/docs/guides/dynamic-loading-catalogs.md) for more info.
+See [Vite's official documentation](https://vitejs.dev/guide/features.html#dynamic-import) for more info about Vite dynamic imports.
```
andrii-bodnar, line 42: add a new line between these two lines since it will be "glued" into one line during the rendering

andrii-bodnar:
@thekip thank you. I think we can start from the `po` support only and then will see according to users' requests.

Regarding the `CHANGELOG.md` file - probably we should leave it empty as is and it will be updated during the release automatically.

packages/vite-plugin/package.json
thekip: 3.16.1 aligned with lerna.json

thekip:
done

andrii-bodnar:
@thekip one more thing regarding the `packages/cli/api.ts` rename - is it something intentional?

thekip:
Yes. With tsc it's not possible to use path mapping. If you noticed, the tsconfig in vite-plugin is not extended from the root one. Because in this case tsc generates structure choosing root as longest path. 

Example: 
./build/projects/vite-plugin/src/index.js 

So I was forced not to use tsconfig's path mapping and use yarn workspace linking instead. Renaming api.js to .ts was needed to make typings works and satisfy compiler

website/docs/ref/vite-plugin.md
```diff
@@ -0,0 +1,45 @@
+# Vite Plugin
```
andrii-bodnar, line 1: Just realized that the new page is not displaying in the left menu. Please add a reference in the `website/sidebars.ts` file

thekip: done

scripts/build/custom.ts
```diff
@@ -0,0 +1,34 @@
...
+      if (error) {
+        reject({stdout, stderr})
+        return
+      }
+      resolve({stdout, stderr})
+    })
+  )
+}
+
+export async function customBuilder(bundle: BundleDef) {
```
Martin005, line 21: Why not just use `export default async function` like all other builders?

thekip: not in my habit, default export/import is evil

thekip:
done

Martin005 (approved):

andrii-bodnar (approved):

andrii-bodnar:
Thank you @thekip, @Martin005! I'm going to release a new version later this week

Issue #1224, opened 2022-04-05
SWC plugin


renchap:
**Is your feature request related to a problem? Please describe.**
I want to switch our projects from Babel to [SWC](https://swc.rs), a Rust-based JS transpiler.
SWC is used by Next.js unless you opt-out from it because you rely on Babel.

**Describe proposed solution**
SWC has a brand new plugin system, which should be able to handle the transformations needed by Lingui's macros.

This is not yet stable / well documented, but some plugins are available here: https://github.com/swc-project/plugins (for Styled Components for example).

Those plugins are compiled to WASM, so they can be written in several languages (including [AssemblyScript](https://www.assemblyscript.org), a Typescript derivative compiling to WASM)

A first step would probably be to continue using Babel for extracting and use this plugin only to transform macros, but in the long term being able to use SWC for extracting could bring a much faster `extract` and heavily reduce dependencies needed for extraction (`swc` + `swc-lingui` and thats pretty much all).

**Describe alternatives you've considered**
- Porting Babel macros to a SWC plugin: not sure this can easily be done, this has been discussed a bit in https://github.com/kentcdodds/babel-plugin-macros/issues/144 but this would probably be a new project. Also `babel-macros` is frozen as the author [no longer uses it](https://github.com/kentcdodds/babel-plugin-macros/issues/177)
- Using another transpiling tool, like `esbuild`: there is no support for `esbuild` + Lingui either (and `esbuild` does not have a plugin support)
- Switching away from Lingui: I really like Lingui, I would prefer not :)
- Keep using Babel: `swc` (and other modern compilers) are much faster, it would really be better to be able to switch away!

**Additional context**

renchap:
This issue is getting a lot of 👍, dont close it!

kopax-polyconseil:
Best post here, swc, ts-jest, both aren't supported by lingui, and our code base growing, our unit test gets poor performances, and the question to remove lingui become more and more obvious.

This is a must for big projects, otherwise lingui will never stay long in any on them.

bojanbass:
Any update on this one?

semoal:
Nothing yet we can't do right now, there's no way we can replicate the macro's thing on swc (at least as far as I know, not a rust expert here) But of course, any help or info will be highly appreciated :)

kopax-polyconseil:
Our project which is used by millions of users now remove lingui translation for nothing, twice time lost. My advice, don't use macro and even more experimental lingui for your app.

[...]

Issue #1178, opened 2021-12-14
Lingui extract doesn't handle i18n._ strings


juanluispaz:
**Describe the bug**
Lingui extract ignores i18n._

**To Reproduce**

```
import { i18n } from "@lingui/core"

export default function App() {
   return i18n._('This should be translated!')
}
```

**Expected behavior**
The text should appear in the .po file

**Additional context**
I review the source code and I found the function [isI18nMethod](https://github.com/lingui/js-lingui/blob/5feb120e8c75d7b8c082d5c4b68185ef5dcc3ebc/packages/babel-plugin-extract-messages/src/index.ts#L124) defined at babel-plugin-extract-messages but it looks like is never used.

- jsLingui version `3.12.1`
- Babel version `7.16.0`
- Your Babel config (`.babelrc`):
`.babelrc
{ "presets": [ "@babel/preset-env", "@babel/preset-react", "@lingui/babel-preset-react", "@babel/preset-typescript" ] }
`
- Lingui config (`.linguirc`)
`.linguirc
{
"locales": ["en", "es", "de"],
"sourceLocale": "en",
"catalogs": [{
"path": "<rootDir>/src/app-react/locales/{locale}/translations",
"include": ["<rootDir>/src/app-react"],
"exclude": ["**/node_modules/**"]
}],
"format": "po-gettext",
"compileNamespace": "ts"
}
`

semoal:
Hola Juan,

To get extracted you have to use the t`` macro. In that way the `babel-plugin-extract-messages` can know which value has to extract, the i18n it's used to interpolate between locales, plurals etc..

Since v3, you don't need to use i18n you can just use directly the macro.

For example:

```jsx
import { t } from "@lingui/macro"

export default function App() {
   return t`This should be translated!`
}

```

juanluispaz:
Hi,

The macros are wonderful, but in the context where I'm using this it is not possible till we finish the migration and we can drop the restriction.

tannerlyons:
@juanluispaz , you can use this workaround:

```
import { i18n } from "@lingui/core"

export default function App() {
   return i18n._(/*i18n*/ "This should be translated!");
}
```

The extractor uses `/*i18n*/` as a marker for things to extract.

borjaMiralles:
I can confirm that this workaround works, thanks @tannerlyons

PR #1309, opened 2022-12-06
feat: allow extract to work with i18n._ calls not created from macro

mitchjhill:
This change updates babel-plugin-extract-messages so that it will pick up usages of the i18n._ method from @lingui/core directly.

This solves the issues mentioned in https://github.com/lingui/js-lingui/issues/1178 and https://github.com/lingui/js-lingui/issues/1224#issuecomment-1317265270 without needing the workaround listed in 1178 (prefixing the id with `/*i18n*/`).

The main driver for this is to allow use of Lingui without macros, so that sites running on Next.js can use the SWC compiler instead of babel.

Aware that this is effectively adding back extract behavior that was removed previously, but hopefully this isn't too big of a deal - let me know if you have any feedback.

### DIFF
--- a/packages/babel-plugin-extract-messages/src/index.ts
+++ b/packages/babel-plugin-extract-messages/src/index.ts
@@ -109,6 +109,28 @@ function extractStringContatentation(t, node, error): string {
   }
 }
 
+function extractCommentString(t, path, valuePath, valueObj): string {
+  if (t.isStringLiteral(valueObj)) {
+    // Comment is a single line string
+    return valueObj.value;
+  }
+
+  // Comment is a multi-line string.
+  const errorIfNotAString = path
+    .get(valuePath)
+    .buildCodeFrameError("Only strings are supported as comments.")
+
+  if (t.isBinaryExpression(valueObj)) {
+    return extractStringContatentation(
+      t,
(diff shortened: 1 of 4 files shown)


capaj:
@mitchjhill just wanted to say

thekip:
I would highly discourage you using this. Use macro whenever possible. If not possible, use https://formatjs.io/

PS Lingui already supports SWC https://github.com/lingui/swc-plugin

capaj:
@thekip I had no idea about the SWC plugin! That looks gr8. Will try that out when I touch i18n next time.

thekip:
why would you want to use lingui without a macro?

capaj:
I want to use it on backend. Backend is plain typescript. We don't even run babel and we don't want to run babel.
We're sending quite a few  push notifications from our BE. At the moment it's just maybe 70 strings, but it will be more in the future.

capaj:
I just tried running SWC and that does not work for us, because SWC handles decorators a bit differently.

thekip:
> I just tried running SWC and that does not work for us, because SWC handles decorators a bit differently.

Probably related: https://github.com/nestjs/graphql/issues/2568

PR #1311, opened 2022-12-12
feat: add plural rule into compiled catalogs

thekip:
## Description
This PR adds support for emiting plural rule (what `make-plural/plurals` is currently used for) into compiled catalog. 
This allows users to avoid installing and importing `make-plural/plurals` in theirs code and use only catalogs instead. 

Following official [documentation](https://lingui.js.org/guides/dynamic-loading-catalogs.html#final-i18n-loader-helper) preferred way to setup is: 

```typescript
import { i18n } from '@lingui/core';
import { en, cs } from 'make-plural/plurals'

i18n.loadLocaleData({
  en: { plurals: en },
  cs: { plurals: cs },
})

export async function dynamicActivate(locale: string) {
  const { messages } = await import(`./locales/${locale}/messages`)
  i18n.load(locale, messages)
  i18n.activate(locale)
}
```

Which is in fact will always load plurals functions for locales which are currently not used on the page. 

After this change, user would be able to write like this: 
```typescript

export async function dynamicActivate(locale: string) {
  const { messages, plurals } = await import(`./locales/${locale}/messages`)
  i18n.load(locale, messages)

  i18n.loadLocaleData(locale, {
    plurals
  })

  i18n.activate(locale)
}
```

## Notes

1. I'm taking the plural function from `make-plural/plurals` and serialize it into the string. Then when compiling catalogs into AST I convert stringified function into AST and inject to the output file. 

Compilation now produces: 

```typescript
module.exports = {
    messages: {"Hello": "Alohà"},
    plurals: (n, ord) => {
        return n == 1 ? "one" : "other"
    }
}
```

### DIFF
--- a/packages/cli/api.d.ts
+++ b/packages/cli/api.d.ts
@@ -1,0 +1,1 @@
+export * from "./src/api"
--- a/packages/cli/src/api/__snapshots__/compile.test.ts.snap
+++ b/packages/cli/src/api/__snapshots__/compile.test.ts.snap
@@ -1,25 +1,49 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`createCompiledCatalog nested message 1`] = `/*eslint-disable*/module.exports={messages:JSON.parse("{\\"nested\\":{\\"one\\":\\"Uno\\",\\"two\\":\\"Dos\\",\\"three\\":\\"Tres\\",\\"hello\\":[\\"Hola \\",[\\"name\\"]]}}")};`;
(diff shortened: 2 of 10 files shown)


packages/cli/api.d.ts
```diff
@@ -0,0 +1 @@
+export * from "./src/api"
```
thekip, line 1: I added this to help IDE and Typescript to see typings from sources. Otherwise everethyng imported from /api was `any`

thekip:
This one would have conflicts after merging https://github.com/lingui/js-lingui/pull/1328

thekip:
Done

Martin005 (changes requested):
These changes look really great!

I have added a few comments and it would be great to include this in the documentation, could you please add it there?

thekip:
I have to postpone it for a while, until we solve this https://github.com/lingui/js-lingui/issues/1339 

After this change catalog became non-serializable to json (because of the function inside) and my example nextjs integration would not work.

thekip:
I think this solution is better https://github.com/lingui/js-lingui/issues/1446

PR #1316, opened 2022-12-28
feat: Add translating event

hajaj-medigate:


andrii-bodnar:
Hi @hajaj-medigate, please provide more context and how this new event would be helpful. 

As I understand, there might be a situation when both the `missing` and `translating` events will be emitted so I actually don't understand the use case

hajaj-medigate:
I want to verify that every text was translated inside react component, to prevent `t` from being called before time.

Martin005 (changes requested):
Please rename the event from `translating` to `getting-translation` as the name currently is a bit misleading. Also, could you please add information about the event to the docs ([similarly to the `missing` event](https://github.com/donocode/js-lingui/blob/main/docs/ref/core.rst#missing))?

hajaj-medigate:
Thank you for the review. @Martin005 
Can you re-review it?
Thanks, Amos

docs/ref/core.rst
```diff
@@ -402,3 +402,15 @@ Information about the locale and message are available from the event.
    i18n.on('missing', (event) => {
       alert(`alert(`Translation in ${event.locale} for ${event.id} is missing!`)`)
    })
+
+getting-translation
+-------
+
+Triggered when a translation is requested with ``i18n._``, regardless the message exists or not in the active locale's messages.
```
Martin005, line 409: @hajaj-medigate This got me thinking: it doesn't make sense to get a translation for something that is missing. Wouldn't it be better if the event was emitted only if it exists in the active locale's meesages?
Meaning the `this.emit(...)` call in `packages/core/src/i18n.ts` would be surrounded by an `else` statement.

hajaj-medigate: No, because `getting-translation` is called when a translation is requested, it doesn't matter if it missing or not.

@Martin005 WDYT?

Martin005: @hajaj-medigate But if the translation is missing there is no point of "getting" it, right? So from my point of view, it makes sense if the `getting-translation` was called only when the translation is not missing.

hajaj-medigate: My use case is more like `requesting-translation` rather than `getting-translation`
should I rename it?

@Martin005

hajaj-medigate: @Martin005 @andrii-bodnar  - any thoughts?

andrii-bodnar: @hajaj-medigate actually, I have some doubts regarding adding a new event and how it can be useful for other Lingui users

thekip:
@hajaj-medigate 
> I want to verify that every text was translated inside react component, to prevent t from being called before time.

Could you show what you want on a real example? if you want to be sure that all messages are translated in the catalog,  you can do that in the compile-time, not a runtime.

PR #1325, opened 2023-01-06
ci: Upgrade CI/CD

andrii-bodnar:


Martin005:
@andrii-bodnar Maybe we should also remove `[skip ci]` from [lerna.json](https://github.com/lingui/js-lingui/blob/main/lerna.json#L10)?

Martin005 (approved):

Issue #1278, opened 2022-08-18
i18n._ function removes unrecognized macros


nmoinvaz:
**Describe the bug**
`i18n._` removes macros that are unrecognized. This prevents programs that also parse macros in strings from handling them after Lingui loads them.

**To Reproduce**
Using the following code:
```
i18n._("code {hello}")
```
The string returned is `code `.

**Expected behavior**
It should return `code {hello}` because no `values` object was passed to `i18n._`.

**Alternatives ideas**
1. Perhaps there is a callback that can be triggered when it encounters a macro it doesn't recognize?
2. Or an option that can be enabled for this behavior?
3. Or another function that does not process macros but can get the locale string.

nmoinvaz:
Worse case, it looks like according to PEG, it should be possible using a literal to bypass parsing.
```
i18n._("code '{hello}'")
```

thekip:
Fixed, starting from next version 

```ts
i18n._("code '{hello}'")
```
Should work

Issue #1033, opened 2021-04-08
Vietnamese support | bug


seeden:
From unknown reason lingui is not able to compile few Vietnamese strings.
It is working fine when I remove it from my .po file.
I am using latest version 3.8.2

Here is content of .po file
```po
#: src/components/farm/FarmYourHarvesterNetwork.tsx:58
msgid "A harvester is a service running on a machine where plot(s) are actually stored. A farmer and harvester talk to a full node to see the state of the chain. View your network of connected harvesters below Learn more"
msgstr "Harvester là một dịch vụ chạy trên một  máy chứa các tệp plot  . Một \bfarmer và harvester tương tác với một nút đầy đủ để xác định trạng thái của chuỗi. Xem các harvester được kết nối của bạn dưới đây Tìm hiểu thêm"


#: src/electron/main.ts:296
msgid "Contribute on GitHub"
msgstr "Đóng góp qua GitHub"
```

I see next error stacktrace:
```sh
/Users/seeden/Documents/git/chia-blockchain/chia-blockchain-gui/node_modules/@lingui/cli/api/compile.js:89
    throw new Error("Can't parse message. Please check correct syntax: \"".concat(message, "\""));
    ^

Error: Can't parse message. Please check correct syntax: "Đóng góp qua GitHub"
    at compile (/Users/seeden/Documents/git/chia-blockchain/chia-blockchain-gui/node_modules/@lingui/cli/api/compile.js:89:11)
    at /Users/seeden/Documents/git/chia-blockchain/chia-blockchain-gui/node_modules/@lingui/cli/api/compile.js:45:51
    at Array.map (<anonymous>)
    at createCompiledCatalog (/Users/seeden/Documents/git/chia-blockchain/chia-blockchain-gui/node_modules/@lingui/cli/api/compile.js:37:43)
    at /Users/seeden/Documents/git/chia-blockchain/chia-blockchain-gui/node_modules/@lingui/cli/lingui-compile.js:91:66
    at Array.forEach (<anonymous>)
    at /Users/seeden/Documents/git/chia-blockchain/chia-blockchain-gui/node_modules/@lingui/cli/lingui-compile.js:61:14
    at Array.forEach (<anonymous>)
    at command (/Users/seeden/Documents/git/chia-blockchain/chia-blockchain-gui/node_modules/@lingui/cli/lingui-compile.js:50:18)
[...]

Issue #1303, opened 2022-11-22
Bug: brackets in template literal results in macro compile failure


bright-wang:
**Describe the bug**
Just very similar to https://github.com/lingui/js-lingui/issues/1075, I also came up with compile failure error:
Error: Can't parse message. Please check correct syntax: "input variable: { ...... }" 
 
 Messageformat-parser trace: Expected identifier but "." found.
    at compile (....../node_modules/.pnpm/@lingui+cli@3.13.3_l57gh7u5ywadp5ivckxo6hsz4m/node_modules/@lingui/cli/api/compile.js:115:11)
    at compileSingleKey ....../node_modules/.pnpm/@lingui+cli@3.13.3_l57gh7u5ywadp5ivckxo6hsz4m/node_modules/@lingui/cli/api/compile.js:42:49)
    at ....../node_modules/.pnpm/@lingui+cli@3.13.3_l57gh7u5ywadp5ivckxo6hsz4m/node_modules/@lingui/cli/api/compile.js:70:12

**To Reproduce**
Steps to reproduce the behavior, possibly with minimal code sample, e.g:

```
import { t } from '@lingui/macro';

export default function App() {
   return <input placeholder={t`input variable: { ...... }`} />
}
```

**Expected behavior**
Template should be compiled with no error...

**Additional context**
Package version info:
 '@lingui/cli': 3.13.3_l57gh7u5ywadp5ivckxo6hsz4m
  '@lingui/macro': 3.13.3_fxhud3nu6zrc5fvtlhahzizpp4

thekip:
Fixed and would be published with next version. You probably need to escape brackets with `'` otherwise they would be treated as placeholder expansion (this is according of spec) 

```
```

Issue #1075, opened 2021-05-27
if the message has curly brackets in itself, compilation is broken


capaj:
**Describe the bug**
if the message has curly brackets in itself, compilation is broken

**To Reproduce**

```
import { t } from "@lingui/macro"

t`curly brackets { } in a message break compilation.`
```
extract and try to compile

**Expected behavior**
compilation works fine
**Additional context**
message format

```
/home/capaj/work-repos/looop/project-alpha/node_modules/@lingui/cli/api/compile.js:89
    throw new Error("Can't parse message. Please check correct syntax: \"".concat(message, "\" \n \n Messageformat-parser trace: ").concat(e.message));
    ^

Error: Can't parse message. Please check correct syntax: "The strings contains invalid variable definition.Please ensure the number of variables matches the number used in the original string. A variable is enclosed in curly brackets { }." 
 
 Messageformat-parser trace: Expected identifier but "}" found.
    at compile (/home/capaj/work-repos/looop/project-alpha/node_modules/@lingui/cli/api/compile.js:89:11)
    at /home/capaj/work-repos/looop/project-alpha/node_modules/@lingui/cli/api/compile.js:45:51
    at Array.map (<anonymous>)
    at createCompiledCatalog (/home/capaj/work-repos/looop/project-alpha/node_modules/@lingui/cli/api/compile.js:37:43)
    at /home/capaj/work-repos/looop/project-alpha/node_modules/@lingui/cli/lingui-compile.js:91:66
    at Array.forEach (<anonymous>)
    at /home/capaj/work-repos/looop/project-alpha/node_modules/@lingui/cli/lingui-compile.js:61:14
    at Array.forEach (<anonymous>)
    at command (/home/capaj/work-repos/looop/project-alpha/node_modules/@lingui/cli/lingui-compile.js:50:18)
    at compile (/home/capaj/work-repos/looop/project-alpha/node_modules/@lingui/cli/lingui-compile.js:135:12)
```

- jsLingui version ``lingui --version`` 3.8.10
- Babel version ``npm list @babel/core`` 7.14.0
- Your Babel config (e.g. `.babelrc`) not relevant

semoal:
Yeah this is an error itself from the `message-format` parser dependency we have in the cli, there's also an issue with Vietnamese support which is fixed in the new package. #1033 

I'll try to find time as soon as possible to migrate to this new parser which probably fixes all this situations where we don't have any control :/

PR #1328, opened 2023-01-09
refactor(core): bump messageformat parser. Add TS types

thekip:
Bumped version of the upstream messageformat parser to 5.0, updated what's broke and add TS typings wherever possible in touched files. 

Resolves 
https://github.com/lingui/js-lingui/issues/1278
https://github.com/lingui/js-lingui/issues/1303

Should be no changes in logic, but public types slightly changed. They became more accurate, so it's better to bump a version just in case. 

How current config of semver works, should i name a PR with special prefix or commit message or it's doing manually?


andrii-bodnar:
@thekip thanks for the contribution!

Probably, I'll ask @Martin005 to review this PR.

> How current config of semver works, should i name a PR with special prefix or commit message or it's doing manually?

It expects that the PR title will be specified according to the following format:

https://github.com/amannn/action-semantic-pull-request#validation

Martin005:
@thekip You made changes only to the `@lingui/core` package.
Could you please also update the `@lingui/cli` and `@lingui/remote-loader`? They currently also use the `messageformat-parser` package.

Also, updating the package may also fix issues #1033 and #1075 🙂

thekip:
My bad, didn't check this. 

This commit is bigger, few things about what changed: 

1. I've found a code duplication (3 times) where tokens from `message-parser` were converted to Lingui internal format. 
I removed this code duplication by exporting from `@lingui/core` a `parseMessage` function and reusing it in all subpackages. 

2. I simplified the way how the compiled catalogs produced. This was for 3 purposes: 
   1. to reuse mentioned function
   2. to make code much cleaner and smaller
   3. to make reading of catalogs actually faster than before. 

Before compiled catalogs were produced as js template literal (js code) and complicated AST manipulation was involved
```js
export const messages = {
  messageId: "message value"
}
```
Now i'm just using JSON.stringify / JSON.parse
```js
export const messages = JSON.parse("{
 "messageId": "message value"
}")
```
Why it's better you can read here https://v8.dev/blog/cost-of-javascript-2019#json

packages/core/src/i18n.ts
```diff
@@ -7,23 +7,26 @@ import { EventEmitter } from "./eventEmitter"
...
 
 export type Locale = string
 export type Locales = Locale | Locale[]
+export type Formats = Record<string, Intl.DateTimeFormatOptions | Intl.NumberFormatOptions>
+
+export type Values = Record<string, unknown>;
 
 export type LocaleData = {
-  plurals?: Function
+  plurals?: (n: number, ordinal?: boolean) => number
```
Martin005, line 20: The return type needs to be `PluralCategory` from the `make-plural` instead of `number`.

Martin005 (changes requested):
Thank you for the changes, I think that all of what you mentioned makes perfect sense including the JSON.stringify/JSON.parse! 🙂

When I tried your branch as a local package, I immediately noticed one mistake, so **please have a look at my comments and resolve them**.

When I resolved the mistake locally, I also tried all 4 issues #1033, #1075, #1278 and #1303 and I can confirm that all of them are fixed with this new version of `@messageformat/parser` package 🥳

thekip:
Fixed

thekip:
I also thinking on how to hide exposed but treated as internal APIs, for example that `formatMessage` is supposed to be internal api, but because we use it in other packages it should be exported. 

There should be either special naming convention, or may be just JSdoc would be enough

thekip:
I see tests are failing, i will have a look

thekip:
I've ported some changes from this PR https://github.com/lingui/js-lingui/pull/1328

> I deleted pure option from createCompiledCatalog because it's not used anywhere in the project. This API considered internal, so we should be on the safe side.
It's probably was created for remote-loader but code has been changed since then and now this option is obsolete.

And also i've ported tests, because previous ones didn't test anything (there was a logical issue)

Martin005 (approved):
Great, thank you for the changes!

PR #1329, opened 2023-01-09
fix(conf): proper typings for LinguiConfig['runtimeConfigModule']

thekip:
https://lingui.js.org/ref/conf.html#id12

runtimeConfigModule should accepts or tuple or object with properties, this PR fixing typings to reflect this


Martin005 (changes requested):
@thekip Please rename the `what` parameter to `identifier`

Martin005 (approved):

PR #1333, opened 2023-01-12
ci: fix the 'size-limit' workflow

andrii-bodnar:


.github/workflows/size-limit.yml
```diff
@@ -1,10 +1,6 @@
 name: Size Testing
 
-on:
-  pull_request_target:
-    types:
-      - opened
-      - synchronize
+on: [pull_request]
```
Martin005, line 3: Isn't `on: pull_request` enough? The array should be used for multiple events [according to docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#using-multiple-events).

.github/workflows/size-limit.yml
```diff
@@ -1,10 +1,6 @@
 name: Size Testing
 
-on:
-  pull_request_target:
-    types:
-      - opened
-      - synchronize
```
Martin005, line 7: Why have you removed these types?

andrii-bodnar: Just updated it to the same trigger as in the `main-suite` workflow. Probably it should be enough, also it will run on each PR update

PR #1336, opened 2023-01-12
refactor(build): typescript + support inner exports

thekip:
Fix size-limit issue 
And fix bundle sizr issue introduced by @message-format/parser exported from @lingui/core

Refactor and rewrite to typescript build scripts


thekip:
Still failing, that strange

thekip:
It worked with flat exports, and started failing again because of multi-entry, probable i've created inner entry incorrectly

thekip:
🎉 builds!  Size testing failing because it's trying to compare with master and master has a failing code now.

Martin005:
@thekip
I have encountered several issues with your implementation:

1. After I built the library and released it locally with the `/scripts/verdaccio-release.js` script, I tried installing it in the current example `create-react-app` and running with `yarn start`, but I got the following error:
```
node:internal/modules/cjs/loader:535
      throw e;
      ^

Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /home/martin/js-lingui/examples/create-react-app/node_modules/@lingui/conf/package.json
```

2. Following the [CONTRIBUTING.md](https://github.com/lingui/js-lingui/blob/main/CONTRIBUTING.md#contributing-the-code):

- When I run `release:test`, I get the following error:
```
node:internal/modules/cjs/loader:998
  throw err;
  ^

Error: Cannot find module '/home/martin/js-lingui/scripts/build'
```

- Also, when I run `tsc` (as a part of the `lint:all` command in the main `package.json`), I get the following error:

```
$ /home/martin/js-lingui/node_modules/.bin/tsc
packages/core/build/cjs/dev/index.d.ts:1:10 - error TS2305: Module '"../compile/compileMessage"' has no exported member 'compileMessage'.

1 import { compileMessage as compile } from "../compile/compileMessage";
           ~~~~~~~~~~~~~~

packages/core/build/esm/dev/index.d.ts:1:10 - error TS2305: Module '"../compile/compileMessage"' has no exported member 'compileMessage'.

1 import { compileMessage as compile } from "../compile/compileMessage";
           ~~~~~~~~~~~~~~
```

andrii-bodnar:
We'd like to have a separate fix for the bug in `main` and separately the rest (just to not block the new version release)

thekip:
> We'd like to have a separate fix for the bug in main and separately the rest (just to not block the new version release)
it's not possible to ship it separately, to fix the `main` building system should be changed.
You can disable or ignore size-limit issue for now and ship without it, code should work, it's just an issue with workflow in the monorepo. 

> Also, when I run tsc (as a part of the lint:all command in the main package.json), I get the following error:
Why it's not failing in CI?

andrii-bodnar:
@thekip what exactly went wrong in the #1328 that broke this size-check? It might be easier to rollback this change than dangerously change the build system.

It worked before and it's not a good idea to ignore this.

thekip:
> so in short
this is a first time usage of the cross-linked library in client-side code
There are also lingui/react which states that uses core package, but actually it uses only typings, so in real js code imports to core are stripped
and currently when webpack starts resolving imports it goes to ./packages/core/ instead of ./package/core/dst because this is how yarn workspace linking works
When we deploy packages, it deployed from ./dst
but when thy used locally, they used from ./ {root of package}
it means  the changes should be bigger to support this
i will try to find a quick workaround for that

thekip:
In my opinion it's better to "dangerously" change the build system (which in fact is not dangerously)

thekip:
`release:test` and `lint:all` are fixed. 

Actually `release:test` are useless today, it's doing the same thing as in CI but with diffrent code. So issue was in file itself not in tests. 

Typings was failing because i marked `compileMessage` as `@internal`, and typescript stripped the typings of it. (that's exactly what i wanted to achieve). 

I deleted /dev folder with all content and reorganized it in different way. Probably previously (prior v3)  the `dev` was somehow exposed from the package, but right now in current `main` code it's no longer exposed, and users have no chance to use anything from there, so it's safe. 


Looking into issue with No "exports" with @lingui/conf

thekip:
Fixed all issues, now it works in examples. But came up with better (less fragile) idea, i will try to implement it and look what i get

thekip:
@Martin005 @andrii-bodnar was managed to simplify process and remove fragile parts. If you're afraid of changes, we can discuss how the process previously looks like, and how it's looks now. 

In short: 

We do not copy package.json from root of package to ./build, instead we publish package  from the package root. This is how it's done in most of the projects on the market. This simplifies things a lot and fix issues with size-check among with others.

scripts/build/packaging.ts
```diff
@@ -0,0 +1,49 @@
+import {existsSync, readdirSync, unlinkSync} from "fs"
+import {BundleType as BundleType} from "./bundles"
+
+import {
+  asyncCopyTo,
+} from "./utils"
+
+export function getBundleOutputPaths(bundleType: BundleType, filename: string, packageName: string) {
+  const _filename = filename.replace(/^@lingui\//, "")
```
Martin005, line 9: Isn't this replace now outdated as you have removed the `@lingui` prefix from all package names in the `bundles.ts` file?

scripts/build/packaging.ts
```diff
@@ -0,0 +1,49 @@
+import {existsSync, readdirSync, unlinkSync} from "fs"
```
Martin005, line 1: `unlinkSync` is now not used and can be removed

scripts/build/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+import * as path from "path"
+import ncp from "ncp"
+import rimraf from "rimraf"
+
+export async function asyncCopyTo(from: string, to: string) {
+  if (!fs.existsSync(from)) return
+
+  await asyncMkDirP(path.dirname(to))
+
+  return new Promise((resolve, reject) => {
```
Martin005, line 12:
```suggestion
  return new Promise<void>((resolve, reject) => {
```

scripts/build/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+
+
+export function asyncMkDirP(filepath: string) {
+  return fsPromises.mkdir(filepath, {
+    recursive: true
+  })
+}
+
+export function asyncRimRaf(filepath: string) {
+  return new Promise((resolve, reject) =>
```
Martin005, line 33:
```suggestion
  return new Promise<void>((resolve, reject) =>
```

scripts/build/rollup.ts
```diff
@@ -320,11 +319,13 @@ async function build(bundle, bundleType) {
...
 }
 
-module.exports = async function (bundle) {
-  await build(bundle, NODE_DEV)
-  await build(bundle, NODE_PROD)
-  await build(bundle, ESM_DEV)
-  await build(bundle, ESM_PROD)
+export default async function (bundle: BundleDef) {
+  await build(bundle, RollupBundleType.NODE_DEV)
+
```
Martin005, line 324: This empty line can be removed

Martin005 (changes requested):

Martin005 (approved):

andrii-bodnar (approved):

PR #1337, opened 2023-01-13
docs: Fix syntax of extracted comments in PO files

Martin005:
This PR fixes docs, particularly syntax of extracted comments in PO files in the Catalog Formats section. It was incorrectly showing that comments for translators would be prefixed with `# ` instead of `#. `
Also, I have visually broken up a section in the same file to be more legible.

### DIFF
--- a/docs/ref/catalog-formats.rst
+++ b/docs/ref/catalog-formats.rst
@@ -25,7 +25,7 @@ The advantages of this format are:
 .. code-block:: po
 
    #: src/App.js:3
-   #  Comment for translators
+   #. Comment for translators
    msgid "messageId"
    msgstr "Translated Message"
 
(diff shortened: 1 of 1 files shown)


docs/ref/catalog-formats.rst
```diff
@@ -53,28 +53,32 @@ This is how the regular PO format exports plurals:
 
 With `po-gettext`, plural messages are exported in the following way, depending on wheter an explicit ID is set:
 
-.. code-block:: po
+1. Message **with custom ID "my_message"** that is pluralized on property "someCount".
+
+   Notice that 'msgid_plural' was generad by appending a '_plural' suffix.
```
andrii-bodnar, line 58:
```suggestion
   Notice that 'msgid_plural' was generated by appending a '_plural' suffix.
```

Martin005: @andrii-bodnar Thanks, I only moved the sections and didn't check the text contents 🙂

andrii-bodnar (approved):

PR #1338, opened 2023-01-13
chore(size-limit): test esm packages and update size-limit dep

thekip:
All bundlers actually consume ESM version of packages. CJS version is usually used by nodejs. So it's make more sense to test ESM versions for the size because it will bundled and shipped to the browser. 

I had to rise the limit for `@lingui/core package` to 5.5kb. It's not related to CJS/ESM change, it's related to `size-limit` dependecy update. It uses fresher bundlers which generate a slightly bigger results for both targets (CJS/ESM) i've tested it both with esbuild and webpack bundlers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "@babel/preset-env": "^7.11.5",
     "@babel/preset-react": "^7.10.4",
     "@babel/preset-typescript": "^7.10.4",
-    "@size-limit/preset-small-lib": "^4.10.1",
+    "@size-limit/preset-small-lib": "^8.1.1",
     "@swc/jest": "^0.2.24",
     "@swc/core": "^1.3.26",
     "swc-node": "^1.0.0",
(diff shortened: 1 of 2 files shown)


thekip:
@Martin005 do you really understand how this package works? Please read the docs, and improve your understanding what exactly size-limit doing and where exactly this 5kb appeared.

andrii-bodnar:
@thekip please briefly explain it here

thekip:
This change did not increase actual size of the package. This affects only reporting of the size by `size-limit` lib. It becomes more accurate because now uses fresher versions of bundlers to calculate a size.

thekip:
Rebased to main and add "react" as "external". Because we assume every app which going to use @lingui/react would already have react in the bundle, so we should not take it into account in budget calculation.

Martin005 (approved):

andrii-bodnar (approved):

Issue #1324, opened 2023-01-06
Select Macro Documentation Example is wrong


thekip:
Example from [documentation]( https://lingui.js.org/ref/macro.html#id3): 
```ts
import { Select } from "@lingui/macro"

// gender == "female"      -> Her book
// gender == "male"        -> His book
// gender == "unspecified" -> Their book
<Select
    value={gender}
    male="His book"
    female="Her book"
    other="Their book"
/>
```

Actually, it should be: 
```ts
import { Select } from "@lingui/macro"

<Select
    value={gender}
    _male="His book"
    _female="Her book"
    other="Their book"
/>
```

Proof: https://github.com/lingui/js-lingui/blob/main/packages/macro/test/jsx-select.ts#L3-L19

If you write code as in example, macro will produce: 

```diff
import { Trans } from "@lingui/react";
<Trans
+  male="He"
+  female={`She`}
  id={"{count, select, male {He} female {She} other {<0>Other</0>}}"}
  values={{
    count: count,
  }}
  components={{
    0: <strong />,
  }}
/>;

```

Which by the way will work, but unnecessary code would be produced.

So here we should either fix the macro to not produce this fields and add this to the test case. 
Or update docs and fix usage with `_{option}` which would be a breaking change. 

Anyway, action point needed here.

thekip:
Looked at the code one more time, and looks like this work by mistake. The original author's intention was to have exact choices with underscore only.  Here is the regex:

`/(_[\d\w]+|zero|one|two|few|many|other)/`

thekip:
Related https://github.com/lingui/js-lingui/issues/1288

tricoder42:
> Looked at the code one more time, and looks like this work by mistake. The original author's intention was to have exact choices with underscore only. Here is the regex:
> 
> `/(_[\d\w]+|zero|one|two|few|many|other)/`

You're right. ICU MessageFormat uses `=` for exact numbers in plural-like formats (e.g. `=0`). I replaced `=` with `_` since you can't use `=` character in prop names. I believe I kept the same naming convention in select format a) to keep it consistent b) to avoid name clashes with macro props.

Issue #1323, opened 2023-01-06
Mess in exported Macro Typescript Typings


thekip:
I'm continuing digging into the sourcecode and test suite. Here what i've found: 

According to the exported [typings](https://github.com/lingui/js-lingui/blob/main/packages/macro/index.d.ts) all JSX Macro components (Trans, Plural, etc) have common `Trans` props, which is following: 

```ts
export type TransProps = {
  id?: string
  comment?: string
  values?: Record<string, unknown>
  context?: string
  children?: React.ReactNode
  component?: React.ComponentType<TransRenderProps>
  render?: (props: TransRenderProps) => ReactElement<any, any> | null
  i18n?: I18n
}

export type ChoiceProps = {
  value?: string | number
} & TransProps &
  ChoiceOptions<ReactNode>

export type SelectProps = {
  value: string
  other: ReactNode
} & TransProps & Values

export const Trans: ComponentType<TransProps>
export const Plural: ComponentType<ChoiceProps>
export const Select: ComponentType<SelectProps>
export const SelectOrdinal: ComponentType<ChoiceProps>
```

This is actually wrong, because JSX Macro (from the source code and tests) could accept only this set properties:

```diff
export type TransProps = {
  id?: string
  comment?: string
-  values?: Record<string, unknown>
  context?: string
-  children?: React.ReactNode // should be only for Trans, not by any other Choice components
-  component?: React.ComponentType<TransRenderProps>
  render?: (props: TransRenderProps) => ReactElement<any, any> | null
  i18n?: I18n // this one is not documented any where, we should add it to the docs
}
```

So passing any of the properties highlihted in the snippet above will not work. These typings just bring mess and confusing for end users. (it's actually props of original `Trans` runtime component, not a macro).

Proofs: https://github.com/lingui/js-lingui/blob/main/packages/macro/src/macroJsx.ts#L179-L194
Other properties would stripped or inlined as choice options in ICU Choice component. 

Am i missing something or it's here by historical reasons and a just need to be cleaned up wisely?

thekip:
Related: https://github.com/lingui/js-lingui/issues/1288

`Select` typings are wrong.

tricoder42:
> Am i missing something or it's here by historical reasons and a just need to be cleaned up wisely?

No, you're right. `Trans` macro has fewer props that `Trans` component. `values`, `component` are generated by macro itself, so you should use them explicitly.

```jsx
// Macro
[...]

Issue #1288, opened 2022-10-03
Ability to use elements within Select values


ckknight:
TL;DR: `<Select>` should either support non-string values or be extremely explicit in the documentation about why they can't/shouldn't be used.

I was attempting to use `<Select>` earlier and for one of the values, I realized I needed to use a nested element (and thus not just rely on a string).

I had attempted to use the following syntax:

```tsx
<Select
  value="happy"
  happy={
    <Trans>
      Hooray! <Icon />
    </Trans>
  }
  sad={
    <>
      Oh no! <Icon />
    </>
  }
  other="Dunno"
/>
```

This has three immediate issues that I can see
1. TypeScript expects _only_ `string` properties here
2. `lingui extract` complains with a message like
    ``` 
    Missing message ID, skipping.
    <Trans __self={this} __source={{
      fileName: _jsxFileName,
      lineNumber: 115,
      columnNumber: 17
    }}>
    ```
3. This error occurs at runtime: Duplicate __self prop found. You are most likely using the deprecated transform-react-jsx-self Babel plugin. Both __source and __self are automatically set when using the automatic runtime. Please remove transform-react-jsx-source and transform-react-jsx-self from your Babel config.

So it seems that `Trans`-inside-`Select` isn't great, even though the `.po` files do look _mostly_ accurate. (`msgstr "{0, select, happy { Hooray! <0/>} sad {{1}} other {Dunno}}"`)

Upon trying without the nested syntax,

```tsx
<Select
  value="happy"
  happy={
    <>
      Hooray! <Icon />
    </>
  }
  sad={
    <>
      Oh no! <Icon />
    </>
  }
  other="Dunno"
/>
```

This does appear to work at runtime, but "Hooray!" and "Oh no!" become untranslatable. The TypeScript issue is still a problem.

The following _does_ seem to be the closest I've come to actually working:
```tsx
<Trans>
  <Select
    value="happy"
    happy={
      <Trans>
        Hooray! <Icon />
      </Trans>
    }
    sad={
      <Trans>
        Oh no! <Icon />
      </Trans>
    }
    other="Dunno"
  />
</Trans>
```

This appears to do what I intend to do at runtime.

In the po file, the entry looks like
```po
msgstr "{0, select, happy { Hooray! <0/>} sad { Oh no! <1/>} other {Dunno}}"
```

Given the extra spaces, it appears that if one puts a parenthesis or some character at the start of the translation, before the `<Select>`, an unexpected whitespace character slips in.

[...]

PR #1340, opened 2023-01-15
chore(macro): lint macro typings

thekip:
This PR is summarizing of various issues found in current public typings of macro.

The goal of this PR to show how many issues has current typings and give a starting benchmark to fix them. Another goal is to show to maintainers that public typings is also something that could be and should be automatically checked and validated in CI/CD.

Once all issues would be fixed, this validation should be added to CI/CD workflow. 

The feedback of this PR i want to get is "green lite" from maintainers to update public typings and fix these issues: 

https://github.com/lingui/js-lingui/issues/1324
https://github.com/lingui/js-lingui/issues/1323
https://github.com/lingui/js-lingui/issues/1288

Changing public typings would definitely affect some users because they become stricter and would catch more misusages and possible errors. So this change should be published in next major release.

How to use this PR: 

Clone the source code and open it in your favourite editor with TS support, you will see all errors. Read a comment next to it to understand the case. 

There is also an automatic tool which you can call from macro's package root via: 

```bash
yarn test:tsd
```


thekip:
@Martin005 @andrii-bodnar  Guys i need your reaction & feedback

Martin005:
@thekip
Thank you for pointing out the problem and showing how we can prevent erroneous typings!
I agree with you that this should be fixed. And after all issues are fixed, we should add the validation to CI/CD workflow.

andrii-bodnar:
We'd like to postpone any breaking changes for a while, but your suggestion is fair

packages/macro/index.d.ts
```diff
@@ -163,46 +188,108 @@ export function select(value: string, choices: ChoiceOptions): string
...
+  children: ReactNode
+} & CommonProps
+
+type PluralChoiceProps = {
+  value: string | number
+  /** Offset of value when calculating plural forms */
+  offset?: number
+  zero?: ReactNode
+  one?: ReactNode
+  few?: ReactNode
```
Martin005, line 213: Why is there the case `two` missing here?

packages/macro/index.test-d.tsx
```diff
@@ -0,0 +1,334 @@
...
+    context={"Context"}
+    id={"custom.id"}
+    one={"..."}
+    other={"..."}
+  />
+)
+
+// todo: write tests for
+//  render prop
+//  i18n
```
Martin005, line 287: Are you going to add these tests?

packages/macro/index.test-d.tsx
```diff
@@ -0,0 +1,334 @@
...
+    Message
+  </Trans>
+)
+
+// @ts-expect-error: children is required here
+m = <Trans id="custom.id" comment="comment" context="context" />
+
+// todo: write tests for
+//  render prop
+//  i18n
```
Martin005, line 241: Are you going to add these tests?

packages/macro/index.test-d.tsx
```diff
@@ -0,0 +1,334 @@
...
+    Message
+  </Plural>
+)
+
+// @ts-expect-error: value is required
+m = <Plural />
+
+m = <Plural value={5} offset={1} one={"..."} other={"..."} _0="" _1={"..."} />
+
+//  @ts-expect-error: offset could be string only
```
Martin005, line 258:
```suggestion
//  @ts-expect-error: offset could be number only
```

packages/macro/index.test-d.tsx
```diff
@@ -0,0 +1,334 @@
...
+///////////////////
+
+m = <Trans>Message</Trans>
+m = (
+  <Trans id="custom.id" comment="comment" context="context">
+    Message
+  </Trans>
+)
+
+// @ts-expect-error: children is required here
```
Martin005, line 236:
```suggestion
// @ts-expect-error: children are required here
```

packages/macro/index.test-d.tsx
```diff
@@ -0,0 +1,334 @@
...
+    _0=""
+    _1={"..."}
+  />
+)
+
+///////////////////
+//// JSX Select
+///////////////////
+m = (
+  // @ts-expect-error: children is not allowed here
```
Martin005, line 307:
```suggestion
  // @ts-expect-error: children are not allowed here
```

packages/macro/index.test-d.tsx
thekip: https://github.com/lingui/js-lingui/issues/1433 i created an issue for that (task number 2)

packages/macro/index.d.ts
thekip: Good point, it wasn't there in the first place i discovered it while comparing to the docs.

packages/macro/index.test-d.tsx
```diff
@@ -0,0 +1,334 @@
...
+    male: "he",
+    female: "she",
+    other: "they",
+  })
+)
+
+expectType<string>(
+  // @ts-expect-error value could be strings only
+  select(5, {
+    male: "he",
```
thekip, line 207: i will add `test-d.ts` files into coverage ignore pattern

thekip: done

packages/macro/index.test-d.tsx
thekip: removed todo

Martin005 (approved):

andrii-bodnar (approved):

PR #1341, opened 2023-01-16
fix(cli): catalogsMergePath doesn't merge catalogs

### LINKED ISSUE #1233: catalogsMergePath doesn't merge all the catalogs
**Describe the bug**
catalogsMergePath doesn't merge all the catalogs

**To Reproduce**
Define multiple catalogs in linguiRc and specify catalogsMergePath 

**Expected behavior**
The catalogs must be merge to a single .po file

**Additional context**

https://github.com/lingui/js-lingui/blob/main/packages/cli/src/lingui-compile.ts#L83
image

- jsLingui version ``3.13.2``
- Babel version `7.17.5`

Martin005:
This PR fixes #1233.
Previously, the catalogs were not merge even if `catalogsMergePath` option was provided in the configuration. The `getCatalogForMerge()` function was never called from the code (it was used only in tests).

### DIFF
--- a/packages/cli/src/lingui-compile.ts
+++ b/packages/cli/src/lingui-compile.ts
@@ -116,12 +116,18 @@ function command(config: LinguiConfig, options) {
 
     if (doMerge) {
       const compileCatalog = getCatalogForMerge(config)
+      const namespace = options.namespace || config.compileNamespace
       const compiledCatalog = createCompiledCatalog(locale, mergedCatalogs, {
         strict: false,
-        namespace: options.namespace || config.compileNamespace,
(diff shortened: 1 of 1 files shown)


thekip:
It should be automatically tested somehow. Could you create a test for that?

Martin005:
> It should be automatically tested somehow. Could you create a test for that?

There are already tests for `getCatalogForMerge()` function. As for testing the `lingui-compile.ts` file, we don't currently have tests for any of our CLI commands – maybe we could add tests in a separate PR.

andrii-bodnar (approved):

PR #1342, opened 2023-01-16
fix(cli): {locale} and {name} replace only once in catalog path

### LINKED ISSUE #1274: {locale} and {name} are only replaced once in the catalog paths
**Describe the bug**

For one of our localization vendors, we need to provide them with unique po file names. We have ~8 components that lingui manages translations for that each have the same filesystem pattern.

catalog path (taken directly from the docs):

```
path: "components/{name}/locales/{locale}/messages",
```

Filesystem:
```
src/
    component-one/
        locales/
            de/
                messages.po
            en/
                messages.po
            ...6 others...
    component-two/
        locales/
            de/
                messages.po
            en/
                messages.po
            ...6 others...
```

To generate **unique filenames and keep the po files separated** into their respective component folders, we'd need something more like the following. The name and locale strings would have to be replaced globally throughout the path string when locating po files and extracting messages.

Catalog path:
```
path: "components/{name}/i18n/locales/{name}_messages_{locale}",
-or-
path: "components/{name}/i18n/locales/{locale}/{name}_messages_{locale}",
```

Ideal Filesystem:
```
components/
    component-one/
        locales/
            component-one_messages_en.po
            component-one_messages_es.po
            ...6 others...
    component-two/
        locales/
            component-two_messages_en.po
            component-two_messages_es.po
            ...6 others...
```

I have a fork w/ some tests that are passing but I don't quite know if that's enough to feel confident... My question is: **Would you be interested in a PR that attempted to do this?**

If the answer is yes, I'll fill out everything below and link my fork!

---

**To Reproduce**

```

   return <Trans>This should be translated!</Trans>
}
```

**Expected behavior**

**Additional context**

- jsLingui version ``lingui --version``
- Babel version ``npm list @babel/core``

Martin005:
This PR fixes #1274 by replacing all occurrences of `{locale}` and `{name}` in the catalog paths.
I have also added tests for checking the implementation.

### DIFF
--- a/packages/cli/src/api/catalog.test.ts
+++ b/packages/cli/src/api/catalog.test.ts
@@ -132,7 +132,7 @@ describe("Catalog", function () {
   })
 
   describe("POT Flow", function () {
-    it('Should merge source messages from template if provided', () => {
+    it("Should merge source messages from template if provided", () => {
       const catalog = new Catalog(
         {
           name: "messages",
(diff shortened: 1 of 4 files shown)


andrii-bodnar:
@Martin005 thank you! 🚀

andrii-bodnar (approved):

Issue #1263, opened 2022-06-28
Major bundlesize increase with @lingui/core 3.14.0


paales:
**Describe the bug**

https://bundlephobia.com/package/@lingui/core@3.14.0

https://github.com/graphcommerce-org/graphcommerce/pull/1536#issuecomment-1168603211

<img width="822" alt="Schermafbeelding 2022-06-28 om 13 39 29" src="https://user-images.githubusercontent.com/1244416/176169729-fdaf4d09-6aa6-4083-9185-f84a1bc930a0.png">

**To Reproduce**

Use new version and measure the resulting bundle size

**Expected behavior**

Bundlesize is in the same ballpark als the older version

ReinoutStevens:
Don't know if this is related, but for us `messageformat-parser` gets included in the production build, even though it should get filtered out when the NODE_ENV is set to production, which is +- 5kb.
I've fixed it by adding a custom webpack rule that uses the null-loader for that package, but it doesn't seem ideal.

paales:
@semoal This one is kinda important, currently stil stuck on the old version :)

paales:
Not stale

sven-meyer-wetter-com:
We noticed this increase in bundle size too. Any change that this could be fixed?

semoal:
Hi guys, sorry for this problem ;| I just checked and the commit that caused the problem was this: 
https://github.com/lingui/js-lingui/pull/1258
I'm open to suggestions about how to handle exports on package.json to bundlers omit automatically ESM/r CJS..

thekip:
@semoal could you elaborate more on that? Do you think bundlers pull to the bundle both version cjs and esm at the same time?

thekip:
Should be fixes by #1343

paales:
@thekip Ohh yeahh! Will check this out soon to see if this resolves the issue :)

PR #1343, opened 2023-01-17
fix(build) fix tree shaking and bundle size

thekip:
Guys, I came with bad news. During working with `size-limit` i noticed that we don't test for the size real entries written in package.json such as  `build/index.js` `build/esm/inde.js` 

Instead, `size-limit` shamelessly set up to test a `production`directly. Assuming that users bundlers will also consume a production bundle. 

I tried to setup size-limit to the real entry, and it showed completely different results. The bundle size increased almost twice.
Firstly, i thought this is a problem with `size-limit` itself, maybe it doesn't replace `process.env.NODE_ENV === "production"` or doesn't apply treeshaking. 
But after hours of digging in the size-limit sourcecode and webpack / esbuild documentation, i've found that `size-limit` pointing to a real problem with current ESM builds. 

## The problem

Current ESM entries written as: 

```ts
import {
  I18nProvider as devI18nProvider, Trans as devTrans, useLingui as devuseLingui, withI18n as devwithI18n
} from "./react.development.js"

import {
  I18nProvider as I18nProviderProd, Trans as TransProd, useLingui as useLinguiProd, withI18n as withI18nProd
} from "./react.production.min.js"

export const I18nProvider = process.env.NODE_ENV === "production" ? I18nProviderProd : devI18nProvider;
export const Trans = process.env.NODE_ENV === "production" ? TransProd : devTrans;
export const useLingui = process.env.NODE_ENV === "production" ? useLinguiProd : devuseLingui;
export const withI18n = process.env.NODE_ENV === "production" ? withI18nProd : devwithI18n;
```

Assuming the bundle will replace `process.env.NODE_ENV === "production"` to the `true` and then only `production` branch would be retained after tree-shaking. 
Unfortunately, bundlers don't work like that. They do replace ENV statement but not tree-shake a "conditional" import. Bundlers are not smart enough and they are very conservative. This code is just not safe to be tree-shaked from theirs point of view. 

So our current ESM builds (which are consumed by all bundlers and shipped to the browser) contains both development and production builds together with bundled dependencies (this is another issue)

Proofs: 
https://esbuild.github.io/api/#conditionally-injecting-a-file

Try it yourself
```bash
cd ./packages/core
esbuild ./build/esm/index.js --bundle --outdir=esbuild --minify --tree-shaking
# inspect output file, both bundles would be in the result as well as message-parser dep
```

## Solution
According to the doc to have this imports to be treeshaked each file should be in its own folder with `package.json` with `{sideEffect:false}`

I didn't go that way, because I think that producing "production" and "development" bundles should be up to the user's bundler. 

It had a sense when library produced as UMD module which you can directly use in browser consuming from CDN. 
But for NPM library it's useless because it's going to be bundled and minified by the user's bundler. 

So I changed the build process to produce only one version of the bundle, with conditions and left the rest of the job for the consumers bundler. 

This make a build process so much simpler and so much closer to the modern build practices. 

## Issue with dependencies
Most of the dependencies were not externalized. Mean they were bundled together with library code. 
Especially `babel/core` and `@format-message/parser`. That happened because `externals` check in rollup options had a logical issue.
I fixed it and wrote more robust check. Everethyng from `node_modules` should be externalized + packages from monorepo. 

## What was changed & updated

1. Remove building production/minified build + dev build in favor of one tree-shakable build for both ESM and CJS
2. Normalize all entries in `package.json` files. Added `exports` field where it was not presented
3. Update Rollup + Plugins to v3
4. Updated babel config to support node v14 instead of v8, and made a separate config for nodejs builds (without react and downleveling syntax for browsers)
5. Rollup typings in one "bundled" typings file according to best practices on the market. 
  > In prior APF versions each entry point would have a src directory next to the .d.ts entry point and this directory contained individual d.ts files matching the structure of the original source code. While this distribution format is still allowed and supported, it is highly discouraged because it confuses tools like IDEs that then offer incorrect autocompletion, and allows users to depend on deep-import paths which are typically not considered to be public API of a library or a package.
6. Leverage Rollup's multioutput feature to produce cjs/esm instead of 2 separate builds processes. This speedup build. 
7. Removed size reporting after rollup build in favor of more real-life reporting by "size-limit", it actually became not compatible with multioutput and I did not want to fix it. 

I also think to change babel to esbuild, but this is not decided yet.


scripts/build/bundles.ts
```diff
@@ -75,24 +72,9 @@ export const bundles: BundleDef[] = [
...
-      !Object.isFrozen(o[prop])
-    ) {
-      deepFreeze(o[prop])
-    }
-  })
-  return o
-}
-
-// Don't accidentally mutate config as part of the build
-deepFreeze(bundles)
```
thekip, line 98: Deleted in favor of Typescript static analysis.

thekip:
Possibly related https://github.com/lingui/js-lingui/issues/1263

scripts/build/rollup.ts
```diff
@@ -239,92 +74,112 @@ function handleRollupError(error) {
     // This looks like an error from a plugin (e.g. Babel).
     // In this case we'll resort to displaying the provided code frame
     // because we can't be sure the reported location is accurate.
-    console.error(error.codeFrame)
+    console.error(error)
```
Martin005, line 77:
```suggestion
    console.error(error.frame)
```

packages/core/src/i18n.ts
```diff
@@ -1,7 +1,7 @@
 import { interpolate, UNICODE_REGEX } from "./context"
 import { isString, isFunction } from "./essentials"
 import { date, number } from "./formats"
-import { compileMessage } from "./compile"
+import { compileMessage } from "@lingui/core/compile"
```
thekip, line 4: That's very important change. If we would access this with relative path, tree shaker would not be able to eliminate this import. 

Because of package import, tree shaker first looks into package's pakage.json, see `{sideEffects: false}` and then remove it from the bundle.

packages/core/src/index.ts
```diff
@@ -1,6 +1,9 @@
 export {
   setupI18n,
   I18n,
+} from "./i18n"
+
+export type {
   AllMessages,
   MessageDescriptor,
   Messages,
```
thekip, line 9: Without typescript plugin, rollup treat type only exports as regular, and obviously fail. Here we do a little help for him and manually mark them as type only.

packages/loader/package.json
```diff
@@ -41,7 +41,8 @@
     "loader-utils": "^2.0.0"
   },
   "devDependencies": {
-    "memory-fs": "^0.5.0"
+    "memory-fs": "^0.5.0",
+    "webpack": "^4.44.2"
```
thekip, line 45: That was defined on monorepo root but actually used here, i fixed it.

scripts/build/rollup.ts
```diff
@@ -1,35 +1,20 @@
...
+import {ModuleFormat, rollup, RollupBuild, RollupError, RollupOptions, RollupWarning} from "rollup"
+import babel from "@rollup/plugin-babel"
+import commonjs from "@rollup/plugin-commonjs"
+import resolve from "@rollup/plugin-node-resolve"
+import babelConfig from './babel.config';
+
+import ora from "ora";
+import chalk from "chalk";
+
+const codeFrame = require("@babel/code-frame")
```
thekip, line 17: this package doesn't have typings

scripts/build/babel.config.js
```diff
@@ -1,15 +1,15 @@
...
+module.exports = function(commonJs) {
   const presets = [
     [
       "@babel/preset-env",
       {
         targets: {
-          node: 8,
-          browsers: "> 1%, last 2 versions"
+          node: 14,
+          browsers: "> 1%, last 2 versions, not dead"
```
thekip, line 10: `not dead` is recommended by browserlist itself. Otherwise, query `last 2 versions` will pull internet explorer and other dead browsers.

scripts/build/rollup.ts
```diff
@@ -1,35 +1,20 @@
...
+import dts from 'rollup-plugin-dts'
+
+import {ModuleFormat, rollup, RollupBuild, RollupError, RollupOptions, RollupWarning} from "rollup"
+import babel from "@rollup/plugin-babel"
+import commonjs from "@rollup/plugin-commonjs"
+import resolve from "@rollup/plugin-node-resolve"
+import babelConfig from './babel.config';
+
+import ora from "ora";
+import chalk from "chalk";
```
thekip, line 15: also i tried to update `chalk` to the latest version, but unfortunately it distributed as ESM package and you can not import ESM package from cjs code. So we stick with this version for a while.

Martin005 (approved):

andrii-bodnar:
@thekip thank you! 🚀

andrii-bodnar (approved):

Issue #1348, opened 2023-01-18
Can't import typescript extractor after updat to 3.16.0


belaczek:
**Describe the bug**
After updating lingui to version 3.16.0, the typescript extractor is no longer exposed by package.json and cannot be imported. Our project relies on it, as the default babel exporter alone has problems with experimental decorators.

**To Reproduce**
Steps to reproduce the behavior, possibly with minimal code sample, e.g:
1. install lingui 3.16.0
2. provide following custom config in lingui.config.ts

```typescript
  extractors: [
    require.resolve('@lingui/cli/api/extractors/typescript'),
  ] as any,

```

**Expected behavior**
`lingui extract` scripts works with imported typescript exporter.

**Additional context**

- jsLingui version: 3.16.0
- Babel version: 7.19.1

thekip:
Would be fixed by https://github.com/lingui/js-lingui/pull/1349

@belaczek what exactly wrong with current extractor and typescript experimental decorators? I will try to take this into account in new version of extractor.

belaczek:
Hi and thanks for reaching back. Our project is somehow specific and full of dependency injections and we have to use `ts-loader` along with `babel-loader` since `babel` alone fails on circular dependencies while `ts-loader` can handle it well.

When I run `lingui extract` with the default extractor, it fails with the following error:

```
Cannot process file xxx.tsx: Duplicate declaration "Trans"
```

I tried to solve this by playing around with our `babel` rules and some other tips mentioned [[here](https://github.com/lingui/js-lingui/issues/952)](https://github.com/lingui/js-lingui/issues/952), but with not much luck. In the end, the only working solution was following:

1. Set `typescript` extractor in `lingui.config.ts`:

```
extractors: [
  require.resolve('@lingui/cli/api/extractors/typescript'),
] as any,
```

2. Create empty `babel` preset for `lingui`:

```
{
  "env": {
    "production": {
      "presets": ["@nrwl/react/babel"]
    },
    "development": {
      "presets": ["@nrwl/react/babel"],
      "plugins": ["react-refresh/babel", "macros"]
    },
    // We want an empty babel config for lingui extract and compile scripts.
    // It uses babel internally and the @nrwl/react/babel preset makes the compilation fail for some reason.
    // The lingui script already injects its own needed plugins anyway.
    "locales": {
      "presets": [],
      "plugins": []
    }
  }
}
```

3. Run the extract script as `BABEL_ENV=locales lingui extract --config lingui.config.ts`.

[...]

PR #1349, opened 2023-01-18
fix(build): explicitly expose api/extractors

thekip:
Fixes https://github.com/lingui/js-lingui/issues/1348

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -29,10 +29,10 @@
     "./api": {
       "require": "./build/api/index.js"
     },
-    "./api/extractors/babel.js": {
+    "./api/extractors/babel": {
       "require": "./build/api/extractors/babel.js"
     },
-    "./api/extractors/typescript.js": {
+    "./api/extractors/typescript": {
       "require": "./build/api/extractors/typescript.js"
     }
   },
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -28,6 +28,12 @@
     },
     "./api": {
       "require": "./build/api/index.js"
+    },
+    "./api/extractors/babel.js": {
+      "require": "./build/api/extractors/babel.js"
+    },
+    "./api/extractors/typescript.js": {
+      "require": "./build/api/extractors/typescript.js"
     }
   },
   "engines": {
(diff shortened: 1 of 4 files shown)


scripts/test-node-api.js
```diff
@@ -0,0 +1,7 @@
+const assert = require("assert")
+assert(require.resolve('@lingui/cli/api'));
+assert(require.resolve('@lingui/cli/api/extractors/typescript.js'));
+assert(require.resolve('@lingui/cli/api/extractors/babel.js'));
+
+assert(require.resolve('@lingui/core'));
+assert(require.resolve('@lingui/core/compile'));
```
thekip, line 7: kinda quick and dirty smoke test that at least all entries we consider public are really exposed.

Martin005 (changes requested):

packages/cli/package.json
```diff
@@ -28,6 +28,12 @@
     },
     "./api": {
       "require": "./build/api/index.js"
+    },
+    "./api/extractors/babel.js": {
+      "require": "./build/api/extractors/babel.js"
+    },
+    "./api/extractors/typescript.js": {
```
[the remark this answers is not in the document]
thekip: You are right, could you change this together with a test  because I will not able to do it quick.

Martin005, line 35:
```suggestion
    "./api/extractors/typescript": {
```

packages/cli/package.json
```diff
@@ -28,6 +28,12 @@
     },
     "./api": {
       "require": "./build/api/index.js"
+    },
+    "./api/extractors/babel.js": {
```
Martin005, line 32:
```suggestion
    "./api/extractors/babel": {
```

scripts/test-node-api.js
```diff
@@ -0,0 +1,7 @@
+const assert = require("assert")
+assert(require.resolve('@lingui/cli/api'));
+assert(require.resolve('@lingui/cli/api/extractors/typescript.js'));
```
Martin005, line 3:
```suggestion
assert(require.resolve('@lingui/cli/api/extractors/typescript'));
```

scripts/test-node-api.js
```diff
@@ -0,0 +1,7 @@
...
+assert(require.resolve('@lingui/cli/api/extractors/babel.js'));
```
Martin005, line 4:
```suggestion
assert(require.resolve('@lingui/cli/api/extractors/babel'));
```

Martin005 (approved):

Issue #1347, opened 2023-01-18
Cannot find module '@lingui/core/compile'


salixor:
**Describe the bug**
Since release 3.16.0, the error "Cannot find module '@lingui/core/compile' from 'node_modules/@lingui/core/build/cjs/index.js'" pops up.
I'm unsure if it is an issue with my setup, or it is related to the global import of "compile" in lingui/core.

**To Reproduce**
No particular steps.

**Expected behavior**
N/A

**Additional context**
- jsLingui version 3.16.0
- Babel version 7.20.12

andrii-bodnar:
@thekip might it somehow be related to the recent changes in the build process?

semoal:
This line is wrong: https://github.com/lingui/js-lingui/blob/main/packages/core/src/i18n.ts#L4
Since we don't do `ts-alias` or something like that after build for replacing local references, when compiled this is not going to get resolved.

This should be `../compiler` instead

andrii-bodnar:
@semoal thank you! There is an explanation of why it was done in such way - https://github.com/lingui/js-lingui/pull/1343#discussion_r1072439050

what do you think about that?

thekip:
You probably running this on a server side?

Martin005:
@thekip Also, the Bundlephobia returns `EntryPointError` for the new versions of [`@lingui/core`](https://bundlephobia.com/package/@lingui/core@3.16.0) and [`@lingui/cli`](https://bundlephobia.com/package/@lingui/cli@3.16.0), but other packages work okay there :thinking:

thekip:
@salixor could you provide more context. I tried to reproduce but with no luck. 

Was it run on the server or not, version of nodejs if bundler was used name and version of the bundler as well. 

Or... the minimal repo

thekip:
@Martin005 i quickly investigated issue with bundlephobia but didn't find a root cause. All entries in package.json pointing into right places. I also compared with other packages and everything seems fine. Need to dig deeper into the bundlephobia sourcecode or just create an issue on theirs repo.

thekip:
[...]

PR #1350, opened 2023-01-19
fix(core): sub entry export for react-native

thekip:
Attempt to fix https://github.com/lingui/js-lingui/issues/1347

Node 12+  
@lingui/core/compile -> package.json exports -> build/cjs/compile.js

React Native or other older CJS runtimes not supporting `exports`:
@lingui/core/compile -> /compile.js ->  build/cjs/compile.js

Development: Jest + Typescript Language Server 
@lingui/core/compile -> tsconfig path mapping -> src/compile/index.ts

Guys, i'm on vacation till Monday, i haven't chance to test it with verdacio + /examples, i assume it would work, but better to test


Martin005:
Can confirm that this PR fixes the React Native and Bundlephobia! 👍

packages/core/compile.js
```diff
@@ -1,5 +1,5 @@
 /**
- * this is used to point to the right source when development with workspaces
+ * This entry for old runtimes which is not support `exports` field in package.json
```
Martin005, line 2:
```suggestion
 * This entry is for old runtimes which do not support `exports` field in package.json
```

Martin005 (changes requested):

Martin005 (approved):

andrii-bodnar:
@thekip @Martin005 thank you!

PR #1351, opened 2023-01-19
fix(cli): incorrect main entry in package.json

Martin005:
This PR fixes the incorrect `main` entry in `package.json` of `@lingui/cli`. This was [causing Bundlephobia to report ](https://bundlephobia.com/package/@lingui/cli@3.16.0) `EntryPointError` on the new 3.16.0 version of the package.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -18,13 +18,13 @@
     "name": "Tomáš Ehrlich",
     "email": "tomas.ehrlich@gmail.com"
   },
-  "main": "./build/index.js",
+  "main": "./build/lingui.js",
   "bin": {
     "lingui": "./build/lingui.js"
   },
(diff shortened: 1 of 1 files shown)


andrii-bodnar (approved):

thekip:
That's wrong fix, you exposed something which should not be exposed. To make it Right fyou should just delete main and . entries, because this package is not going to be consumed any other why other than from 'bin'

Your fix saying is literally saing "hey, I have  consumable entrypoint" for everyone, which actually is not. 

Also report for CLI on bundlephobia is useless, no one is going to actually bundle a cli tool

thekip:
@andrii-bodnar @Martin005 fyi

andrii-bodnar:
makes sense 🤔

PR #1356, opened 2023-01-22
chore(verdaccio): more obvious way of clean working tree before verdaccio release

thekip:
All these direct amending of git index is not obvious and intuitive. Let's make it more safe and straightforward.

### DIFF
--- a/scripts/verdaccio-release.js
+++ b/scripts/verdaccio-release.js
@@ -1,7 +1,6 @@
 const { exec: _exec } = require("child_process")
 const chalk = require("chalk")
 const ora = require("ora")
-const os = require("os")
 
 async function releaseInVerdaccio() {
   const spinner = ora()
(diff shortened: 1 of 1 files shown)


Martin005:
@thekip What about using `git stash` and popping the stash afterwards instead of committing the changes?

thekip:
Stash will reset current files and save theirs state into "stash", this is not what we want. With commit, it's safer because if something went wrong, you can open history and revert commit. With direct git index manipulation i few times was in the situation where i have no files changed according to git log, but i see that files are actually changed, and it was really hard to figure out and really hard to bring them back "as changed"

Martin005 (approved):

andrii-bodnar:
@thekip something went wrong here https://github.com/lingui/js-lingui/pull/1368 (the verdaccio test started to fail) after this merge, could you please check why it's happening? Is it related?

thekip:
@andrii-bodnar yes its related. I'm trying to understand why in general this verdaccio release included in GHA wrokflow

thekip:
@andrii-bodnar ok, found, this actually really nice script, don't undersatnd why it's not included in the PR pipiline. I will fix in separate PR

PR #1358, opened 2023-01-23
refactor(extractor)!: don't use intermediate catalogs

thekip:
Faster, safer, less code

- [x] add TS typings
- [x] remove storing to intermediate catalogs and following traversing fs to build final catalog
- [x] introduce `onMessageExtracted` callback in babel-plugin options 
- [x] fix unit tests on the babel plugin side
- [x] fix test on the extractor side

BREAKING CHANGES: 

- The public type of extractor was changed. 
```diff
- extract(filename: string, targetDir: string, options?: ExtractOptions): Promise<void> | void
+ extract(filename: string, onMessageExtracted: (msg: ExtractedMessage) => void, options?: ExtractOptions): Promise<void> | void
```
  This should affect only small amount of users, only those who use custom extractors. 

- Also intermidiate catalogs no longer created, but i'm sure no one is using it.


packages/babel-plugin-extract-messages/test/fixtures/jsx-without-macros.js
```diff
@@ -1,13 +1,12 @@
 import { Trans } from "@lingui/react"
 
 ;<span id="ignore" />
-;<Trans /> // this should be ignored
-;<Trans message="Missing ID" /> // this should be ignored
```
thekip, line 5: These cases moved into their own `it` in suite

packages/cli/src/api/catalog.ts
```diff
@@ -336,7 +342,7 @@ export class Catalog {
       (fallbackLocales?.default &&
         getTranslation(fallbackLocales.default as string)) ||
       // Get message default
-      catalog[key]?.defaults ||
+      // catalog[key]?.defaults ||
```
thekip, line 345: There is no field "defaults", it's something what left from previous versions.

packages/conf/index.d.ts
```diff
@@ -1,5 +1,15 @@
...
+export type ExtractedMessage = {
+    id: string,
+
+    message?: string,
+    context?: string,
+    origin?: [filename: string, line: number],
+
+    comment?: string,
+}
+
```
thekip, line 12: There is a mess in @lingui/conf typings. They are copy-pasted two times. That's why this type also appears here 2 times

packages/babel-plugin-extract-messages/test/fixtures/js-message-descriptor.js
```diff
@@ -1,11 +1,7 @@
-/*i18n*/'Message'
```
thekip, line 1: i think original author's ideawas to test extraction from StringLiteral, but actually this code is a directive prologue according to ecma languge grammar. It would be a StringLiteral only when it assigned to the variable or used sometging else. 

So i deleted it from here and created a specific `it` in the suite

packages/babel-plugin-extract-messages/test/fixtures/js-message-descriptor.js
```diff
@@ -1,11 +1,7 @@
...
-
 const msg = /*i18n*/{id: 'Message'}
 
 const withDescription = /*i18n*/{id: 'Description', comment: "description"}
 
-const withMultilineComment = /*i18n*/{id: 'Multiline', comment: "this is " + "2 lines long"}
-
```
thekip, line 8: Created a specific suite, and also tested negative cases where variable passed as value.

thekip:
It's probably better to merge it to the "next" branch, or create a new one for the upcomming release.

andrii-bodnar:
@thekip please assign yourself all the related issues so it will be easier to recognize what issues actually are in progress

andrii-bodnar:
@Martin005 @thekip from my perspective, we probably need to create a fresh copy of the `next` branch from the `main` for such breaking changes and then move forward with pre-releases from the `next` branch until it is ready for official release and merge to the `main`.

In this case, the `main` and v3 will not be blocked for the minor and patch releases.

I'm going to investigate the possible flow in more detail.

packages/cli/src/api/catalog.ts
```diff
@@ -23,17 +22,18 @@ const LOCALE_REPLACE_RE = /{locale}/g
...
-type MessageOrigin = [string, number?]
+type MessageOrigin = [filename: string, line?: number]
 
 export type ExtractedMessageType = {
   message?: string
   origin?: MessageOrigin[]
-  extractedComments?: string[]
+  // todo: this actually not fulfilled by extractor, but actively used in PO formatter
+  //  need to check what was the difference from extractedComments and delete one
   comments?: string[]
```
thekip, line 32: https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html

`comments` - from translator to developer
`extractedComments` - from developer to translator 

So extractor not related to `comments` at all. These comments should be preserved while merging catalogs.

packages/cli/src/lingui-extract.ts
```diff
@@ -26,6 +26,7 @@ export default async function command(
   config: LinguiConfig,
   options: Partial<CliExtractOptions>
 ): Promise<boolean> {
+  // todo validate and delete
```
thekip, line 29: delete

thekip:
We have to consider this also https://github.com/lingui/js-lingui/pull/1299

Right now not all failures are reported as errors. Many of them reported just as warning in console. (for example missing ID)

andrii-bodnar:
@thekip does this PR requires a documentation update?

We can start writing the v3 -> v4 guide simultaneously with that and extend this instruction by each further PR (just to not forget anything to mention in this guide)

thekip:
This one is not requiring changing documentation.

thekip:
rebased

packages/cli/src/lingui-extract.ts
Martin005: Please remove this comment if it is outdated.

packages/cli/src/api/catalog.ts
```diff
@@ -23,17 +30,18 @@ const LOCALE_REPLACE_RE = /{locale}/g
...
 
-type MessageOrigin = [string, number?]
+type MessageOrigin = [filename: string, line?: number]
 
 export type ExtractedMessageType = {
   message?: string
   origin?: MessageOrigin[]
-  extractedComments?: string[]
+  // todo: this actually not fulfilled by extractor, but actively used in PO formatter
+  //  need to check what was the difference from extractedComments and delete one
```
Martin005, line 39: Please remove these comments

thekip:
Done

thekip:
Also rebased again onto `next` branch

Martin005 (approved):

andrii-bodnar (approved):

PR #1362, opened 2023-01-23
fix(build): Revert typescript update

Martin005:
This PR reverts the update of the Typescript package made in #1357 which caused the following error in the `create-react-app` example when running the `react-scripts start` command:

> `TypeError: Cannot assign to read only property 'jsx' of object **'#<Object>'`

This shows we need to have stronger integration tests put in place.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "strip-ansi": "^6.0.0",
     "swc-node": "^1.0.0",
     "ts-jest": "^26.4.1",
-    "typescript": "^4.9.4",
+    "typescript": "^4.0.3",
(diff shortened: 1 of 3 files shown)


andrii-bodnar (approved):

andrii-bodnar:
@thekip FYI

Please take a look at this issue with typescript update

thekip:
Did you try to investigate a real root cause?

thekip:
I was not able to reproduce it.

thekip:
@Martin005 fyi

thekip:
@Martin005 any clarifications here? We need never typescript...

Issue #1363, opened 2023-01-23
Typescript not finding types when using NodeNext


blahbleepew:
**Describe the bug**
Typescript fails to find index.d.ts when using NodeNext / esm modules module resolution.

Relevant thread: microsoft/TypeScript#50466

**To Reproduce**

```
//tsconfig.json
"compilerOptions": {
    "target": "ESNext",
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
}
```

Here is a patch that fixes it:

```
diff --git a/package.json b/package.json
index 58aa79f9a78ff9abcbfd82672e89047450e8cb1c..a2fa3818553b8664b03d670ed16c5e160d76c137 100644
--- a/package.json
+++ b/package.json
@@ -31,8 +31,14 @@
   },
   "exports": {
     ".": {
-      "require": "./build/cjs/index.js",
-      "import": "./build/esm/index.js"
+      "require": {
+        "types": "./build/index.d.cts",
+        "default": "./build/cjs/index.cjs"
+      },
+      "import": {
+        "types": "./build/index.d.ts",
+        "default": "./build/esm/index.js"
+      }
     },
     "./compile": {
       "require": "./build/cjs/compile.js",
```

thekip:
What package it's related to?

dinassh:
I have the same issue with the `@lingui/core` package with versions `3.16.0` and `3.16.1`.

The provided patch fixes the issue for me.

thekip:
What version of typescript was used? I'm trying to reproduce, but everything works seems fine. 

You can check my repo and fork if needed
https://github.com/thekip/lingui-nodenext-issue

thekip:
Looks like `"strict": true` is also needed to reproduce

dinassh:
@thekip Thank you for taking a look!

I can reproduce on your example with an additional `tsconfig.json` compiler option which we are using on our project: 
```json
    "strict": true
```

With this configuration I'm getting an error:
```sh
yarn test
yarn run v1.22.19
warning package.json: No license field
$ tsc
index.ts:1:22 - error TS7016: Could not find a declaration file for module '@lingui/core'. '~/lingui-nodenext-issue/node_modules/@lingui/core/build/cjs/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/lingui__core` if it exists or add a new declaration (.d.ts) file containing `declare module '@lingui/core';`

1 import { i18n } from '@lingui/core'
[...]

PR #1366, opened 2023-01-24
chore(): package.json properly expose TS typings for dual packages

thekip:
fixes: https://github.com/lingui/js-lingui/issues/1363


andrii-bodnar:
I would suggest renaming the `e2e-tests` -> `tests` and creating inside a separate directory `e2e`. In case some new test types are added in the future

e2e-tests/typescript-nodenext-resolution/package.json
```diff
@@ -0,0 +1,14 @@
+{
```
andrii-bodnar, line 1: what about `"private": true`?

thekip: it doesn't matter, lerna setup to publish only from `./packages`

andrii-bodnar: anyway, make it consistent. The `e2e-tests/node-api/package.json` has this flag

thekip:
> I would suggest renaming the e2e-tests -> tests and creating inside a separate directory e2e. In case some new test types are added in the future

I would rather rename it to "sandboxes", or something like this. Because they are a not a test in a basic meaning, they are also involved in yarn workspaces install mechanism etc. 
So creating a subfolder may make this all setup is too complicated and confusing.

e2e-tests/typescript-nodenext-resolution/readme.md
```diff
@@ -0,0 +1,3 @@
+Check support typescript's NodeNext moduleResolution with dual `@lingui` packages
+
+https://github.com/microsoft/TypeScript/issues/50466
```
andrii-bodnar, line 3: move it to the `index.ts` as a comment, this readme doesn't make a lot of sense

andrii-bodnar:
sandboxes will confuse even more. I mean something like here - https://github.com/ionic-team/stencil/tree/main/test

thekip:
Lets rename it to test, but without additional level. Each folder in that directory - different test case, which might be e2e or any other type

thekip:
Done. I also changed version constaint to "*" because lerna will not bump versions in these folders, but we should rely on the latest available in monorepo.

andrii-bodnar (approved):

PR #1370, opened 2023-01-25
docs(): various refinements

andrii-bodnar:


packages/cli/src/lingui-extract.ts
```diff
@@ -127,7 +127,7 @@ if (require.main === module) {
     hasErrors = true
     const msg =
       "--babelOptions option is deprecated." +
-      " Please set extractBabelOptions in configuration https://lingui.js.org/ref/conf.html#extractBabelOptions"
+      " Please set extractBabelOptions in configuration https://lingui.dev/ref/conf#extractBabelOptions"
```
Martin005, line 130: This anchor doesn't work as anchors are case sensitive. Please update `#extractBabelOptions` to be in lower-case

packages/core/README.md
```diff
@@ -25,9 +25,9 @@ This package is licensed under [MIT][license] license.
 
 [license]: https://github.com/lingui/js-lingui/blob/main/LICENSE
 [linguijs]: https://github.com/lingui/js-lingui
-[documentation]: https://lingui.js.org/
-[tutorial]: https://lingui.js.org/tutorials/js.html
-[reference]: https://lingui.js.org/ref/core.html
+[documentation]: https://lingui.dev
+[tutorial]: https://lingui.dev/tutorials/js
```
Martin005, line 29: This was moved to `https://lingui.dev/tutorials/javascript`

Martin005 (changes requested):
There are still a few places in the code that point to `lingui.js.org`. Could you please update them as well? 🙂

andrii-bodnar:
There are still a few `lingui.js.org` links in the `CHANGELOG.md` files, I think it will be better not to touch these files and keep this history as is.

Martin005 (approved):

PR #1371, opened 2023-01-25
chore(verdaccio): setup verdaccio integration test on PR + fix execution

thekip:
Using trial & error approach found that if we version & publish in one command we can completely remove manipulation with git state. 

I also enabled this WF for pull requests because it's very nice to have e2e tests run automatically. 

And i'm going in next iterations to extend this script to run tests not only in CRA project but also in rest examples as well

### DIFF
--- a/.github/workflows/release-test.yml
+++ b/.github/workflows/release-test.yml
@@ -1,6 +1,9 @@
 name: release-workflow-test
 
 on:
+  pull_request:
+    branches:
+      - '*'
   push:
     branches:
       - main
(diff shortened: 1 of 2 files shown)


.github/workflows/release-test.yml
```diff
@@ -28,7 +31,7 @@ jobs:
 
       - name: Publish with Verdaccio
         run: |
-          git checkout ${{ github.head_ref }}
+          git switch -c "pull-request"
```
thekip, line 34: `action-checkout` left git in "detached state". Original 
`git checkout ${{ github.head_ref }}`

Was aimed to fix this and make state attached again. But this notation doesn't work with pull request from forks (as this one) because `github.head_ref` is pointing to the ref local to the fork not an original repo. 

`git switch -c "pull-request"` also do the trick, and works with pull request as well.

andrii-bodnar (approved):

Martin005:
@thekip @andrii-bodnar After this PR, we are not doing `lerna version patch --exact` and this means we will have dynamic internal package versioning 💔

thekip:
This not affect a release process, it's only for integration testing

thekip:
And also --exact has a diffrent purpose

https://github.com/lerna/lerna/tree/main/commands/version#--exact

thekip:
I checked it a couple times more, and this works like a charm. Even better than before!

PR #1374, opened 2023-01-26
chore: improve PR template

andrii-bodnar:


.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
```diff
@@ -1,10 +1,23 @@
...
+
+## Types of changes
+
+[//]: # (What types of changes does your code introduce to Lingui?)
+[//]: # (_Put an `x` in the boxes that apply_)
+
+- [ ] Bugfix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
+- [ ] Documentation update
```
Martin005, line 13: Maybe it would be good to add "Chore" as a type of change 🙂

andrii-bodnar: Good point 🙂

andrii-bodnar: Or maybe we can leave here only these essential types. Detailed type anyway will be specified in the PR title, it's not required to select these checkboxes in this case (chore, ci, refactoring, etc)

Martin005: I think we can leave only the essential types here for now

Martin005 (approved):

PR #1375, opened 2023-01-26
chore(): improve contributing guide

thekip:
I've found that with yarn it's possible to update by scope + with custom registry in one command without a necessity to rollback registry. 

Added this to the contribution guide.

Fell free to change wording or improve formatting if you have any ideas. I'm bad at writing docs.


thekip:
It works in a bit different way than `update-by-scope` package, but don't let it confuse you. It updates packages and wright new version in yarn.lock. It don't touch `package.json` if new version satisfy semver written for packages.

andrii-bodnar (approved):

PR #1380, opened 2023-01-27
fix: types declaration inside package entrypoints

khmm12:
ESM module resolver doesn't fallback to main + module + types when `exports` present, that means types won't be discovered if they're not declared inside `exports`.

Links:
- https://www.typescriptlang.org/docs/handbook/esm-node.html
- https://nodejs.org/api/packages.html

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -31,10 +31,12 @@
   },
   "exports": {
     ".": {
+      "types": "./build/index.d.ts",
       "require": "./build/cjs/index.js",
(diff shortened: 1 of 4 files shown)


khmm12:
Oops I found.
https://github.com/lingui/js-lingui/issues/1363
https://github.com/lingui/js-lingui/pull/1366

Then I leave the PR up to maintainers :-)

Martin005:
@khmm12 Thank you, this has been already resolved by #1366, so I am going to close this one 🙂

Issue #1295, opened 2022-10-26
Custom TypescriptLoader for cosmic config breaks import/require module resolution


thekip:
**Describe the bug**
Create typescript config file `lingui.config.ts` with any import or require
```typescript
import { ALL_LANGUAGES_ARRAY, DEFAULT_LANGUAGE } from './languages';

module.exports = {
  locales: ALL_LANGUAGES_ARRAY.map((l) => l.key),
  sourceLocale: DEFAULT_LANGUAGE,
  catalogs: [{
    path: '<rootDir>/translates/{locale}',
    include: ['<rootDir>/src'],
  }],
  format: 'po',
};
```

Run the extractor. And you will get an error pointing that module not found 

If you rename file to plain js it works. 
If you put a `console.log(__dirname)` in both files you will see that js files executed from it's original place, where ts file executed from insides of lingui package. 

That's caused by custom TypescriptLoader for cosmiconfig defined here 
https://github.com/lingui/js-lingui/blob/da070ba294435484fde10d82ecb895d2410ec1b7/packages/conf/src/index.ts#L580-L587

Which is basically just eval the code in place. 
 
**Rationality** 
Although the configuration is usually very simple, it's not the case where you have a monorepo with different frameworks and tools. Writing and maintaining list of languages in every app in such monorepo could be cumbersome and easy to forget to add language in one of places. That's why we use one monorepo-wide file with all languages definitions written in TS. 

**Proposed solution:**
Get rid of this custom, "fast" and actually, broken loader in favour of standart one. 

**Additional context**

- jsLingui version ``3.14.0``

semoal:
This should be fixed on https://github.com/lingui/js-lingui/pull/1247, 3.15.0

Give it a try, :)

thekip:
It still an issue. The mentioned PR updated a package which not used in the code. Current codebase still has: 

```ts
    loaders: {
      ".ts": TypeScriptLoader,
    },

   //....
/** Typescript loader using just typescript API and eval(), instead of using ts-node/register which is slower */
function TypeScriptLoader(filePath: string) {
  const tsc = require("typescript")
  const fileContent = fs.readFileSync(filePath, "utf-8")
  const { outputText } = tsc.transpileModule(fileContent, { compilerOptions: { module: tsc.ModuleKind.CommonJS }});
  const configFileParsed = eval(outputText)
  return configFileParsed
}
```

Which is not working as it should

PR #1386, opened 2023-01-31
fix(conf): proper typescript cosmiconfig loader

thekip:
fixes: https://github.com/lingui/js-lingui/issues/1295

### DIFF
--- a/packages/conf/src/index.test.ts
+++ b/packages/conf/src/index.test.ts
@@ -1,5 +1,4 @@
 import path from "path"
-import mockFs from "mock-fs"
 import { validate } from "jest-validate"
 import {
   getConfig,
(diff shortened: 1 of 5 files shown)


packages/conf/src/index.test.ts
```diff
@@ -143,34 +142,24 @@ describe("@lingui/conf", function () {
...
-        mockFs.restore()
-      })
-
       it("if fallbackLocale is defined, we use the default one on fallbackLocales", () => {
-        mockFs({
-          ".linguirc": JSON.stringify({
-            locales: ["en-US"],
-            fallbackLocale: "en",
-          }),
-        })
```
thekip, line 170: mockFs is actually very dangerous because it breaks resolution of files in underlying libraries. typescript-comiconfig is not an exclusion. It didn't work with mockFs, so I removed it in favour of real file in FS.

Martin005 (approved):

andrii-bodnar (approved):

PR #1388, opened 2023-01-31
fix(compile): remove verbose output when using template (*.pot)

thekip:
# Description

When using flow with ".pot" file it generates unneeded log entries on every catalog load. 
This was done unintentional and fixed in this PR. 

I also added integration test for the `compile` cli command covering this and related cases. 

There is also a growing problem with `@lingui/conf` package. Typings in src and and index.d.ts are mismatched, and bringing potential problems and blocking from implementing standard flow. I hope i will take this into work in the near future. 

## Types of changes

- [ ] Documentation update

Fixes # (issue)

## Checklist


packages/cli/src/lingui-compile.ts
```diff
@@ -18,9 +18,18 @@ const noMessages: (catalogs: Object[]) => boolean = R.pipe(
...
+  namespace?: string,
+}
+
+export function command(config: LinguiConfig, options: CliCompileOptions) {
   const catalogs = getCatalogs(config)
 
+  // fixme: this is definitely doesn't work
   if (noMessages(catalogs)) {
     console.error("Nothing to compile, message catalogs are empty!\n")
     console.error(
```
thekip, line 35: This doesn't work completely.  I wrote test, and it fails, I think fixing this logic may introduce a breaking change for some users (because some of theirs catalogs would be rejected from compilation).
 I don't know what to do with this, may be delete completely? What is the value of this validation?

packages/cli/src/lingui-compile.ts
```diff
@@ -35,23 +44,25 @@ function command(config: LinguiConfig, options) {
   const doMerge = !!config.catalogsMergePath
   let mergedCatalogs = {}
 
-  console.error("Compiling message catalogs…")
+  console.log("Compiling message catalogs…")
```
thekip, line 47: There was inconsistency in "log" and "error". Sometimes error used in place of log and vice-versa

packages/cli/src/lingui-compile.ts
```diff
@@ -35,23 +44,25 @@ function command(config: LinguiConfig, options) {
...
 
-  config.locales.forEach((locale) => {
+  for (const locale of config.locales) {
```
thekip, line 49: Changed to the loop to be able to "early return" from function. 

`process.exit(1)` used previously for early return is not test-friendly

packages/cli/src/lingui-compile.ts
```diff
@@ -35,23 +44,25 @@ function command(config: LinguiConfig, options) {
...
+    // todo: this validation should be in @lingui/conf
     if (locale !== config.pseudoLocale && !plurals[language]) {
-      console.log(
+      console.error(
         chalk.red(
           `Error: Invalid locale ${chalk.bold(locale)} (missing plural rules)!`
         )
       )
       console.error()
+      // todo: should we exit here?
```
thekip, line 59: As written in the todo this validation should be in `@lingui/conf`, i checked the git history and looks like it's here by historical reasons (all validations were here at the beginning) 

I didn't move that because it's out of the scope of current PR, and you don't like big PRs, right 😉?

packages/cli/src/test/compile.test.ts
```diff
@@ -0,0 +1,203 @@
...
+      }]
+    })
+
+    const result = command(config, {})
+    expect(result).toBeFalsy()
+  })
+
+  describe('Merge Catalogs', () => {
+    // todo
+  })
```
thekip, line 46: @Martin005 Would you be minded to write tests for the latest changes  to the merging functionality introduced by you?

packages/conf/src/index.ts
```diff
@@ -136,13 +138,22 @@ export function getConfig({
     ? configExplorer.load(configPath)
     : configExplorer.search(defaultRootDir)
   const userConfig = result ? result.config : {}
+
+  return makeConfig({
+    rootDir: result ? path.dirname(result.filepath) : defaultRootDir,
+    ...userConfig,
+  }, {skipValidation})
+}
```
thekip, line 146: i split `getConfig`into loading part (`getConfig`) which do all cosmiconfig related stuff and processing part (`makeConfig`) which validates, populate defaults, etc 

The latter is makes testing much easier.

packages/cli/src/lingui-compile.ts
andrii-bodnar: Is the issue in this `if (noMessages(catalogs))` condition?

packages/cli/src/lingui-compile.ts
andrii-bodnar: If possible, it's always a good idea to keep PRs small or at least create separate commits for unrelated changes 🙂 

We can keep this validation here for a while. As I understand, if it were in `@lingui/conf`, it would prevent such an error at an earlier stage?

Regarding the second `todo` - should we commit such todos? It risks remaining eternal todo

packages/cli/src/lingui-compile.ts
thekip: Yes, I think previously `noMessages` was supposed to receive message catalogs in "plain format" something like: 

```ts
{
 catalog1: {
    mesageId1: "text"
  },
 catalog2: {
    mesageId1: "text"
  }
}
```

But currently it receives instances of `Catalog` class, which always has some internal fields and this check is always passed

packages/cli/src/lingui-compile.ts
thekip: > Regarding the second todo - should we commit such todos? It risks remaining eternal todo
It's a question to maintainers here. What do you think, should we exit here? 

Pros: 

- If you run this in CI, it will return `1` and will fail job and you got your error earlier. 

Cons: 

I don't know how users use it and how can it affect them. Maybe they use some not-real locale intentionally?

andrii-bodnar: It could be some custom language

thekip: > We can keep this validation here for a while. As I understand, if it were in @lingui/conf, it would prevent such an error at an earlier stage?

It's ideologically should be in @lingui/conf. All other validation for locales applied there.
Moreover, I think the most popular setup now is not using `cli compile` at all, instead it uses loaders (webpack, etc) and this check would never be applied. 

Also, this check could be useful in other places, not only in compile.

thekip: > It could be some custom language

For what? If it is a pseudolocale it would not throw error. So i think in the vast majority of cases it would be user misconfiguration.

andrii-bodnar: That makes sense

thekip: > should we exit here?

Anyway, when it would be moved to @lingui/config it would not be possible to early return (at least with current implementation)

andrii-bodnar:
> @andrii-bodnar any doubts here?

Will finish the review today

packages/cli/src/api/catalog.ts
```diff
@@ -280,10 +280,6 @@ export class Catalog {
   ) {
     const catalog = catalogs[locale] || {}
 
-    if (!catalog.hasOwnProperty(key)) {
-      console.error(`Message with key ${key} is missing in locale ${locale}`)
```
andrii-bodnar, line 284: This might be useful. Maybe it would be better to change it to the `console.warn()` and make it non-blocking? Or print it only in `--verbose` mode

packages/cli/src/test/compile.test.ts
andrii-bodnar: We can create issues labeled with the `good first issue` for such cases to encourage the community to contribute

packages/cli/src/lingui-compile.ts
andrii-bodnar: It would be indeed better to exit with the `1` exit code here. It's one of the common practices for CLI tools - exiting with this code in case of errors.

packages/cli/src/lingui-compile.ts
```diff
@@ -68,14 +79,14 @@ function command(config: LinguiConfig, options) {
...
-            missingMsgIds.forEach((msgId) => console.log(msgId))
+            missingMsgIds.forEach((msgId) => console.error(msgId))
           } else {
             console.error(
               chalk.red(`Missing ${missingMsgIds.length} translation(s)`)
             )
           }
           console.error()
-          process.exit(1)
+          return false
```
andrii-bodnar, line 89: will it be correctly handled and exit with the `1` code as a result?

thekip: Yes, when command return false the orchestrator script would return 1 (it's below in the code)

packages/cli/src/api/catalog.ts
thekip: It's a duplicating logic, which is left here by historical reasons. There is a "allowEmpty" option which is aimed for exactly that purpose. When "allowEmpty: false" (wich is usually in production) the CLI will exit with error and message. 

When `allowEmpty = false && verbose = true` it will also print all messages which is empty.  
Check the tests i wrote for more info. 

This piece of code was not covered by tests, and probably it was an unreachable path before i implemeted "*.pot" files as a source for keys in this PR https://github.com/lingui/js-lingui/pull/1296

thekip:
What's left to merge this pr?

andrii-bodnar:
@thekip the codecov reports a 2.70% coverage decrease, could you please check why this is happening?

thekip:
Because codecov incorrectly calculate coverage. It calculates how many procents of **imported** files are covered. So if you have 100 files and 1 test which imports exactly one file and cover all lines in that file. codecov will show 100% coverage even when other 99 files not touched at all. 

Exactly that is happened here, I added to test file which was never tested and imported in suite before and this file expanded "base" for a coverage.

thekip:
It's not related to codecov itself, it related to how coverage tools works in general. They know only about files which were included in program when coverage was collected

andrii-bodnar (approved):

andrii-bodnar:
@thekip thank you!

packages/cli/src/lingui-compile.ts
Martin005: @thekip @andrii-bodnar There was a `process.exit(1)` here in the past. But as PR #1280 (which removed it) correctly pointed out, there are some languages which don't have plurals ([are not present in CLDR](https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/language_plural_rules.html)). So I think we don't think we should exit here at all.

thekip: @Martin005 nice catch. I think detecting validity of locale by checking the absence of plural-rules is incorrect in general. I think we still need validate them and exit if error, but use cldr for that.  This is of course out of the context of this PR. I will remove the `todo` here

thekip:
rebased and updated

Martin005 (approved):

thekip:
I have no idea why windows tests a failing. There nothing related in this PR

andrii-bodnar:
@thekip it happens sometimes for unknown reason 😬 
just restarted the job

PR #1389, opened 2023-01-31
fix(macro): JS macros don't strip non-essential props in production

### LINKED ISSUE #1379: JS macro don't strip non-essential props on production.
The JSX macro (`<Trans>`, `<Plural>`, etc) has a feature, when you set `NODE_ENV === 'production'` it strips properties needed only for translators and saving the final bundle size. 

Unlike JSX macro, js doesn't do that. Even though it's mentioned in the docs. It seems it was never implemented. 

By the way, SWC Plugin implementing both optimization 💪

Martin005:
# Description
This PR fixes #1379.
The current implementation of macros was stripping non-essential props (`message` and `comment`) only in JSX macros, not in JS macros. This PR brings the stripping of these props in JS macros on par with JSX macros.

There are several changes to the `macroJs.js` file, but most of them are changes from Prettier. The only change to code relating to this bug is the implementation at the end of the `processDescriptor()` function.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/macro/src/macroJs.ts
+++ b/packages/macro/src/macroJs.ts
@@ -175,10 +175,6 @@ export default class MacroJs {
 
     let descriptor = this.processDescriptor(path.node.arguments[0])
 
-    if (process.env.NODE_ENV === "production") {
-      descriptor = this.stripNonEssentialProps(descriptor)
-    }
-
     path.replaceWith(descriptor)
   }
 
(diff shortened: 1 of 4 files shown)


packages/macro/src/macroJs.ts
```diff
@@ -159,16 +173,29 @@ export default class MacroJs {
     // reset the expression counter
     this._expressionIndex = makeCounter()
 
-    const descriptor = this.processDescriptor(path.node.arguments[0])
+    let descriptor = this.processDescriptor(path.node.arguments[0])
+
+    if (process.env.NODE_ENV === "production") {
+      descriptor = this.stripNonEssentialProps(descriptor)
+    }
```
thekip, line 180: Could you move this check into `processDescriptor` method and avoid code duplication in `replaceDefineMessage` and `replaceTAsFunction`?

andrii-bodnar (approved):

PR #1397, opened 2023-02-02
chore: setup husky + lint staged + prettier

thekip:
# Description

Setup eslint + prettier on the precommit hook with husky. 

## Types of changes

- [ ] Documentation update

Fixes # (issue)

## Checklist


.prettierrc.json
```diff
@@ -0,0 +1 @@
+{}
```
thekip, line 1: empty config file to let editors and other tools know you are using Prettier

https://prettier.io/docs/en/install.html

package.json
```diff
@@ -15,7 +15,7 @@
     "watch": "cross-env TZ=UTC NODE_ICU_DATA=node_modules/full-icu jest -c jest.config.js --watch",
     "watch:integration": "cross-env TZ=UTC NODE_ICU_DATA=node_modules/full-icu jest -c jest.config.integration.js --watch",
     "lint:types": "tsc",
-    "lint:eslint": "eslint packages",
+    "lint:eslint": "eslint './packages/**/*.{ts,tsx,js,jsx}'",
```
thekip, line 18: `eslint packages` by default would match only "*.js" files

package.json
```diff
@@ -68,18 +69,20 @@
...
+    "lint-staged": "^13.1.0",
     "memory-fs": "^0.5.0",
     "minimist": "^1.2.5",
     "mock-fs": "^4.13.0",
     "mockdate": "^3.0.2",
     "ncp": "^2.0.0",
     "npm-cli-login": "^0.1.1",
     "ora": "^5.1.0",
-    "prettier": "^2.1.2",
+    "prettier": "2.8.3",
```
thekip, line 85: Version of prettier should be save exactly:

> This makes sure that everyone in the project gets the exact same version of Prettier. Even a patch release of Prettier can result in slightly different formatting, so you wouldn’t want different team members using different versions and formatting each other’s changes back and forth.

.prettierrc.json
andrii-bodnar: what about [`.prettierrc`](https://github.com/lingui/js-lingui/blob/main/.prettierrc)?

package.json
```diff
@@ -24,7 +24,8 @@
     "release:latest": "lerna publish from-package --dist-tag latest --yes",
     "release:next": "lerna publish from-package --canary --preid next --pre-dist-tag next --yes",
     "build:docs": "cd website && yarn install && yarn build",
-    "size-limit": "size-limit"
+    "size-limit": "size-limit",
+    "prepare": "husky install"
```
andrii-bodnar, line 28: Probably it should be documented in the `CONTRIBUTING.md`

package.json
thekip: By the way, eslint has almost no rules. It has few plugins added, but hey have no sense for typescript projects. 
I'm going to setup eslint from scratch in the near future.

.prettierrc.json
thekip: Fixed

package.json
thekip: What you mean by that? As far as i undersand prepare should [automaticlly be called](https://docs.npmjs.com/cli/v9/using-npm/scripts#life-cycle-scripts) on local install. So everething should work "out of the box" for new contributors, isn't it?

andrii-bodnar: Hm, indeed. I thought it should be called manually. Nevermind

andrii-bodnar (approved):

Martin005 (approved):

Martin005:
@thekip Could you please resolve the conflicting `yarn.lock`?

thekip:
done

Martin005:
@thekip It seems you will need to fix the lint issues:

- in `cli/src/api/pseudoLocalize.ts` the first line importing ramda needs to be changed from `import R from "ramda"` to `import * as R from "ramda"`
- in `macro/test/types.ts` the first line imports `MessageDescriptor`, but that needs to be imported as type.

thekip:
Bumped version of `@typescript-eslint/parser` to avoid: 

> WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.
> You may find that it works just fine, or you may not.
> SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.5.0
> YOUR TYPESCRIPT VERSION: 4.9.5
> Please only submit bug reports when using the officially supported version.

And removed `eslint-plugin-import` completely as it useless in Typescript projects. 

@Martin005 what you mentioned above it's false-positives due to the fact `eslint-plugin-import` aimed to add to old-plain js projects some of the benefits of static import analysis and obviously it doesn't know anything about typescript files and used ts compiler options

Martin005 (approved):

andrii-bodnar (approved):

Issue #1401, opened 2023-02-03
Cannot find module 'ts-node'


niksauer:
**Describe the bug**
```
$ lingui compile
node:internal/modules/cjs/loader:998
  throw err;
  ^

Error: Cannot find module 'ts-node'
Require stack:
- /Users/nik/Developer/flexibits/flexibits-web-frontend/node_modules/@lingui/conf/node_modules/cosmiconfig-typescript-loader/dist/cjs/index.js
- /Users/nik/Developer/flexibits/flexibits-web-frontend/node_modules/@lingui/conf/build/index.js
- /Users/nik/Developer/flexibits/flexibits-web-frontend/node_modules/@lingui/cli/build/lingui-compile.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:995:15)
    at Function.Module._load (node:internal/modules/cjs/loader:841:27)
    at Module.require (node:internal/modules/cjs/loader:1067:19)
    at require (node:internal/modules/cjs/helpers:103:18)
    at Object.<anonymous> (/Users/nik/Developer/flexibits/flexibits-web-frontend/node_modules/@lingui/conf/node_modules/cosmiconfig-typescript-loader/dist/cjs/index.js:28:22)
    at Module._compile (node:internal/modules/cjs/loader:1165:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1219:10)
    at Module.load (node:internal/modules/cjs/loader:1043:32)
    at Function.Module._load (node:internal/modules/cjs/loader:878:12)
    at Module.require (node:internal/modules/cjs/loader:1067:19) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Users/nik/Developer/flexibits/flexibits-web-frontend/node_modules/@lingui/conf/node_modules/cosmiconfig-typescript-loader/dist/cjs/index.js',
    '/Users/nik/Developer/flexibits/flexibits-web-frontend/node_modules/@lingui/conf/build/index.js',
    '/Users/nik/Developer/flexibits/flexibits-web-frontend/node_modules/@lingui/cli/build/lingui-compile.js'
  ]
}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

**To Reproduce**
```
{
  "dependencies": {
    [...]
    "@lingui/react": "^3.17.0",
    "react-scripts": "5.0.1",
  },
  "scripts": {
    [...]
[...]

PR #1403, opened 2023-02-03
fix(conf): lazy load cosmiconfig's TypeScriptLoader

thekip:
# Description

This PR is attempt to fix: https://github.com/lingui/js-lingui/issues/1401

[cosmiconfig-typescript-loader](https://github.com/Codex-/cosmiconfig-typescript-loader/blob/main/lib/loader.ts) written in the way that it doesn't matter do you really load a Typescript file in config or not, once this loader is added to cosmiconfig it requires ts-node and typescript. This is indeed unwanted behavior because some of the users may not have them and may not want to use them at all. 

This PR changes 2 things: 

1. Lazy load the loader, once cosmiconfig discovered a config with ts extension
2. Add peerDependecies on `typescript` and `ts-node` for the `@lingui/conf` package. And also i marked them as optional. 

Why this happened? Why our test didn't catch this? 

It's because node modules resolution system works like that. 
Take 'examples/create-react-app' as an example. It has it's own pacakge.json and own `node_modules`. 
Furthermore, it doesn't list `ts-node` in this file, and indeed `node_modules` folder doesn't have this dependency. So why this project doesn't fall with that change? 

It's because we have another `node_modules` folder on the root level of monorepo and node found this module there. 

- root
  - node_modules <-- from here
  - examples
      - create-react-app
          - node_modules
          - package.json

     
## Types of changes

- [ ] Documentation update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/conf/package.json
+++ b/packages/conf/package.json
@@ -34,5 +34,17 @@
   ],
   "devDependencies": {
     "ts-node": "^10.9.1"
+  },
+  "peerDependencies": {
+    "ts-node": ">=10",
+    "typescript": ">=4"
+  },
+  "peerDependenciesMeta": {
+    "ts-node": {
+      "optional": true
+    },
+    "typescript": {
+      "optional": true
+    }
   }
 }
--- a/packages/conf/src/index.ts
+++ b/packages/conf/src/index.ts
@@ -2,9 +2,8 @@ import type { GeneratorOptions } from "@babel/core"
 import path from "path"
 import fs from "fs"
 import chalk from "chalk"
-import { cosmiconfigSync } from "cosmiconfig"
+import { cosmiconfigSync, type LoaderSync } from "cosmiconfig"
 import { multipleValidOptions, validate } from "jest-validate"
-import { TypeScriptLoader } from "cosmiconfig-typescript-loader";
 
 export type CatalogFormat = "lingui" | "minimal" | "po" | "csv"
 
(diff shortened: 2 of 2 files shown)


andrii-bodnar:
Have no opinion on that but it looks ok if it solves the problem. @thekip did you manage to reproduce it?

@niksauer @ryanb can you please check if this fix works in your cases?

ryanb:
@thekip thanks for your work on this.

@andrii-bodnar I've tried adding the `@lingui/conf` package pointing to this branch using [GitPkg](https://gitpkg.vercel.app)

```
yarn add 'https://gitpkg.now.sh/thekip/js-lingui/packages/conf?fix/typescript-cosmiconfig'
```

This appears to work on my local, however my deployment pipeline results in this error.

```
/app/release/20230204001656/node_modules/@lingui/conf/index.js:1
export * from "./src"
^^^^^^

SyntaxError: Unexpected token 'export'
    at wrapSafe (internal/modules/cjs/loader.js:1001:16)
    at Module._compile (internal/modules/cjs/loader.js:1049:27)
    ...
```

Perhaps I'm not installing the package correctly? Is there an alternative way to install the package from this branch?

I only saw the `ts-node` error during deployment as well. I haven't yet determined why it's only showing up during deployment.

thekip:
@ryanb please follow a contribution guide. You need to build package and upload to Verdaccio, then you would be able to install it as usually. Of course on the ci that would not work because verdaccio would not be available outside your machine.

niksauer:
Can confirm this fix is working, having followed the [contribution guide](https://github.com/lingui/js-lingui/blob/main/CONTRIBUTING.md#using-development-version-in-your-project). Thanks!

andrii-bodnar (approved):

Martin005 (approved):

PR #1415, opened 2023-02-08
refactor(conf): split code, generate typings

thekip:
# Description

Refactoring of `@lingui/conf` package.  No changes in logic or behavior. This PR makes package clean and maintainable. 

- Code was split into files
- Deprecation migrations extracted from base file and grouped in own files
- Removed `index.d.ts` which is in turn a d.ts snapshot of former index.ts made by someone, some time ago. 
- Build changed to TSC, so typings now produced automatically and they are aligned with code in ./src
- privates is no longer exposed from package. Type definitions not bloated by `defaultConfig`, `exampleConfig` etc
- Introduced type `LinguiConfigNormalized`:
   `LinguiConfig` is used for typing user provided config. Some of the properties might be in various forms. Config package process user provided config, normalize this various forms into one and that shape become `LinguiConfigNormalized` and supposed to use in the library internals
- I wrote new tests, fixed old ones, and fixed all places where Typescript complained about types mismatch

## Types of changes

- [ ] Documentation update

Fixes # (issue)

## Checklist


packages/cli/src/test/compile.test.ts
```diff
@@ -172,13 +172,13 @@ msgstr ""
 
           const actualFiles = readFsToJson("/test")
 
+          mockFs.restore()
+
```
thekip, line 176: Just found that one of the tests is always "One snapshot written", that's because mockFs wasn't restored before snapshotting.

thekip:
Changed base to main as discussed in [discord](https://discord.com/channels/974702239358783608/1073174343275773992)

packages/conf/src/migrations/catalogMigration.ts
```diff
@@ -0,0 +1,69 @@
...
+      {
+        path: pathJoinPosix(newLocaleDir, "{locale}", "messages"),
+        include: srcPathDirs ?? ["<rootDir>"],
+        exclude: srcPathIgnorePatterns ?? ["**/node_modules/**"],
+      },
+    ]
+  }
+
+  return newConfig as LinguiConfig
+}
```
thekip, line 69: I think we can remove it in this PR, because this is going to be released with version 4. And was deprecated in version 3, together with FallbackLanguage migration.

packages/conf/src/types.ts
```diff
@@ -0,0 +1,77 @@
+import { GeneratorOptions } from "@babel/core"
+import { ExtractOptions } from "@lingui/cli/src/api/extractors"
```
thekip, line 2: This import is incorrect. conf should not rely on cli package, please see dicussion: 

https://discord.com/channels/974702239358783608/1072883644273795082

andrii-bodnar:
@thekip could you please explain more about code coverage-related changes? The codecov reports a huge coverage decrease

thekip:
> @thekip could you please explain more about code coverage-related changes? The codecov reports a huge coverage decrease

It's not related. The changes just disable coverage gathering by default in the config, and enabling reporting in yarn commands, which should be used in WF. This was done to make local development more convenient by not gathering coverage when you run tests one by one during coding. 

That should not affect coverage anyhow. Probably it's again wrong base chosen by codecov. 
Anyway i reverted this changes they are not related to this PR and was added to commit by mistake. I would create a separate PR with that changes.

packages/conf/src/utils/replaceRootDir.ts
```diff
@@ -0,0 +1,21 @@
+export function replaceRootDir<T>(config: T, rootDir: string): T {
```
Martin005, line 1: As this function is used only for replacing th "<rootDir>" string in the config file, I think that `config` can be of type `LinguiConfig`

thekip: Actually not, it not bind to shape of LinguiConfig. It can take anything, and return anything. And it's not directly compatible with LinguiConfig. This function called in a recursion, and subsequent calls could be with type "string" or array, etc. So if you put `config: LinguiConfig` typescript would complain

thekip: try it yourself please.

thekip: done

packages/conf/src/migrations/catalogMigration.ts
thekip: @Martin005 @andrii-bodnar i'm waiting for an answer here.

Martin005 (approved):

packages/conf/src/migrations/catalogMigration.ts
thekip: i will remove it next iteration. Let's merge this. I want to unblock other PRs

andrii-bodnar (approved):

PR #1431, opened 2023-02-12
feat(cli): Extract - Flatten ICU messages

### LINKED ISSUE #1186: Suggestion: option to flatten ICU Messages
FormatJS's `extract` command has an option called `--flatten`. The documentation is self-explanatory so I'll paste it here:

> Whether to hoist selectors & flatten sentences as much as possible. E.g:
> `"I have {count, plural, one{a dog} other{many dogs}}"`
> becomes `"{count, plural, one{I have a dog} other{I have many dogs}}"`.
> The goal is to provide as many full sentences as possible since fragmented
> sentences are not translator-friendly.

Do you think it would make sense to have something similar in Lingui?

Martin005:
# Description
Fixes #1186 

Based on suggestion from #1186, this PR adds flattening of ICU messages for the `extract` and `extract-template` CLI commands.

## Types of changes

- [ ] Documentation update

## Checklist

- [X] I have added necessary documentation (if appropriate)

### DIFF
--- a/packages/core/src/flatten/flattenMessage.ts
+++ b/packages/core/src/flatten/flattenMessage.ts
@@ -1,0 +1,51 @@
+import { parse, Token } from "@messageformat/parser"
+import { printICU, isPluralOrSelect } from "./printICU"
+
+/** @internal */
+export function flattenMessage(message: string): string {
+  try {
+    let tokens = parse(message) as Token[]
+    const flattenedTokens = flattenTokens(tokens)
+    return printICU(flattenedTokens)
+  } catch (e) {
+    console.error(`${(e as Error).message} \n\nMessage: ${message}`)
+    return message
+  }
+}
+
+function flattenTokens(tokens: Array<Token>): Array<Token> {
+  for (let i = 0; i < tokens.length; i++) {
+    const token = tokens[i]
+    if (isPluralOrSelect(token)) {
+      const cloned = cloneDeep(token)
+      const { cases } = cloned
+      cloned.cases = cases.reduce((all, k, index) => {
+        const newValue = flattenTokens([
+          ...tokens.slice(0, i),
+          ...cases[index].tokens,
+          ...tokens.slice(i + 1),
+        ])
+        all[index] = { ...k, tokens: newValue }
+        return all
+      }, [])
+
+      return [cloned]
+    }
+  }
+  return tokens
+}
+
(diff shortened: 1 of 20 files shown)


andrii-bodnar (approved):

thekip (changes requested):
How this supposed to work in general? When the message would be flattened, it would be a different string. And it would not match to id string from runtime.

packages/core/src/flatten/flattenMessage.ts
```diff
@@ -0,0 +1,51 @@
...
+        const newValue = flattenTokens([
+          ...tokens.slice(0, i),
+          ...cases[index].tokens,
+          ...tokens.slice(i + 1),
+        ])
+        all[index] = { ...k, tokens: newValue }
+        return all
+      }, [])
+
+      return [cloned]
```
thekip, line 32: Without `reduce`, mutating data and `cloneDeep`: 

```ts
return [{
  ...token,
  cases: token.cases.map((k, index) => {
    const newValue = flattenTokens([
      ...tokens.slice(0, i),
      ...cases[index].tokens,
      ...tokens.slice(i + 1)
    ])

    return {...k, tokens: newValue}
  })
}]
```

thekip:
@Martin005 

> How this supposed to work in general? When the message would be flattened, it would be a different string. And it would not match to id string from runtime.

I see here a couple ways how this could be done: 

1. This logic should be applied in the macro and not in extractor. This way macro will generate flatten messages, left them in the bundle where extractor would pick them up. But we should consider SWC version of macro, and porting these changes to the SWC would be hard (because it depends on `@message-parser`). So i would not recommend this option. 

2. Implement it together with hash based id proposal. This way a hash would be generated by macro from the original, untouched message. And extractor with `--flat` would process only "messages" not ids. The only one problem here - this will not work for users who not used macro and not use explicit ids.  Actually, for v4 we can force explicit ids to be used for non-macro usage.

Martin005:
@thekip I have been thinking and I believe that the second option would be the best way to go.

thekip:
Yes, this is exactly how i'm see implementation of hash-based ids with macro. 

Two cases: 

1. Without macro, `i18n._()` call, id is required. Message would not be used as substitution.
2. Macro `t` calls will generate hash based ids from message if explicit ids is not specified, and this behavior would be enabled as default.

Issue #1175, opened 2021-12-09
JSX macros aren't working inside JSX expressions in JSX macros


khmm12:
**Describe the bug**

Since v3 JSX macros aren't working inside JSX expressions in JSX macros. v2 macros are working as expected.

The issue is reproducible in any nesting level.

**To Reproduce**

```jsx
import { Trans } from '@lingui/macro'

export default function Component(props) {
  return <Trans>Hello, {props.world ? <Trans>world</Trans> : <Trans>guys</Trans>}</Trans>
}
```

**Expected behaviour**
```jsx
import { Trans } from '@lingui/react'

export default function Component(props) {
  return (
    <Trans
      id='Hello, {0}'
      values={{
        0: props.world ? <Trans id='world' /> : <Trans id='guys' />
      }}
    />
  )
}
```

**Actual behaviour**
```jsx
import { Trans } from '@lingui/react'

export default function Component(props) {
  return (
    <Trans
      id='Hello, {0}'
      values={{
        0: props.world ? <Trans>world</Trans> : <Trans>guys</Trans>
      }}
    />
  )
}
```

**Additional context**

- `lingui`: `3.13.0`
- `@babel/core`: `7.16.0`

semoal:
Thx @khmm12  will take a look as soon as possible. Indeed looks a regression

tulodzieckim-simplito:
Any news about this?

khmm12:
No, still actual.

khmm12:
Up

thekip:
If some one takes this task, please, also check that in the SWC plugin. The test case should ported there.

thekip:
Checked this in SWC, this also don't work. I made a quick fix for 1 level of nesting for js macro. But if there would be more levels, that would break again. To fix this, fundamental changes needed. Still think how to do that.

thekip:
fixed by https://github.com/lingui/js-lingui/pull/1436 and would be released with v4

PR #1436, opened 2023-02-14
feat(macro): support JSX macro inside conditional expressions

thekip:
# Description

This PR adding syntax support for `ConditionalExpressions` in JSX Macro: 

```ts
import { Trans } from '@lingui/macro'
<Trans>Hello, {props.world ? <Trans>world</Trans> : <Trans>guys</Trans>}</Trans>
```

To make reviewing easier, i split changes into 2 commits.

- First  commit refactoring JSX macro to operate with babel's `path` instead of `node`
- Second commit fix actual issue. 

We need `path` because babel doesn't allow traversing over plain `node`. And manual traversing wasn't a case here because there might be a syntax which we didn't expect. (multiple nested ternaries or function call which accepts jsx node, etc)

Also issue in SWC version repo: https://github.com/lingui/swc-plugin/issues/16

## Types of changes

- [ ] Documentation update

fixes: https://github.com/lingui/js-lingui/issues/1175

## Checklist

### DIFF
--- a/packages/macro/src/index.ts
+++ b/packages/macro/src/index.ts
@@ -52,7 +52,7 @@ const jsxMacroTags = new Set(["Trans", "Plural", "Select", "SelectOrdinal"])
 function macro({ references, state, babel, config }: MacroParams) {
   const opts: LinguiMacroOpts = config as LinguiMacroOpts
 
-  const jsxNodes: NodePath[] = []
+  const jsxNodes = new Set<NodePath>()
   const jsNodes: NodePath[] = []
   let needsI18nImport = false
 
@@ -64,9 +64,13 @@ function macro({ references, state, babel, config }: MacroParams) {
         jsNodes.push(node.parentPath)
       })
     } else if (jsxMacroTags.has(tagName)) {
+      // babel-plugin-macros return JSXIdentifier nodes.
+      // Which is for every JSX element would be presented twice (opening / close)
+      // Here we're taking JSXElement and dedupe it.
+
       nodes.forEach((node) => {
         // identifier.openingElement.jsxElement
-        jsxNodes.push(node.parentPath.parentPath)
+        jsxNodes.add(node.parentPath.parentPath)
       })
     } else {
       throw nodes[0].buildCodeFrameError(`Unknown macro ${tagName}`)
@@ -82,7 +86,9 @@ function macro({ references, state, babel, config }: MacroParams) {
     if (macro.replacePath(path)) needsI18nImport = true
   })
 
-  jsxNodes.filter(isRootPath(jsxNodes)).forEach((path) => {
+  const jsxNodesArray = Array.from(jsxNodes)
+
+  jsxNodesArray.filter(isRootPath(jsxNodesArray)).forEach((path) => {
     if (alreadyVisited(path)) return
     const macro = new MacroJSX(babel, { stripNonEssentialProps })
     macro.replacePath(path)
@@ -92,7 +98,7 @@ function macro({ references, state, babel, config }: MacroParams) {
     addImport(babel, state, i18nImportModule, i18nImportName)
   }
 
-  if (jsxNodes.length) {
+  if (jsxNodes.size) {
     addImport(babel, state, TransImportModule, TransImportName)
   }
 }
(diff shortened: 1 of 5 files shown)


packages/macro/src/macroJsx.test.ts
```diff
@@ -1,18 +1,17 @@
...
     plugins: ["jsx"],
-  }) as JSXElement
+  }) as any
 
 function createMacro() {
   return new MacroJSX({ types }, { stripNonEssentialProps: false })
 }
 
-describe("jsx macro", () => {
+xdescribe("jsx macro", () => {
```
thekip, line 14: This unit tests now quite hard to fix. Because there no way to parse a codesnippet and receive a `path` as it was done with `node`. 

I will try to fix it, but if not, we can simply delete them because they're just testing the same things as e2e test (which is more preferable in this context)

packages/macro/src/macroJsx.ts
```diff
@@ -332,56 +334,69 @@ export default class MacroJSX {
...
     // !!! Important: Calculate element index before traversing children.
     // That way outside elements are numbered before inner elements. (...and it looks pretty).
     const name = this.elementIndex()
-    const children = R.flatten(
-      node.children.map((child) => this.tokenizeChildren(child)).filter(Boolean)
-    )
+    const children = path
+      .get("children")
+      .flatMap((child) => this.tokenizeChildren(child))
+      .filter(Boolean)
```
Martin005, line 397: This logic is the same as in the `tokenizeTrans()` function. It would be good to have this logic only in one place.

thekip: done

packages/macro/src/macroJsx.test.ts
thekip: Fixed

packages/macro/test/jsx-trans.ts
```diff
@@ -196,6 +196,50 @@ const cases: TestCase[] = [
...
+      `,
+    expected: `
+        import { Trans } from '@lingui/react'
+
+        ;<Trans
+            id={"Hello, {0}"}
+            values={{
+              0: props.world ? <Trans id={'world'} /> : <Trans id={'guys'} />
+            }}
+          />
```
thekip, line 213: I'm thinking that we can automatically generate from 1 level ternary an ICU select expression: 

`Hello, { props.world, select, =true {world} =false {guys} }`

Or i'm overengineering...

packages/macro/tsconfig.json
```diff
@@ -1,7 +1,10 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "lib": [
+      "ES2019.Array"
+    ],
```
thekip, line 6: This is for Array.prototype.flat / Array.prototype.flatMap 
Supported started from Node.js 11 AFAIK

packages/macro/src/macroJsx.test.ts
```diff
@@ -302,20 +318,22 @@ describe("jsx macro", () => {
...
+              }),
+              format: "select",
+              options: {
+                male: "he",
+                female: "she",
+                other: "they",
+              },
             },
-          },
+          ],
```
thekip, line 336: Was a single element - become an array

packages/macro/test/jsx-trans.ts
Martin005: `Select` needs to always have the `other` case, so I don't really see a way how to do that.

Martin005 (approved):

packages/macro/src/index.ts
```diff
@@ -135,6 +140,13 @@ function addImport(
   }
 }
 
+/**
+ * Filtering nested macro calls
+ *
+ * <Macro>
+ *   <Macro /> <-- this would be filtered out
+ * <Macro>
```
andrii-bodnar, line 148: Shouldn't be a closing tag here?

thekip: yes

packages/macro/test/jsx-trans.ts
thekip: `Hello, { props.world, select, =true {world} other {guys} }`

But it's overengineering anyway

andrii-bodnar (approved):

PR #1437, opened 2023-02-14
fix(macro): Trans inside Plural has leading whitespace

### LINKED ISSUE #1292: Trans inside Plural has surrounding whitespace
I have the latest lingui and faced the same [issue](https://github.com/lingui/js-lingui/issues/747)

```tsx

<Plural
      one={
        <Trans>
          Care Navigator{" "}
          <Badge
            aria-label={t`You have unread message`}
          >
            #
          </Badge>
        </Trans>
      }
      other={
        <Trans>
          Care Navigator{" "}
          <Badge
            aria-label={t`You have unread message`}
          >
            #
          </Badge>
        </Trans>
      }
      value={unreadSendBird}
/>

```

becomes

```
msgid " Care Navigator <0>#</0>"
msgid_plural " Care Navigator <1>#</1>"
msgstr[0] " Care Navigator <0>#</0>"
msgstr[1] " Care Navigator <1>#</1>"

```

Version: 
```
"@lingui/cli": "^3.14.0",
"@lingui/loader": "^3.14.0",
"@lingui/macro": "^3.14.0",
"@lingui/react": "^3.14.0",
```

Martin005:
# Description
Fixes #1292 

## Types of changes
- [X] Bugfix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

- [X] I have added necessary documentation (if appropriate)


thekip:
@Martin005 please, to `next` branch. We agreed on codefreeze on a main branch. 

It also should be ported to the SWC plugin, otherwise would be id mismatch with messages in catalog.
Extractor uses results from the js macro, so any changes to the extraction results would affect SWC plugin.

packages/macro/test/jsx-plural.ts
```diff
@@ -89,6 +89,47 @@ const cases: TestCase[] = [
...
+  {
+    input: `
+        import { Trans, Plural, t } from '@lingui/macro';
+        <Plural
+              one={
+                <Trans>
+                  Care Navigator{" "}
+                  <Badge
+                    aria-label={t\`You have unread message\`}
+                  >
```
thekip, line 101: Let's focus in test in one exact case what we are want to check. Make a test case as minimal as possible (without `aria-label={t\`You have unread message\`}`)

thekip (changes requested):

Martin005:
@thekip There are already a few commits to the `main` branch: https://github.com/lingui/js-lingui/pull/1422/commits and if we want to release some `3.x` version with these bugfixes, this is the better way to go.
As for the SWC plugin, I am already looking into it and will create a PR in a moment.

andrii-bodnar:
Yes, it's a small fix that can be easily merged and released within a 3.x version

thekip:
This small fix will also break existing catalogs with translations because of different whitespaces. 

Also SWC plugin version should be fixed with extractor version otherwise - id mismatch.

thekip:
@Martin005 could you add a unit test into `packages/macro/src/macroJsx.test.ts` "jsx macro" -> "normalizeWhitespace" suite

packages/macro/test/jsx-plural.ts
```diff
@@ -89,6 +89,42 @@ const cases: TestCase[] = [
         }} />;
       `,
   },
+  {
+    input: `
+        import { Trans, Plural} from '@lingui/macro';
+        <Plural
```
thekip, line 95: could you also write a name for the test case, otherwise it's not clear what exactly we test there.

packages/macro/test/jsx-plural.ts
```diff
@@ -89,6 +89,42 @@ const cases: TestCase[] = [
...
+              one={
+                <Trans>
+                  Care Navigator{" "}
+                  <Badge>
+                    #
+                  </Badge>
+                </Trans>
```
thekip, line 102: Is having a 

```tsx
   <Badge>
    #
    </Badge>
```

Is important for the sake of testing this case?

Would it be enough just:

```ts
import { Trans, Plural} from '@lingui/macro';
<Plural
  one={
    <Trans>
      ...
    </Trans>}
  other={
    <Trans>
      ...
    </Trans>
  }
  value={count}
/>;
```
To reproduce the case?

Martin005:
@thekip
> This small fix will also break existing catalogs with translations because of different whitespaces.

Yeah, I keep forgetting that... I will change the base branch to `next`.

> Also SWC plugin version should be fixed with extractor version otherwise - id mismatch.

Here is the PR for SWC plugin: https://github.com/lingui/swc-plugin/pull/17

Martin005:
@thekip
>could you add a unit test into packages/macro/src/macroJsx.test.ts "jsx macro" -> "normalizeWhitespace" suite

Added in [this commit](https://github.com/lingui/js-lingui/pull/1437/commits/bbb477639099f69b043523474df9bf7c38c42030)

thekip (approved):

thekip:
In the meantime, we can think about how make generating id's from original message less fragile and depend on the spaces. I have been thinking about that and don't see a clear and easy solution. 

Removing all whitespaces from the message before generating the hash? 

Not an option because: `<Trans>Hello{' '}</Trans>` and `<Trans>Hello</Trans>` would collide.

andrii-bodnar (approved):

andrii-bodnar:
@thekip you're right, any message modification before generating hash can have unexpected downsides, so maybe it would be better not to do this

Issue #1360, opened 2023-01-23
RFC: Context + hash based message id


thekip:
Current lingui implementation has a flaw:  you have your original message in the bundle at least 2 times + 1 translation. 
For a line "Hello world" you would have one in the source code as id in i18n call: 
```ts
const message = i18n._('Hello world!');
```
Another one as key in the message catalog and then a translation itself:
```ts
export const messages = {
 'Hello world': 'Cześć swiat!'
}
```

Strings could be very long, not just a couple words, so this may bring more kb to the bundle. 

If you use context for the string it would be even more kilobytes because for each msg which should have a context you would have context explicitly written in the bundle + in the msg catalog. 

```ts
const message = t({
   context: 'My context',
   message: 'Hello world!'
})

export const messages = {
  'My context':  {
     'Hello world': 'Cześć swiat!'
  }
}
```
A much better option would be generating a "stable" ID based on the msg + context as hash with fixed length. 

So internal structure would look like 
```
{
 [<hash>]: 'Cześć swiat!'
}
```
And lookups in runtime would be much simpler. 

Hash would be calculated at build time by macros. So macros instead of

```ts
const message = t({
   context: 'My context',
   message: `Hello`
})

// ↓ ↓ ↓ ↓ ↓ ↓

import { i18n } from "@lingui/core"
const message = i18n._(/*i18n*/{
   context: 'My context',
   id: `Hello`
})
```

Would generate: 

```ts
import { i18n } from "@lingui/core"
const message = i18n._(/*i18n*/{
   id: "<hash(message + context)>",
   message: `Hello`, // only for development purpose, would be dropped in production
})
```

The context feature would affect ONLY messageid generation, and thus would only be supported when using macro. For the non-macro users using context has no sense because they provide a messageid manually. 

Default message and context would be stripped from runtime code and replaced to the ID generated in compile time.

andrii-bodnar:
@thekip if I understand correctly, these hashes will appear in the localization files like PO or JSON as keys, right?

thekip:
In po files everything would be the same, because po files support context natively. For json formats it is correct.

That's exactly as it works in other most popular libraries, such as https://formatjs.io/docs/getting-started/message-extraction

tricoder42:
You're right, I never get to implementing this idea. It seems possible:

1. extract messages from source code and translate
2. compile message catalog - this will "minify" message Ids
3. build code for production — this time, macros use minified ids

[...]

Issue #1167, opened 2021-11-16
Make plurals fallback to those of locale fallback


germanferrero:
**Is your feature request related to a problem? Please describe.**
TLDR; Make plurals fallback to those of locale fallback.

We are migrating to Lingui for [our preact web application](https://github.com/germanferrero/lime-app/tree/a509fbb760953d46d869468415b4586ccbbbf438).
We choosed `@lingui/detect-locale` to detect the browser locale with `fromNavigator`. But, we only have full support for english (`en`) , spanish (`es`) and portuguese (`pt`). In any other case, we would like to fallback to `en` locale, and use plurals rules for that locale too.

At [.linguirc](https://github.com/germanferrero/lime-app/blob/a509fbb760953d46d869468415b4586ccbbbf438/.linguirc) we set up the following, in order to accomplish this:
```javascript
{
...,
"fallbackLocales": {
      default: "en"
   }
}
```

Now, when we a user with japanese (`ja`) locale use the app, we get this error message in the console:
```
Plurals for locale ja aren't loaded. Use i18n.loadLocaleData method to load plurals for specific locale. Using other plural rule as a fallback.
```

And effectively "other plural" rule is always used.
   
**Describe proposed solution**
This issue is related to https://github.com/lingui/js-lingui/issues/798.
But that one was considered closed by falling back to the other plural rule always.
I think it would be better, and more intuitive to fallback both locale and plural rules to the default fallback locale. 

**Additional context**
Other non so user-friendly experience with fallback locales that we have:
- If the user's browser is en-US for instance, by default fallback to its CLDR parent locale `en`, but again the plurals rules do not fallback to `en` plural rules.

**Describe alternatives you've considered**
We ended up trimming the "country" specification from the browser locale before locale activation to avoid plural rules errors regarding CLDR parent locales. But of course this is not what we want to do with our i18n.

germanferrero:
> Hey, maybe stale bot shouldn't be marking this as stale, don't you think?

...

thekip:
`fallbackLocales` settings in `lingui.config.js` has effect only on compilation of catalog. It is not intended to affect you runtime logic anyhow. 

It's up to host application how to process situations when user asked a version which you do not support. 
In your case, it's simple:
```ts
const availableLanguages = ["en", "es", "pt"]; 
const locale = availableLanguages.includes(requestedLocale) ? requestedLocale : "en";

i18n.activate(locale);
```

PR #1440, opened 2023-02-15
feat: Hash based id

thekip:
Implementation for: https://github.com/lingui/js-lingui/issues/1360

For easier reviewing, i split this feature in the different commits. I recommend checking them one by one.

- [x] implement in js macro and update tests
- [x] implement in jsx macro and update tests
- [x] cleanup runtime / extractors from "runtime" context
- [x] changes in formats
- [x] porting macro changes to SWC Plugin https://github.com/lingui/swc-plugin/pull/19
- [x] fallbackLocales, fallback to template, and fallback in case of empty translation https://github.com/lingui/js-lingui/pull/1443
- [x] documentation

### implementation details
1. Js macro now always generate call with message descriptor, no matter what input was
    ```js
     i18n._({id: "mjnlP0", message: "Ola!"})
    ```
2. Id is generated as base64(sha256(value)).length(6). Unlike the formatjs where this placeholder is configurable, i explicitly didn't want to give flexibility here. Because it comes down to complexity of porting this to Rust. 
2. i found very exhausting to updating tests with new ids, since we don't use jest snapsohots for them. So i created a small babel plugin used in tests which is stripping ids from the test cases where checking id is not a purpose of test. 
4. i created a separate package `message-id` which helps with creating ids from messages. I need it in few places that's why it's extracted.  
5. context has effect only on message id generation and in PO formatter (all during compile time). All usages in `compile` and runtime was removed. Context property also stripped in production mode. 
6. po formatter works a bit tricky. We don't want to have our hash based id in the `msgid` field since it against the gettext ideology. But we also don't want to break compatibility with custom ids provided by users. For that purpose during serialization we're storing a [flag ](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html#:~:text=Comment%20lines%20starting%20with%20%23%2C%20contain%20flags%3B%20more%20about%20these%20below.)`generated-id` in po file, which then used when deserializing to regenerate hash id. 
   input: 
   ```js
   
    {
      id: <hash>,
      message: "Ola!",
      context: "My Contex",
    }
   ```
    output: 
   ```po
   #, generated-id
    msgctxt "my context"
    msgid "Ola!"
    msgstr ""
   ```
   We also can do vice-versa, store flag when explicit id provided, this is TBD. (and like this option more because in my mind there should more autogenerated id's than custom)

## Types of changes

- [ ] Documentation update

Fixes # (issue)

## Checklist


JSteunou:
Thank you for leading this hard work @thekip this is a top notch long time awaiting feature :)

About the hash, would sha256 be a bit more time consuming than sha1 ?

thekip:
> About the hash, would sha256 be a bit more time consuming than sha1 ?

I have been thinking about what hash function to use for a while. The criteria were - short / with low collision rate / non cryptographical / implemented in Rust

I considered
- https://sites.google.com/site/murmurhash/
- adler32

Related SO topics:
https://stackoverflow.com/questions/4567089/hash-function-that-produces-short-hashes

But i couldn't choose and just took what formatjs is using as default, assuming they choose it by the reason. 

I think for regular catalogs with few thousands of messages,  the difference in speed between  sha256 or sha1 would be negligible

JSteunou:
> I think for regular catalogs with few thousands of messages, the difference in speed between sha256 or sha1 would be negligible

That what it seems indeed https://github.com/hex7c0/nodejs-hash-performance :+1: ~same speed, less collision

thekip:
@Martin005 @andrii-bodnar guys i want a feedback on these changes before I start porting it to the Rust. Updating test suite is extremely exhausting, so i don't want to do it more than one time.

andrii-bodnar:
@thekip currently I have that one suggestion about the new package

also, I'm worrying about this PO flag - it looks like the PO Gettext specification has already a predefined list of allowed flags, isn't it? Or it allows any other flags also?

thekip:
By the way, i'm actually more like idea when we put a flag for explicit id, instead of generated.  I'm assuming more users should use it this way then another

andrii-bodnar:
@thekip the same for me

We can make this flag even more informative for the explicit ids, something like that:

```
#, id: some.custom.id
msgctxt "..."
msgid "..."
msgstr ""
```

It can provide a piece of additional information for translators. For example, in Crowdin it will look like this:

image1

instead of this:

image2

andrii-bodnar:
@thekip sure, I didn't mean to use flags for that, it was more like an example

It looks fine in your example with `extracted-comments`

thekip:
Take these to examples: 

```ts
t({id: "my.custom.id", message: "Ola1!"})
t({message: "Ola2!" context: "My Context"})
```

Generated output would be:
```po
#, explicit-id (this will affect interpetation)
msgid "my.custom.id"
msgstr "Ola1!"

#. js-lingui-id: <hash> (this is just for context)
msgctxt "My Context"
msgid "Ola2!"
msgstr ""
```

Explicit id passed from the code already mapped to `msgid` in current implementation. We need to somehow differentiate where id is explicit or implicit (generated). 

When we read catalog, if we see flag "explicit-id" we treat this id as-is. If we don't see such flag we take `msgctxt` and `msgid` and produce hashed id from them.

andrii-bodnar:
@thekip got you, thanks!

The latest example looks good to me

thekip:
SWC plugin ready!

thekip:
@andrii-bodnar 
>  have that one suggestion about the new package

Have an opinion that `utils`, `helpers` etc is kinda antipattern because after some time it becomes a place for everything what we haven't found a right place for. 

I was able to make it a bit cleaner. I put this function into @lingui/cli/api and use it from there.

andrii-bodnar:
> becomes a place for everything what we haven't found a right place for

yes, it's possible and a new method in this case should be added only in case of highly justified need

> I was able to make it a bit cleaner. I put this function into @lingui/cli/api and use it from there.

It's even better I think

andrii-bodnar:
@thekip @Martin005 I just realized that the `msgstr` in the [latest example](https://github.com/lingui/js-lingui/pull/1440#issuecomment-1433012607) for a string with context needs to be filled also by the string, otherwise it becomes unprocessable by localization tools

```yml
#, explicit-id
msgid "my.custom.id"
msgstr "Ola1!"

#. js-lingui-id: <hash>
msgctxt "My Context"
msgid "Ola2!"
msgstr "Ola2!" # Needs to contain the source string also
```

In the example above, the localization Tool needs to take the source string for translation from the `msgstr` node because the `msgid` is already used for ids. So if the `msgstr` is empty, it won't be imported. So we should stay consistent and fill all the `msgstr` values.

thekip:
Yesterday I thought that i was close to finish this task, but inverting logic with `explicit-id` flag revealed many issues in the other part of code which should be fixed to get this merged. 

Many of them due to the fact that while compiling catalogs lingui operates with only `key` and `translation`, where `key` is a natural language string. Many places in the code works based on that assumptions, and in many places there is a fallback to the "key" if no translation found. 

With new hashbased id, key is always a hash, and this fallback just doesn't work. Instead, i should change the logic to operate with `id`, `source string`, `translation`, and in case there are no translation fallback to source string first and to id as last resort. 

Unfortunately, there is complicated fallbacks logic, which is not fully covered by tests, so i need to refactor this first.

thekip:
@andrii-bodnar 
> In the example above, the localization Tool needs to take the source string for translation from the msgstr node because the msgid is already used for ids. So if the msgstr is empty, it won't be imported. So we should stay consistent and fill all the msgstr values.

Sorry, but it's wrong. If you fulfill `msgstr` with source message while extraction localization tool will not understand that this line does not have a translation, because translation already provided.  

What you mentioned should be done for catalog where "sourceLocale == locale", in other words if you're extracting for english and english is set as your source locale, `msgstr` should be fullfilled with source message.

thekip:
What really bother me with my latest example is the fact that @Martin005 's feature https://github.com/lingui/js-lingui/pull/1431 would not work here either. Because we still generate a hash from catalogs from a source string, and if sourcestring would differ from what we have in bundle, than translation would not work. 

I'm still thinking that moving this flattening thing to the macro may be is not so bad idea. Because it would be consistent with the rest of the code. 

It's probably could be done simpler way, without parsing/reprinting icu's, we can discuss that in discord or in the original pr

andrii-bodnar:
@thekip yes, I was talking only about the source locale, to the target locales the `msgstr` should be empty indeed

thekip:
@andrii-bodnar this is already done in the code, and actually broken by this PR, look for my previous comment about fallbacks. I'm going to fix it.

andrii-bodnar:
@thekip fyi, there is one issue regarding the locales fallback - https://github.com/lingui/js-lingui/issues/1167 (I'm not sure if it's directly related to the mentioned fallbacks logic but maybe can be considered also)

thekip:
waiting for merging this https://github.com/lingui/js-lingui/pull/1443 after rebasing test suite would pass

packages/macro/src/macroJs.ts
```diff
@@ -222,78 +207,80 @@ export default class MacroJs {
...
+      }
 
-    let messageNode = node.value
-    if (tokens != null) {
-      const messageFormat = new ICUMessageFormat()
-      const { message: messageRaw, values } = messageFormat.fromTokens(tokens)
-      const message = normalizeWhitespace(messageRaw)
-      messageNode = this.types.stringLiteral(message)
+      if (!idProperty && this.types.isStringLiteral(messageNode)) {
+        // todo: context could be different types, get implementation from extractor
```
Martin005, line 246: Please act on this `todo`

packages/macro/package.json
```diff
@@ -35,7 +35,7 @@
     "@babel/runtime": "^7.20.13",
     "@babel/types": "^7.20.7",
     "@lingui/conf": "3.17.1",
-    "ramda": "^0.27.1"
+    "@lingui/cli": "3.17.1"
```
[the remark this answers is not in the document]
thekip: - `@lingui/core` -  api for **runtime** things, to run in browser
- `@lingui/cli/api` - api for **compile-time** things, to run in cli

So no, I don't think this import is weird.

packages/macro/src/macroJsx.ts
```diff
@@ -127,7 +131,7 @@ export default class MacroJSX {
       )
     }
 
-    if (context) {
+    if (!this.stripNonEssentialProps && context) {
```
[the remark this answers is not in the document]
thekip: done

packages/macro/test/jsx-trans.ts
```diff
@@ -403,29 +452,46 @@ const cases: TestCase[] = [
...
+              />
+            ),
+          }}
+        />;
+
       `,
   },
   {
     name: "Use a js macro inside a JSX Attribute of a non macro JSX component",
+    stripId: true,
```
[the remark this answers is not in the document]
thekip: done

packages/macro/src/macroJs.ts
thekip: done, now it supports template literal as well.

packages/macro/test/js-select.ts
```diff
@@ -16,13 +16,24 @@ const cases: TestCase[] = [
...
+              numOfGuests: numOfGuests,
+            },
+          }
+        );
       `,
   },
   {
+    // todo: the original test case is weird.
+    //  The `comment field` was never escaped here
+    //   and never allowed as property
```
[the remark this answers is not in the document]
thekip: done

packages/macro/src/macroJsx.ts
```diff
@@ -101,21 +103,23 @@ export default class MacroJSX {
...
-        attributes.push(this.safeJsxAttribute(MESSAGE, message))
-      }
     } else {
-      attributes.push(this.safeJsxAttribute(ID, message))
+      attributes.push(
+        this.createStringJsxAttribute(ID, generateMessageId(message, context))
+      )
+    }
+
+    if (!this.stripNonEssentialProps && message) {
```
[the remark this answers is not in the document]
thekip: done

thekip:
Found a critical issue (thanks to new e2e test for extractor!), context from macro might be in different falsy forms such as `null` or `undefined`. That forms produced different results in `generateMessageId()`. Now fixed, e2e test + unit test for function added.

thekip:
image
i love development under windows 💩!

thekip:
Guys, if it's ready, let's merge it and i will do docs update in next pr. This one is already huge.

packages/cli/src/api/generateMessageId.test.ts
```diff
@@ -0,0 +1,32 @@
...
+    const withPrefix = generateMessageId(context + "my message")
+
+    expect(withContext != withSuffix).toBeTruthy()
+    expect(withContext != withPrefix).toBeTruthy()
+  })
+
+  it("All kind of falsy context should give the same result", () => {
+    const expected = `vQhkQx`
+    expect(generateMessageId("my message", "")).toBe(expected)
+    expect(generateMessageId("my message", "")).toBe(expected)
```
Martin005, line 28: Why are you testing here the same thing twice?

packages/cli/src/test/compile.test.ts
```diff
@@ -136,36 +112,40 @@ msgstr ""
       }
     )
 
-    it("Should show error and stop compilation of catalog if message doesnt have a translation (with template)", () => {
-      mockFs({
-        "/test": {
-          "messages.pot": `
+    it(
+      "Should show error and stop compilation of catalog " +
+        " if message doesnt have a translation (with template)",
```
Martin005, line 117:
```suggestion
        "if message doesnt have a translation (with template)",
```

andrii-bodnar:
Let's be consistent, I don't see any problems with adding the docs as a separate commit

thekip:
> What really bother me with my latest example is the fact that @Martin005 's feature https://github.com/lingui/js-lingui/pull/1431 would not work here either. Because we still generate a hash from catalogs from a source string, and if sourcestring would differ from what we have in bundle, than translation would not work.

> I'm still thinking that moving this flattening thing to the macro may be is not so bad idea. Because it would be consistent with the rest of the code.

> It's probably could be done simpler way, without parsing/reprinting icu's, we can discuss that in discord or in the related pr

Guys, i want your reaction on that until it's not too late.

website/docs/guides/plurals.md
```diff
@@ -21,7 +21,7 @@ Good thing is that **as developers, we have to know only plural forms for the so
 
 If we use English in the source code, then we'll use only *one* and *other*:
 
-``` js
```
thekip, line 24: Syntax highlighting in the intelij based IDEs doesn't work when there is a space.

Martin005:
@thekip We have agreed with @andrii-bodnar that it would be best to postpone the flattening in extraction (drafted in #1431) for now.

thekip:
@Martin005 actually to fix issue with flattening we can make `js-lingui-id` required in po files. And don't calculate id when reading from (message + context) but only take this hash. Something like this is made for po-gettext extractor. Where comments used to store original icu components. 

```po
#. js-lingui-id: <hash> 
msgctxt "..."
msgid "..."
msgstr ""
```

website/docs/ref/macro.md
```diff
@@ -517,19 +612,20 @@ Use `select` inside [`t`](#t) macro if you want to add custom `id` or `comment`
...
-  comment?: string
+  comment?: string,
+  context?: string,
 }
 
 defineMessage(message: MessageDescriptor)
 ```
 
-Either `id` or `message` property is required. `id` is a custom message id. If it isn't set, the `message` is used instead.
+Either `id` or `message` property is required. `id` is a custom message id. If it isn't set, the `message` is used for generating an id.
```
Martin005, line 626:
```suggestion
Either `id` or `message` property is required. `id` is a custom message ID. If it isn't set, the `message` (and `context` if provided) are used for generating an ID.
```

website/docs/ref/macro.md
```diff
@@ -600,33 +698,41 @@ const message = defineMessage({
...
 
-All macros share following props:
+[`Trans`](/docs/ref/react.md#trans) is the basic macro for static messages, messages with variables, but also for messages with inline markup:
 
 #### `id`
 
 Each message in catalog is identified by **message ID**.
 
-While all macros use generated message as the ID, it's possible to override it. In such case, generated message is used as a default translation.
+While macro use message to generate ID, it's possible to override it.
```
Martin005, line 727:
```suggestion
While macro uses message (and `context` property if provided) to generate ID, it's possible to override it.
```

website/docs/ref/macro.md
```diff
@@ -241,59 +275,73 @@ By default, the `i18n` object is imported from `@lingui/core`. If you use a cust
...
+You don't need to provide ID manually. Macro will automatically create a short id from your message.
+
+#### `context`
+
+The same text elements with different context are extracted with different IDs. For example, if the word "right" uses the following two definitions in two different locations, the word is translated differently and merged back into the application as different translation entries.
+
+- correct as in "you are right"
+- direction as in "turn right"
+
+If the same text elements meet the following conditions, the text elements are extracted only once and use the same ID.
```
Martin005, line 298: When reading the docs, this section doesn't look it belongs here – the `context` property is not metioned in the `MessageDescriptor` type above.

website/docs/ref/macro.md
```diff
@@ -637,47 +743,60 @@ Comment for translators to give them additional context about the message. It's
...
 <Trans>Refresh inbox</Trans>;
 
 // ↓ ↓ ↓ ↓ ↓ ↓
-import { Trans } from "@lingui/react"
-<Trans id="Refresh inbox" />
+import { Trans } from "@lingui/react";
+<Trans id={"EsCV2T"} message="Refresh inbox" />;
+```
+
+Generates different ID when `context` is provided:
```
Martin005, line 764:
```suggestion
Lingui generates different IDs when `context` is provided:
```

website/docs/ref/macro.md
```diff
@@ -811,3 +952,16 @@ import { Select } from "@lingui/macro"
     other="Their book"
 />
 ```
+
+Use `<Select>` inside `<Trans>` macro if you want to provide `id`, `context` or `comment`
```
Martin005, line 956:
```suggestion
Use `<Select>` inside `<Trans>` macro if you want to provide `id`, `context` or `comment`.
```

website/docs/ref/macro.md
```diff
@@ -783,6 +910,20 @@ import { SelectOrdinal } from "@lingui/macro"
 />
 ```
 
+Use `<SelectOrdinal>` inside `<Trans>` macro if you want to provide `id`, `context` or `comment`
```
Martin005, line 913:
```suggestion
Use `<SelectOrdinal>` inside `<Trans>` macro if you want to provide `id`, `context` or `comment`.
```

website/docs/ref/macro.md
```diff
@@ -748,6 +867,14 @@ import { Plural } from "@lingui/macro"
 */
 ```
 
+Use `<Plural>` inside `<Trans>` macro if you want to provide `id`, `context` or `comment`
```
Martin005, line 870:
```suggestion
Use `<Plural>` inside `<Trans>` macro if you want to provide `id`, `context` or `comment`.
```

website/docs/tutorials/react-patterns.md
```diff
@@ -23,15 +23,32 @@ function render() {
 
 You don't need anything special to use [`Trans`](/docs/ref/macro.md#trans) inside your app (except of wrapping the root component in [`I18nProvider`](/docs/ref/react.md#i18nprovider)).
 
+## Choosing between Generated and Explicit ID
```
Martin005, line 26:
```suggestion
## Choosing between generated and explicit ID
```

website/docs/tutorials/react-patterns.md
```diff
@@ -23,15 +23,32 @@ function render() {
...
+
+The both approaches has they pros and cons. We recommend using generated ids as it scales and gives better developer experience.
```
Martin005, line 28:
```suggestion
Both approaches have their pros and cons. We recommend using generated IDs as it scales and gives better developer experience.
```

website/docs/ref/macro.md
```diff
@@ -50,6 +50,7 @@ The advantages of using macros are:
 
 - You don't need to learn ICU MessageFormat syntax. You always use familiar JS and JSX code.
 - Components and functions are type checked.
+- Short id generated for your messages
```
Martin005, line 53:
```suggestion
- Short ID generated for your messages
```

website/docs/ref/macro.md
```diff
@@ -241,59 +275,73 @@ By default, the `i18n` object is imported from `@lingui/core`. If you use a cust
...
+   message?: string,
+   values?: Record<string, any>,
    comment?: string
 }
 ```
 
-`id` is message ID and the only required parameter. `id` and `message` are extracted to message catalog. Only `id`, `values`, and `formats` are used at runtime, all other attributes are removed from production code for size optimization.
+`id` is message ID and the only required parameter. `id` and `message` are extracted to message catalog. Only `id` and `values` are used at runtime, all other attributes are removed from production code for size optimization.
+
+You don't need to provide ID manually. Macro will automatically create a short id from your message.
```
Martin005, line 289:
```suggestion
You don't need to provide ID manually. Macro will automatically create a short ID from your message.
```

website/docs/ref/macro.md
```diff
@@ -600,33 +698,41 @@ const message = defineMessage({
...
 
 const message = /*i18n*/{
-  context: "Context about the link",
   id: "msg.navigation.about"
 }
 ```
 
 `message` and `comment` are used in message catalogs only.
+
+`context` is used only for id generation and stripped from output.
```
Martin005, line 707:
```suggestion
`context` is used only for generating ID and is stripped from the output.
```

website/docs/tutorials/react-patterns.md
```diff
@@ -23,15 +23,32 @@ function render() {
...
+
 ### Using ID generated from message
 
 #### With [`Trans`](/docs/ref/macro.md#trans)
 
-In the examples above, the content of [`Trans`](/docs/ref/macro.md#trans) is transformed into message in MessageFormat syntax. By default, this message is used as the message ID. Considering the example above, messages `LinguiJS example` and `Hello <0>{name}</0>.` are extracted and used as IDs.
+In the examples above, the content of [`Trans`](/docs/ref/macro.md#trans) is transformed into message in MessageFormat syntax. By default, this message is used for generating ID. Considering the example above, catalog would be fulfilled by these entries:
```
Martin005, line 34:
```suggestion
In the example code above, the content of [`Trans`](/docs/ref/macro.md#trans) is transformed into message in MessageFormat syntax. By default, this message is used for generating ID. Considering the example above, catalog would be fulfilled by these entries:
```

website/docs/tutorials/react-patterns.md
```diff
@@ -41,6 +58,25 @@ export default function ImageWithCaption() {
...
+- direction as in "turn right"
+
+If the same text elements meet the following conditions, the text elements are extracted only once and use the same ID.
+
+```jsx
+import { Trans } from "@lingui/macro";
+<Trans context="direction">right</Trans>;
+<Trans context="correctness">right</Trans>;
+
+// ↓ ↓ ↓ ↓ ↓ ↓
```
Martin005, line 74:
```suggestion
// ↓ ↓ ↓ ↓ ↓ ↓

```

website/docs/tutorials/react-patterns.md
```diff
@@ -91,10 +127,6 @@ export default function ImageWithCaption({ count }) {
 }
 ```
 
-:::tip
-It's a good practice to set explicit IDs for phrases to make it easier to identify phrases out of context and to track where they're used. IDs usually follow a naming scheme that includes *where* the phrase is used.
-:::
-
```
Martin005, line 97: Why have you removed this section?

website/docs/ref/macro.md
```diff
@@ -178,47 +194,65 @@ customI18n._(/*i18n*/{
...
 
 ### Examples of JSX macros
 
-``` js
+```jsx
 <Trans>Attachment {name} saved</Trans>
 
-// ↓ ↓ ↓ ↓ ↓ ↓
-
+// ↓ ↓ ↓ ↓ ↓ ↓,
```
Martin005, line 233:
```suggestion
// ↓ ↓ ↓ ↓ ↓ ↓

```

website/docs/ref/macro.md
```diff
@@ -600,33 +698,41 @@ const message = defineMessage({
...
+While macro use message to generate ID, it's possible to override it.
 
-``` jsx
-import { Trans } from "@lingui/macro"
-<Trans id="message.attachment_saved">Attachment {name} saved.</Trans>
+```jsx
+import { Trans } from "@lingui/macro";
+<Trans id="message.attachment_saved">Attachment {name} saved.</Trans>;
 
 // ↓ ↓ ↓ ↓ ↓ ↓
```
Martin005, line 733:
```suggestion
// ↓ ↓ ↓ ↓ ↓ ↓

```

website/docs/ref/macro.md
```diff
@@ -637,47 +743,60 @@ Comment for translators to give them additional context about the message. It's
...
 
-``` jsx
-import { Trans } from "@lingui/macro"
+If the same text elements meet the following conditions, the text elements are extracted only once and use the same ID.
+
+```jsx
+import { Trans } from "@lingui/macro";
 <Trans>Refresh inbox</Trans>;
 
 // ↓ ↓ ↓ ↓ ↓ ↓
```
Martin005, line 759:
```suggestion
// ↓ ↓ ↓ ↓ ↓ ↓

```

website/docs/ref/macro.md
```diff
@@ -637,47 +743,60 @@ Comment for translators to give them additional context about the message. It's
...
+```
+
+Generates different ID when `context` is provided:
+
+```jsx
+import { Trans } from "@lingui/macro";
+<Trans context="direction">right</Trans>;
+<Trans context="correctness">right</Trans>;
+
+// ↓ ↓ ↓ ↓ ↓ ↓
```
Martin005, line 771:
```suggestion
// ↓ ↓ ↓ ↓ ↓ ↓

```

website/docs/ref/macro.md
```diff
@@ -705,21 +824,21 @@ Components and HTML tags are replaced with dummy indexed tags (`<0></0>`) which
...
 Props of [`Plural`](/docs/ref/macro.md#plural-1) macro are transformed into [`plural`](/docs/ref/message-format.md) format.
 
-``` jsx
-import { Plural } from "@lingui/macro"
-<Plural value={numBooks} one="Book" other="Books" />
+```jsx
+import { Plural } from "@lingui/macro";
+<Plural value={numBooks} one="Book" other="Books" />;
 
 // ↓ ↓ ↓ ↓ ↓ ↓
```
Martin005, line 831:
```suggestion
// ↓ ↓ ↓ ↓ ↓ ↓

```

website/docs/guides/plurals.md
```diff
@@ -54,20 +59,22 @@ The important thing is that *we don't need to change our code to support languag
...
-       '{numBooks, plural, one {# book} other {# books}}',
-       { numBooks }
-    )
+    ```js
+    i18n._({
+      id: 'd1wX4r',
+      // stripped on production
+      // message: '{numBooks, plural, one {# book} other {# books}}',
+      values:  { numBooks }
+    })
```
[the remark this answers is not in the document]
thekip: Check the text above this snippet. I'm intentionally commented it out.

website/docs/ref/macro.md
thekip: Change it as you wish.

thekip:
Guys, everything related to docs, just do it yourself. I'm just not able to do it all by myself.

andrii-bodnar:
@thekip please just resolve the current suggestions and then we'll take care of its polishing

packages/cli/src/api/formats/po-gettext.test.ts
```diff
@@ -296,14 +126,15 @@ msgid_plural "# days"
 msgstr[0] "# den"
 msgstr[1] "# dny"
 msgstr[2] "# dní"
-`
+  `
```
andrii-bodnar, line 129: is it ok?

packages/macro/test/jsx-trans.ts
```diff
@@ -46,7 +48,7 @@ const cases: TestCase[] = [
       `,
   },
   {
-    name: "Macro with custom ID (template expression)",
+    name: "Preserve custom ID  (template expression)",
```
andrii-bodnar, line 51:
```suggestion
    name: "Preserve custom ID (template expression)",
```

packages/macro/test/jsx-trans.ts
```diff
@@ -91,29 +93,39 @@ const cases: TestCase[] = [
...
         <Trans>Hi {yourName}, my name is {myName}</Trans>;
       `,
     expected: `
         import { Trans } from "@lingui/react";
-        <Trans id={"Hi {yourName}, my name is {myName}"} values={{
-          yourName: yourName,
-          myName: myName,
-        }} />;
+        <Trans 
+            id={"<stripped>"}
```
andrii-bodnar, line 104: indentations?

thekip: if you want, you can polish all indentations by yourself, i don't have time for that

packages/cli/src/api/formats/po-gettext.test.ts
```diff
@@ -251,25 +81,25 @@ describe("po-gettext format", () => {
...
-          one {{numArticles, plural,
-            one {1 book and 1 article}
-            other {1 book and {numArticles} articles}
-          }}
-          other {{numArticles, plural,
-            one {{numBooks} books and 1 article}
-            other {{numBooks} books and {numArticles} articles}
-          }}
-        }`,
+            one {{numArticles, plural,
```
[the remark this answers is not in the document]
thekip: if you want, you can polish all indentations by yourself, i don't have time for that

website/docs/ref/macro.md
```diff
@@ -50,6 +50,7 @@ The advantages of using macros are:
 
 - You don't need to learn ICU MessageFormat syntax. You always use familiar JS and JSX code.
 - Components and functions are type checked.
+- Short ID generated for your messages
```
andrii-bodnar, line 53:
```suggestion
- Short ID generated for your messages.
```

website/docs/ref/macro.md
```diff
@@ -241,59 +276,73 @@ By default, the `i18n` object is imported from `@lingui/core`. If you use a cust
...
-   formats?: Object,
+   id: string,
+   message?: string,
+   values?: Record<string, any>,
    comment?: string
 }
 ```
 
-`id` is message ID and the only required parameter. `id` and `message` are extracted to message catalog. Only `id`, `values`, and `formats` are used at runtime, all other attributes are removed from production code for size optimization.
+`id` is message ID and the only required parameter. `id` and `message` are extracted to message catalog. Only `id` and `values` are used at runtime, all other attributes are removed from production code for size optimization.
```
andrii-bodnar, line 288:
```suggestion
`id` is the message ID and the only required parameter. `id` and `message` are extracted to the message catalog. Only `id` and `values` are used at runtime, all other attributes are removed from production code for size optimization.
```

website/docs/ref/macro.md
```diff
@@ -241,59 +276,73 @@ By default, the `i18n` object is imported from `@lingui/core`. If you use a cust
...
+
+You don't need to provide ID manually. Macro will automatically create a short ID from your message.
```
andrii-bodnar, line 290:
```suggestion
You don't need to provide your ID manually. Macro will automatically create a short ID from your message.
```

website/docs/ref/macro.md
```diff
@@ -241,59 +276,73 @@ By default, the `i18n` object is imported from `@lingui/core`. If you use a cust
...
+
+#### `context`
+
+The same text elements with different context are extracted with different IDs. For example, if the word "right" uses the following two definitions in two different locations, the word is translated differently and merged back into the application as different translation entries.
```
andrii-bodnar, line 294:
```suggestion
The same text elements with different contexts are extracted with different IDs. For example, if the word "right" uses the following two definitions in two different locations, the word is translated differently and merged back into the application as different translation entries.
```

website/docs/ref/macro.md
```diff
@@ -438,27 +495,38 @@ i18n._(`{numBooks, plural,
...
+Use `plural` inside [`t`](#t) or [`defineMessage`](#definemessage) macro if you want to add custom `id`, `context` or `comment` for translators.
+
+```js
+const message = t({
+   id: "my.custom.id",
+   comment: "My Comment",
+   message: plural(count, {
+     one: "# Book",
+     other: "# Books"
+   })
```
andrii-bodnar, line 516:
```suggestion
  id: "my.custom.id",
  comment: "My Comment",
  message: plural(count, {
    one: "# Book",
    other: "# Books"
  })
```

website/docs/ref/macro.md
```diff
@@ -471,25 +539,39 @@ const message = selectOrdinal(count, {
...
+const message = t({
+   id: "my.custom.id",
+   comment: "My Comment",
+   message: selectOrdinal(count, {
+     one: "#st",
+     two: "#nd",
+     few: "#rd",
+     other: "#th",
+   })
+})
```
andrii-bodnar, line 562:
```suggestion
  id: "my.custom.id",
  comment: "My Comment",
  message: selectOrdinal(count, {
    one: "#st",
    two: "#nd",
    few: "#rd",
    other: "#th",
  })
})
```

website/docs/ref/macro.md
```diff
@@ -500,15 +582,29 @@ const message = select(gender, {
...
+
+```js
+const message = t({
+   id: "my.custom.id",
+   comment: "My Comment",
+   message: select(gender, {
+     male: "he",
+     female: "she",
+     other: "they"
+   })
```
andrii-bodnar, line 605:
```suggestion
  id: "my.custom.id",
  comment: "My Comment",
  message: select(gender, {
    male: "he",
    female: "she",
    other: "they"
  })
```

website/docs/tutorials/react-patterns.md
andrii-bodnar: moved this to the section above

Martin005 (approved):

andrii-bodnar (approved):

Issue #1416, opened 2023-02-08
different config loaded for `macro` and `cli` when use `extract --config` if config file is under `cwd/subfolder`


taozhou-glean:
**Describe the bug**

we saw `No locales found!` but message extracted successful:

```
No locales defined!

Add 'locales' to your configuration. See https://lingui.dev/ref/conf#locales
Catalog statistics for typescript/core/i18n/locales/{locale}/messages:
┌────────────────────┬─────────────┬─────────┐
│ Language           │ Total count │ Missing │
├────────────────────┼─────────────┼─────────┤
│ en (source)        │     102     │    -    │
│ pseudolocalization │     102     │   102   │
└────────────────────┴─────────────┴─────────┘
```

after debugging it, it looks both `cli/lingui-extract` and `macro` will try to load config separately:

```
    at getConfig (/xxx/node_modules/@lingui/conf/build/index.js:90:10)
    at Object.<anonymous> (/xxx/node_modules/@lingui/cli/build/lingui-extract.js:80:38)
```
vs
```
    at getConfig (/xxx/node_modules/@lingui/conf/build/index.js:90:10)
    at Object.<anonymous> (/xxx/node_modules/@lingui/macro/build/index.js:20:36)
```

from the [code](https://github.com/lingui/js-lingui/blob/8bc212d2846af5609d89f51efed952b089244e4e/packages/conf/src/index.ts#L132), it looks like the `macro` only tries to find the config file from cwd and up, and `cli` will consider the `--config` param.

The issue is that our set up is more like following:

```
- package.json
  - project
    - lingui.config.ts
```

so it won't find anything if starts from the level of `package.json`.

the workaround is to use `LINGUI_CONFIG=project/lingui.config.ts`, but it would be more ideal if `--config` also just works.

thekip:
Yes, they not depend on each other. The simplest solution would be just set `process.env.LINGUI_CONFIG = argv. config` inside the CLI command. But good solution is upcoming with fundamental changes in v4

taozhou-glean:
similar issue observed when using the loader, which is taking the config path from loader option or using the `resourcePath` as `cwd` to search for config file.. and it does not take `process.env. LINGUI_CONFIG` :( so have to pass the `config` with webpack config. Probably should at least honor `LINGUI_CONFIG`

PR #1442, opened 2023-02-16
refactor: pass config directly to macro

thekip:
# Description

Extracted non-related changes from https://github.com/lingui/js-lingui/pull/1440

Before this change config was reading separately in cli and macro. Read issue: https://github.com/lingui/js-lingui/issues/1416

Now config passed down to the macro from cli. This greatly helps in testing, because you can pass config to command and everything work just fine.

Other changes: 

- I extracted normalization for `runtimeConfigModule` into `conf` package and wrote a separate test. So now macro is not responsible for this normalization.

## Types of changes

- [ ] Documentation update

Fixes https://github.com/lingui/js-lingui/issues/1416

## Checklist

### DIFF
--- a/packages/conf/src/migrations/normalizeRuntimeConfigModule.ts
+++ b/packages/conf/src/migrations/normalizeRuntimeConfigModule.ts
@@ -1,0 +1,41 @@
+import { LinguiConfig, LinguiConfigNormalized } from "../types"
+
+type ModuleSrc = [source: string, identifier?: string]
+
+const getSymbolSource = (
+  defaults: ModuleSrc,
+  config: LinguiConfig["runtimeConfigModule"]
+): ModuleSrc => {
+  const name = defaults[1]
+  if (Array.isArray(config)) {
+    if (name === "i18n") {
+      return config
+    }
+    return defaults
+  }
+
+  return config[name] || defaults
+}
+
+export function normalizeRuntimeConfigModule(
+  config: Pick<LinguiConfig, "runtimeConfigModule">
+) {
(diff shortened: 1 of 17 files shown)


packages/macro/src/index.ts
```diff
@@ -49,9 +27,28 @@ const jsMacroTags = new Set([
...
 
+let config: LinguiConfigNormalized
+
+function getConfig(_config?: LinguiConfigNormalized) {
+  if (_config) {
+    config = _config
+  }
+  if (!config) {
+    config = loadConfig({ configPath: process.env.LINGUI_CONFIG })
+  }
```
thekip, line 38: if config passed directly to macro - use it
if no, use try to load config. Then store it in global variable to avoid loading it on every macro call

packages/conf/src/migrations/normalizeRuntimeConfigModule.ts
```diff
@@ -0,0 +1,41 @@
+import { LinguiConfig, LinguiConfigNormalized } from "../types"
+
+type ModuleSrc = [source: string, identifier?: string]
+
+const getSymbolSource = (
+  defaults: ModuleSrc,
+  config: LinguiConfig["runtimeConfigModule"]
```
Martin005, line 7: Maybe it would be better to name this variable `runtimeConfigModule` instead of `config` as only the `runtimeConfigModule` property is passed to this function and not the whole config.

Martin005 (approved):

thekip:
Rebased

andrii-bodnar (approved):

PR #1443, opened 2023-02-17
refactor: getTranslations, prepare for hash-based ids

thekip:
# Description

https://github.com/lingui/js-lingui/pull/1440#issuecomment-1434353740

> Yesterday I thought that i was close to finish this task, but inverting logic with explicit-id flag revealed many issues in the other part of code which should be fixed to get this merged.
>
> Many of them due to the fact that while compiling catalogs lingui operates with only key and translation, where key is a natural language string. Many places in the code works based on that assumptions, and in many places there is a fallback to the "key" if no translation found.
>
> With new hashbased id, key is always a hash, and this fallback just doesn't work. Instead, i should change the logic to operate with id, source string, translation, and in case there are no translation fallback to source string first and to id as last resort.
>
> Unfortunately, there is complicated fallbacks logic, which is not fully covered by tests, so i need to refactor this first.

## Types of changes

- [ ] Documentation update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/cli/src/api/getTranslationsForCatalog.test.ts
+++ b/packages/cli/src/api/getTranslationsForCatalog.test.ts
@@ -1,0 +1,341 @@
+import { getTranslationsForCatalog } from "./getTranslationsForCatalog"
+import { AllCatalogsType, Catalog, CatalogType } from "./catalog"
+
+function getCatalogStub(
+  catalogs: AllCatalogsType,
+  template: CatalogType = {}
+): Catalog {
+  const catalogStub: Partial<Catalog> = {
+    readAll(): AllCatalogsType {
+      return catalogs
+    },
+
+    readTemplate(): CatalogType {
+      return template
+    },
+  }
+
+  return catalogStub as Catalog
+}
+
+function lang(
+  locale: string,
+  messages: Array<(locale: string) => CatalogType>
+) {
+  return {
+    [locale]: messages.reduce((acc, msgFn) => {
+      return {
+        ...acc,
+        ...msgFn(locale),
+      }
+    }, {}),
+  }
+}
+
+function message(id: string, source: string, noTranslation = false) {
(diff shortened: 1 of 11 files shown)


packages/cli/src/api/getTranslationsForCatalog.test.ts
```diff
@@ -0,0 +1,337 @@
...
+    })
+  })
+
+  it("Should take source messages when requested locale is sourceLocale", () => {
+    // prettier-ignore
+    const catalogStub = getCatalogStub({
+      ...lang("en", [
+        message("hashid1", "Lorem", true),
+        message("hashid2", "Ipsum", true)
+      ])
```
Martin005, line 184: I think you should add translated messages to the catalog to really test that `getTranslationsForCatalog` returns "**source messages** if requested locale is sourceLocale" as the test name says:

```suggestion
      ...lang("en", [
        message("hashid1", "Lorem", true),
        message("hashid2", "Ipsum", true)
      ]),
      
      ...lang("pl", [
        message("hashid1", "Lorem"),
        message("hashid2", "Ipsum")
      ])
```

Martin005 (commented):
Great job, thanks 🚀

packages/cli/src/api/getTranslationsForCatalog.test.ts
thekip: description is a bit misleading,  actually it doesn't matter for us would i take from source messages or translations. Because for source locale source == translation.  

Extractor always populates translations for sourceLocale. What is really matter is that when sourceLocale doesn't have a translation, we should fallback to the source message and don't call "missing" callback

i will change a description

thekip: changed

Martin005 (approved):

andrii-bodnar (approved):

Issue #1454, opened 2023-02-21
`Invalid 'target_languages': pseudo.` while sync with Translation.io 


taozhou-glean:
**Describe the bug**

```
Synchronization with Translation.io failed: Invalid 'target_languages': pseudo. Please use existing language codes (https://translation.io/docs/languages) or custom languages (https://translation.io/blog/custom-languages).
```

PR #1455, opened 2023-02-21
fix: exclude pseudo locale during sync 

taozhou-glean:
# Description

Looks like we only exclude pseudo while `init` but not `sync`, might be just an oversight, as I was touching it, so I also added the types for the config object.

## Types of changes

- [ ] Documentation update

Fixes https://github.com/lingui/js-lingui/issues/1454

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works: no tests exists for services yet

### DIFF
--- a/packages/cli/src/services/translationIO.ts
+++ b/packages/cli/src/services/translationIO.ts
@@ -4,7 +4,7 @@ import PO from "pofile"
 import https from "https"
 import glob from "glob"
 import { format as formatDate } from "date-fns"
-import { LinguiConfig } from "@lingui/conf"
+import { LinguiConfigNormalized } from "@lingui/conf"
 
 const getCreateHeaders = (language: string) => ({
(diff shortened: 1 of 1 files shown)


packages/cli/src/services/translationIO.ts
```diff
@@ -14,8 +15,16 @@ const getCreateHeaders = (language) => ({
   Language: language,
 })
 
+const getTargetLocales = (config: LinguiConfig) => {
```
thekip, line 18: Please use `LinguiConfigNormalized` type here, and in all other cases in this file

thekip (changes requested):

packages/cli/src/services/translationIO.ts
taozhou-glean: Done

Martin005 (approved):

andrii-bodnar:
@taozhou-glean thank you!

Issue #1361, opened 2023-01-23
RFC: Split macro to vanilla js + react


thekip:
Current macro assuming usage only with React. But there are cases where people want to use it with vanilla js or even different framework. 

Current macro depending on `@types/react` and `@lingui/react` which brings react to the host project which might be unwanted behavior. 

There was an attempt to fix this here https://github.com/lingui/js-lingui/issues/936
But the solution is far from perfect. 

I propose to split macro into two different packages: 

`@lingui/js-macro` (or just macro)
`@lingui/react-macro`

This in future would allow us to adopt other jsx based-frameworks.

An automatic codemod could be written to help people migrate to the new version. 

**Related to:**
- #708

andrii-bodnar:
This is a good idea, I already thought exactly about a Vue package and other frameworks.

I think we can follow the existing naming:
- `@lingui/macro`
- `@lingui/react`
- `@lingui/vue`
- what about React Native?
- etc.

Found one old issue related to this topic - #708 (Added to the issue description)

thekip:
For macro to work, it should end on `macro` 
https://github.com/kentcdodds/babel-plugin-macros/blob/main/other/docs/author.md#filename

So `@lingui/react/macro` or `@lingui/react-macro` but not `@lingui/macro/react`

tricoder42:
I definitely like this idea. I always wanted to make this library framework-agnostic. Using single macro package for only for simplicity, but I understand it can be annoying.

JSteunou:
I currently need this feature. I'm in a Vue project and I do not want to install the all React ecosystem. How can I help? Is there some plan I can follow?

thekip:
I think we can plan that for lingui v5. I see this in the following manner:

The actual transform of code would be kept as-is, e.q. one transform process both js and jsx (react) nodes. The only changes would be in typescript typings. 

Lingui would have packages called:

- `@lingui/macro` - binding for babel-macro-plugin + typings for `t` `plural` `select` and `selectOrdinal` macros.
- `@lingui/react-macro` - everything from `@lingui/macro` + typings for `Trans` `Plural` `Select` `useLingui`, etc. So react users does not need to import from 2 packages, everything from  `@lingui/macro` is re-exported from react package. 
- `@lingui/babel-macro-plugin` - the actual babel plugin which do transformations, used under the hood in two above packages. Could be even marked as optional peer dependency, but need to test it.

JSteunou:
[...]

PR #1461, opened 2023-02-23
refactor(macro): split js (core) and jsx (react) macro

thekip:
https://github.com/lingui/js-lingui/issues/1361

@Martin005 @andrii-bodnar please take a look here. I want to get feedback before i continue.

Overview: 
```mermaid
graph LR
    core("@lingui/core/macro") --> lib("@lingui/macro-lib")
    react("@lingui/react/macro") --> lib
    macro("@lingui/macro") --> react
    macro --> core

```


andrii-bodnar:
Looks interesting, but I don't really like the nascence of the new package, can we avoid it? Previously we discussed it and agreed that it was possible

thekip:
> Looks interesting, but I don't really like the nascence of the new package, can we avoid it? Previously we discussed it and agreed that it was possible

No. Look at the graph i wrote in the description. 

Why you so don't like introducing new packages? It's the way how it works.  It's internal and nobody should care. 
Look at the [formatjs](https://github.com/formatjs/formatjs/tree/main/packages) or [jest](https://github.com/facebook/jest/tree/main/packages) packages.

andrii-bodnar:
Sure, I saw the graph. I can't say that I don't like new packages, but I prefer not to introduce new "entities" if possible and there is no urgent need

If it's inevitable, it'll be so 🤷‍♂️

andrii-bodnar:
The most I worrying here is the potential `@lingui/code` bundle size increase

thekip:
There would be no bundle size increase (the macro is not bundled), but installation size of `@lingui/core`and `@lingui/react` would increase. For users who use macro nothing will be changed, installation size of all packages would be the same. For users who not using macro - yes, now macro and babel are always installed. But frankly speaking, i would not worry too much about non-macro usage, the main use case is with macro and we should concentrate our effort to make it highly polished.

thekip:
i don't know why, but i don't like what happened here 😬

We indeed need to split macro for different frameworks, but the idea that all babel dependencies would be installed even if you are not using them makes me nervous. 

Also i don't understand how we should organize tests in the repo. Now they live in one place, and we have cases where jsx + and js macro tested together. How this should be done after split i don't know.

andrii-bodnar:
Let's postpone it to the future major release (v5 or v6), maybe later we will have a better understanding of how to properly do this

PR #1463, opened 2023-02-23
feat(macro): support renamed identifiers in macro

thekip:
# Description

Support for all macros.: 

```ts
import { t as t2 } from '@lingui/macro';
const a = t2`Message`;
```

Looks like this is a rare case, because no one has raised the issue so far. But that is quite common case if code after other transpiler would be passed to lingui (custom extractor case)

## Types of changes

- [ ] Documentation update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/macro/src/index.ts
+++ b/packages/macro/src/index.ts
@@ -6,8 +6,10 @@ import MacroJSX from "./macroJsx"
 import { NodePath } from "@babel/traverse"
 import {
   ImportDeclaration,
+  Identifier,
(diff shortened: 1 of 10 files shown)


andrii-bodnar:
What about the SWC plugin? Does it also need this or it already supports this?

thekip:
It already supports.

andrii-bodnar (approved):

Martin005 (approved):

PR #1467, opened 2023-02-24
refactor(catalog): split to separate files, add tests

thekip:
`packages/cli/src/api/catalog.ts` was a big god module doing everything.  I slowly split it to separate atomic functions covered with unit tests to make them more resusable and testable. 

Key changes: 

1. Formatters now responsible for handling errors while reading / writing files. Means it enough to write `format.write()` or `format.read()` without additional boilerplate to handle ENOENT or EEXIST errors. Make usage cleaner. Corresponding tests added.  
3. All `getCatalogs` family methods extracted to separate file. Removed code duplication between them.
4. mergeCatalog extracted, and separate test written
5. extractFromFiles extracted, separate test not written, but it's covered from thru `Catalog.make`method
6. types extracted from `Catalog` file to separate file.


packages/cli/src/api/utils.ts
```diff
@@ -9,6 +12,15 @@ export function prettyOrigin(origins: [filename: string, line?: number][]) {
   }
 }
 
+export function replacePlaceholders(
+  input: string,
+  values: Record<string, string>
+): string {
+  return input.replace(/\{([^}]+)}/g, (m, placeholder) => {
```
Martin005, line 19: When I provide the following input:
`/foo/bar/{a{b}`
and the following values:
`{"b": "test"}`

this regex won't replace the `{b}` in the input with the string `test` as the matched part of the string is `a{b` instead of `b`.
You are testing just happy paths in the test file `utils.test.ts`.

thekip: this is obviously a configuration mistake, i'm not going to handle this case.

thekip: it's not going to be a full-powered template engine, as long as what we do is not a templates. We just need to replace one placeholders to value.

andrii-bodnar (approved):

thekip:
@Martin005

Issue #1405, opened 2023-02-06
Lingui Extract cause allocation failure


thekip:
Moved from here https://github.com/lingui/swc-plugin/issues/14

```
[biwano@localhost app]$ npx lingui extract

<--- Last few GCs --->

[181958:0x558f89923640]    67708 ms: Scavenge (reduce) 2042.5 (2081.9) -> 2042.0 (2082.2) MB, 8.5 / 0.0 ms  (average mu = 0.298, current mu = 0.315) allocation failure 
[181958:0x558f89923640]    67727 ms: Scavenge (reduce) 2042.8 (2082.2) -> 2042.2 (2082.2) MB, 11.5 / 0.0 ms  (average mu = 0.298, current mu = 0.315) allocation failure 
[181958:0x558f89923640]    67745 ms: Scavenge (reduce) 2043.4 (2082.5) -> 2042.8 (2082.8) MB, 7.6 / 0.0 ms  (average mu = 0.298, current mu = 0.315) allocation failure 


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0x558f859ead94 node::Abort() [/usr/bin/node]
 2: 0x558f858a30b6 node::OnFatalError(char const*, char const*) [/usr/bin/node]
 3: 0x558f85bd7b5a v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [/usr/bin/node]
 4: 0x558f85bd7de6 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/usr/bin/node]
 5: 0x558f85d77eb9  [/usr/bin/node]
 6: 0x558f85d77ffc  [/usr/bin/node]
 7: 0x558f85d90535 v8::internal::Heap::PerformGarbageCollection(v8::internal::GarbageCollector, v8::GCCallbackFlags) [/usr/bin/node]
 8: 0x558f85d90df4 v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [/usr/bin/node]
 9: 0x558f85d93e34 v8::internal::Heap::AllocateRawWithLightRetrySlowPath(int, v8::internal::AllocationType, v8::internal::AllocationOrigin, v8::internal::AllocationAlignment) [/usr/bin/node]
[...]

PR #1474, opened 2023-03-01
chore: use jiti for ts config loading

thekip:
# Description
 As i mentioned: 

> i've found very promising https://github.com/unjs/jiti and https://github.com/unjs/c12 we should give them a try. These packages powering a Nuxt framework.

Unfortunately we can't use `c12` because it load asynchonously but we can incorporate `jiti` instead of `cosmiconfig-typescript-loader` / `ts-node`. 

Looks like jiti was created for such usecases, where `ts-node` is not. The benefit of using jiti over `ts-node`: 

1. Supports cache (v8, require and hard cache on the file system of transpiled files)
2. Executes only this require in "context" of transpilation. 

The problem with ts-node is that once it registered it starts processing all require calls in the current VM. That caused that not only config with it's dependecies get transpiled but also other code running in the same instance after register call. 

Also all subsequent calls to `TypeScriptLoader` with `ts-node` would cause registering another instance of compiler for require module system and this is end up with [out memory at certain point](https://github.com/lingui/js-lingui/issues/1405). 

## Types of changes

- [ ] Documentation update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/conf/package.json
+++ b/packages/conf/package.json
@@ -23,12 +23,12 @@
     "build": "rimraf ./build && tsc"
   },
   "dependencies": {
-    "@babel/runtime": "^7.20.13",
     "@babel/core": "7.20.12",
+    "@babel/runtime": "^7.20.13",
     "chalk": "^4.1.0",
     "cosmiconfig": "^8.0.0",
-    "cosmiconfig-typescript-loader": "^4.3.0",
     "jest-validate": "^29.4.3",
+    "jiti": "^1.17.1",
     "lodash.get": "^4.4.2"
   },
   "files": [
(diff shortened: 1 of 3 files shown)


packages/conf/src/getConfig.ts
```diff
@@ -32,20 +30,23 @@ export function getConfig({
 } = {}): LinguiConfigNormalized {
   const defaultRootDir = cwd || process.cwd()
   const moduleName = "lingui"
+
   const configExplorer = cosmiconfigSync(moduleName, {
     searchPlaces: [
+      `${moduleName}.config.js`,
+      `${moduleName}.config.ts`,
+      `${moduleName}.config.mjs`,
```
thekip, line 38: just give them more priority, thinks it will do less `lstat` calls

packages/conf/package.json
```diff
@@ -39,17 +39,5 @@
...
-    "ts-node": ">=10",
-    "typescript": ">=4"
-  },
-  "peerDependenciesMeta": {
-    "ts-node": {
-      "optional": true
-    },
-    "typescript": {
-      "optional": true
-    }
```
thekip, line 53: Jiti claims they have zero dependencies. Don't know how that possible, because they use at least babel for transformation.

thekip: Got that, they are bundle everything with webpack.

andrii-bodnar (approved):

PR #1475, opened 2023-03-01
fix: allow passing any expression as custom i18n instance

janicduplessis:
# Description

Currently passing a custom i18n instance to the `t` macro only supports passing identifier, although any expression should be valid.

For example this currently crashes during extraction:

```ts
t(context.intl)`Hello`;
```

This changes the check from `isIdentifier` to `isExpression` so we can support function calls or member properties.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/packages/macro/src/macroJs.ts
+++ b/packages/macro/src/macroJs.ts
@@ -66,7 +66,7 @@ export default class MacroJs {
       message,
       values,
     }: { message: ParsedResult["message"]; values: ParsedResult["values"] },
-    linguiInstance?: babelTypes.Identifier
+    linguiInstance?: babelTypes.Expression
   ) => {
     const properties: ObjectProperty[] = [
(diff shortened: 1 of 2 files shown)


thekip:
@janicduplessis thanks for you contribution. We have an active development in `next` branch. Please port your changes there.

janicduplessis:
@thekip Ok, I rebased against the next branch

thekip (approved):

andrii-bodnar (approved):

andrii-bodnar:
@thekip do we need the same fix for the SWC plugin?

thekip:
I'm not sure. We need to write the same test case and check how it behaves. 

Extra point: we need somehow to write a universal test fixtures for both SWC and Babel transform, to be always sure that they behave the same

PR #1478, opened 2023-03-03
chore: update yarn to berry, fix deps

thekip:
I gave up fighting with legacy Yarn


thekip:
originally, wanted to use yarn berry only for monorepo packages, but looks like once it configured for root package.json it automatically applied for all subfolders/sub-package.json (website, examples) 

So why not

andrii-bodnar (approved):

PR #1482, opened 2023-03-06
Fix typo in migration-4.md

bravo-kernel:


website/docs/releases/migration-4.md
```diff
@@ -17,7 +17,7 @@ For most projects, it should work without extra configuration as long as it is a
 
 ```diff title="lingui.config.js"
 module.exports = {
--  extractorBabelOptions: { [...] }
+-  extractBabelOptions: { [...] }
```
thekip, line 20: There is also one line above with the same typo, could you also change this

PR #1494, opened 2023-03-09
fix(react): remove deprecated defaultProps from Trans cmp

### LINKED ISSUE #1493: Deprecation warning
There is warning in console when using `Trans` component from `@lingui/macro`

`Warning: Trans: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.`

Im using NextJs 13 and `"@lingui/macro": "^4.0.0-next.1",`

thekip:
fix #1493


packages/react/src/Trans.tsx
```diff
@@ -27,8 +27,8 @@ export function Trans(props: TransProps): React.ReactElement<any, any> | null {
   const { i18n, defaultComponent } = useLingui()
   const { render, component, id, message, formats } = props
 
-  const values = { ...props.values }
-  const components = { ...props.components }
+  const values = { ...(props.values || {}) }
+  const components = { ...(props.components || {}) }
 
   if (values) {
```
hehex9, line 33: noticed that `values` will always be truthy here

thekip: Nice catch, but looks like changes from this PR did not affect this.

andrii-bodnar:
@thekip main branch?

thekip:
Yes

andrii-bodnar (approved):

PR #1511, opened 2023-03-13
feat(format): initial support for external formatters

thekip:
# Description

Preparation for external formatters.

Currently, all extractors required in the CLI eagerly. It means no matter do you use CSV or po-gettext their dependency are loaded to the memory and affect cold-start time of CLI command. (and actually all bundler loaders plugins as well) 

So the first point was to make them loading lazily. 

This quickly revealed a problem with `mock-fs` in tests. 

If during execution of a command `require()` calls happened, they would fail because FS is mocked and no such file to require found. So it's breaking the code. 

In this PR i got rid of mock-fs in tests which were affected by this problem. 

And add initial support for external formatters. 

Notes: 
Before we can release "externalFormatters" for public it's important to make formatters work in the async context.  That will change their shape, obviously. I'm going to do this in next iteration to keep this PR small. 

Docs would be written when final shape of formatter would stabilized. 

## Types of changes

- [ ] Documentation update

Fixes # (issue)

## Checklist


andrii-bodnar:
@thekip thank you! How do you see the implementation of new formats?

It would be great if it allow the creation of separate shareable packages for different formats that can be installed via npm and then used by Lingui. In that way, every developer will be able to implement their own formatter and share it with other developers.

thekip:
> How do you see the implementation of new formats?

Exactly like you described. 

Even more, i'm going to extract all existing formats from `@lingui/cli` and make them available as separate packages. 

The only `@lingui/formatter-po` would be included in `@lingui/cli` dependencies, for the rest user would have to install them first. 

This is slightly reduce installation size, because if you don't use CSV formatter you don't have to install papaparse. 

It also will help to gather some kinda statistics which packages people uses and which not. 

For migration period there would be a warning

```js
// lingui.config.js
{
  format: 'minimal'
}
``` 
 > Error: Formatter `minimal` is no longer included by default. You have to install it with `npm -D @lingui/formatter-minimal`

Something like this. 

Also if you want to specify some formatter parameters you have to do it in the factory function instead of `formatterOptions`. This should reduce maintenance burden of current formatters. 

```js
// lingui.config.js
import formatter from "@lingui/formatter-minimal"
{
  locales: ['en', 'pl'],
  format: formatter({myAwesomeOption: true})
}
```

andrii-bodnar:
Sounds great! I think formatters should have access to as much contextual information about the string as possible. If needed, it will allow the generation of quality localization files with all the possible context for translators

thekip:
Made formatters work in the async context. (check last commit) 
To achieve async context in formatters, it was necessary to make the whole call tree up to the command async as well. 

Also reanimated and fixed tests for webpack loader, which was ignored.

thekip:
Looking into current formatters api i'm wondering about correctness and responsibilities. 

Currently, the api is `read(filename)` and `write(filename)` (simplified). 

It means that each formatter is responsible for two things: 

- Parsing/serializing of format, for example `js data` -> `.po` or `js data` -> `json`
- And working with FS. It means formatters go to FS by it's own, and reading/writing part is mostly duplicated between all formatters. 

I think it might be incorrect. Responsibility of formatter should be only first point from the list. 

So API should looks like as `parse(string | buffer): Catalog` and `serialize(catalog: Catalog): string | buffer`

This way all heavy lifting related to work with a FS would be out of the formatters reponsibilities. 

@andrii-bodnar what do you think?

thekip:
But there are some tricky logic in `write` functions in the formatters, which makes this ide a bit hard to implement: 

```ts
// po formatter
    async write(filename, catalog, ctx) {
      let po: PO

      const raw = await readFile(filename)
      if (raw) {
        po = PO.parse(raw)
      } else {
        po = new PO()
        po.headers = getCreateHeaders(ctx.locale)
        if (ctx.locale === undefined) {
          delete po.headers.Language
        }
        // accessing private property
        ;(po as any).headerOrder = Object.keys(po.headers)
      }
      po.items = serialize(catalog, options)
      await writeFileIfChanged(filename, po.toString())
    },
```

or 
```ts
      const messages = serialize(catalog)
      let file = await readFile(filename)

      const shouldUseTrailingNewline = file === null || file?.endsWith("\n")
      const trailingNewLine = shouldUseTrailingNewline ? "\n" : ""
      await writeFile(
        filename,
        `${JSON.stringify(messages, null, 2)}${trailingNewLine}`
      )
```


Anyway, i'm done with this PR. 
In follow up PRs i will extract formatters from @lingui/cli, add deprecation for `@lingui/conf` parameters and add docs.

andrii-bodnar:
@thekip I also think that working with FS definitely is not formatter's responsibility. Ideally, the formatter should be responsible only for parsing and serializing, and reading/writing should happen outside.

This will also reduce code duplication and the developer of the custom formatter will not care about the FS, only for the parsing and building of the file.

thekip:
@andrii-bodnar please, merge

thekip:
Finally, was able to make it (see last commit): 

> I also think that working with FS definitely is not formatter's responsibility. Ideally, the formatter should be responsible only for parsing and serializing, and reading/writing should happen outside.
>
> This will also reduce code duplication and the developer of the custom formatter will not care about the FS, only for the parsing and building of the file.

andrii-bodnar:
I'd like to review it in more detail, will merge it tomorrow

packages/cli/src/api/formats/index.ts
```diff
@@ -1,53 +1,37 @@
...
-  write(
-    filename: string,
-    catalog: CatalogType,
-    options?: CatalogFormatOptionsInternal
-  ): void
-  read(filename: string): CatalogType | null
-  parse(content: unknown): CatalogType | null
-}
+export function getFormat(
+  _format: CatalogFormat | CatalogFormatter,
```
andrii-bodnar, line 19: maybe it would be better to split this method? It's not the best approach to allow passing such different types and then check for this type inside the method

thekip: and what next? Do this check in all 5+ places where `getFormat()` is used?  

In an ideal world this check should not exist at all, but for backward compatibility it's here.

andrii-bodnar: yeah, it's also not a solution 😬

packages/cli/src/api/formats/po-gettext.ts
```diff
@@ -7,12 +7,18 @@ import gettextPlurals from "node-gettext/lib/plurals"
...
 import { CatalogType, MessageType } from "../types"
 import { readFile, writeFileIfChanged } from "../utils"
-import type { CatalogFormatOptionsInternal, CatalogFormatter } from "./"
+import type { CatalogFormatter } from "@lingui/conf"
 import { deserialize, serialize as serializePo } from "./po"
 
 // Workaround because pofile doesn't support es6 modules, see https://github.com/rubenv/pofile/pull/38#issuecomment-623119284
 type POItem = InstanceType<typeof PO.Item>
 
+export type PoGetTextFormatterOptions = {
```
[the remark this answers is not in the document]
thekip: fixed

thekip:
test suite should be rerun

thekip:
kinda flaky test, looks like depending on something  in host system glob may return results in one order or another. That will produce slightly different cli output which is not matched to snapshot

thekip:
fixed flaky test

PR #1515, opened 2023-03-14
feat: clarify the order by `messageId` in v4 and support order by message

taozhou-glean:
# Description

SInce the hash id support is in, the `messageId` now becomes the `linguiId` when custom id is absent. This is a change of behavior compared to v3, so its worth clarifying in the doc.

And to preserve the v3 behavior as much as we can, I am adding a new orderBy (`message`), so when no custom id is used, the order by `message` should match what `messageId` is expected in v3.

## Types of changes

- [x] Documentation update

Fixes # (issue) https://github.com/lingui/js-lingui/issues/1514

## Checklist

### DIFF
--- a/packages/cli/src/api/catalog.test.ts
+++ b/packages/cli/src/api/catalog.test.ts
@@ -553,6 +553,42 @@ describe("order", () => {
     // Jest snapshot order the keys automatically, so test that the key order explicitly
     expect(Object.keys(orderedCatalogs)).toMatchSnapshot()
   })
+
+  it("should order messages by message", () => {
+    const catalog = {
+      msg1: makeNextMessage({
+        message: "B",
+        translation: "B",
+        origin: [
+          ["file2.js", 2],
+          ["file1.js", 2],
+        ],
+      }),
+      msg2: makeNextMessage({
+        message: "A",
+        translation: "A",
+        origin: [["file2.js", 3]],
+      }),
+      msg3: makeNextMessage({
+        message: "D",
+        translation: "D",
+        origin: [["file2.js", 100]],
+      }),
+      msg4: makeNextMessage({
+        message: "C",
+        translation: "C",
+        origin: [["file1.js", 1]],
+      }),
+    }
+
+    const orderedCatalogs = order("message")(catalog)
+
+    // Test that the message content is the same as before
+    expect(orderedCatalogs).toMatchSnapshot()
+
+    // Jest snapshot order the keys automatically, so test that the key order explicitly
+    expect(Object.keys(orderedCatalogs)).toMatchSnapshot()
+  })
 })
 
 describe("writeCompiled", () => {
--- a/packages/cli/src/api/catalog.test.ts
+++ b/packages/cli/src/api/catalog.test.ts
@@ -565,7 +565,7 @@ describe("order", () => {
         ],
       }),
       msg2: makeNextMessage({
-        message: "A",
+        // message is optional.
         translation: "A",
         origin: [["file2.js", 3]],
       }),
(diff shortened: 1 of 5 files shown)


packages/conf/src/types.ts
```diff
@@ -37,7 +37,7 @@ export type CatalogFormatOptions = {
   disableSelectWarning?: boolean
 }
 
-export type OrderBy = "messageId" | "origin"
+export type OrderBy = "messageId" | "message" | "origin"
```
taozhou-glean, line 40: I thought about renaming or adding `linguiId` as an option, but feel like it may be misleading itself, as we do use custom message id as messageId if exists. And support purely order by `linguiId` seems not that helpful at the moment. So only adding the `message` here.

andrii-bodnar (approved):
@taozhou-glean thank you!

website/docs/ref/conf.md
```diff
@@ -369,7 +369,11 @@ Order of messages in catalog:
 
 #### messageId
 
-Sort by the message ID.
+Sort by the message ID, `js-lingui-id` will be used if no custom id provided.
+
+#### message
+
+Sort by message.
```
thekip, line 376:
```suggestion
Sort by source message.
```

packages/cli/src/api/catalog.ts
```diff
@@ -337,3 +338,20 @@ export function orderByOrigin<T extends ExtractedCatalogType>(messages: T): T {
...
+export function orderByMessage<T extends ExtractedCatalogType>(messages: T): T {
+  return Object.keys(messages)
+    .sort((a, b) => {
+      const aMsg = messages[a].message
+      const bMsg = messages[b].message
+
+      if (aMsg < bMsg) return -1
+      if (aMsg > bMsg) return 1
+
+      return 0
```
thekip, line 351:
```suggestion
      const aMsg = messages[a].message || ""
      const bMsg = messages[b].message || ""

      return aMsg.localeCompare(bMsg)
```

packages/cli/src/api/catalog.test.ts
```diff
@@ -553,6 +553,42 @@ describe("order", () => {
...
+        message: "D",
+        translation: "D",
+        origin: [["file2.js", 100]],
+      }),
+      msg4: makeNextMessage({
+        message: "C",
+        translation: "C",
+        origin: [["file1.js", 1]],
+      }),
+    }
```
thekip, line 582: There might be the case when source message is not present at all. Add it to the test-case please.

packages/cli/src/api/catalog.test.ts
```diff
@@ -553,6 +553,42 @@ describe("order", () => {
...
+
+    const orderedCatalogs = order("message")(catalog)
+
+    // Test that the message content is the same as before
+    expect(orderedCatalogs).toMatchSnapshot()
```
thekip, line 587: drop this

packages/cli/src/api/catalog.test.ts
```diff
@@ -553,6 +553,42 @@ describe("order", () => {
...
+
+    // Jest snapshot order the keys automatically, so test that the key order explicitly
+    expect(Object.keys(orderedCatalogs)).toMatchSnapshot()
```
thekip, line 590: Replace to `toMatchInlineSnapshot()` for better test readability

thekip (changes requested):
Thanks for contribution. I also thinks that sort by message should be a default one. Because this would be less breaking for most supported usecase. 

For those who uses explicit ids there always an option to enable "messageId" sort manually

andrii-bodnar:
@taozhou-glean sounds good to me

thekip (approved):

taozhou-glean:
> @taozhou-glean sounds good to me

cool, will change the default `orderBy` in https://github.com/lingui/js-lingui/pull/1519, less change than expected, gonna rebase once this is merged, not including it in this PR just to make change more readable from commit perspective.

Issue #1512, opened 2023-03-13
FEATURE REQUEST: V4 does not support extracting from code using legacy decorators


capaj:
**Describe the bug**
When I run this: lingui extract I am getting `SyntaxError: /home/capaj/work-repos/official/official-graphql2/graphql/src/models/love-buttons/LoveButtons.ts: Decorators cannot be used to decorate parameters. (63:4)`

**To Reproduce**
have a file like this:

```ts
import { Authorized, Ctx, Query, Resolver } from 'type-graphql'
import { IContext, IContextAuth } from '../makeAuthContext'
import {
  getActiveMembershipForUser,
  MembershipModel
} from '../models/MembershipModel'
import { i18n } from '@lingui/core'

@Resolver(MembershipModel)
export class MembershipResolver {
  @Authorized()
  @Query(() => MembershipModel, {
    nullable: true
  })
  activeMembership(@Ctx() ctx: IContextAuth) {
    const { user } = ctx
    return getActiveMembershipForUser(user?.aws_user_id)
  }
}

const message = i18n._('Hello worlkd')
```

**Expected behavior**
a message gets extracted

**Additional context**
When I remove all decorators from my files it works ok and extract my message as expected.

- jsLingui version ``lingui --version``: 4.0.0.next-1
- Babel version ``npm list @babel/core`` 7.21.0
- Your Babel config (e.g. `.babelrc`) or framework you use (Create React App, Meteor, etc.)

```
 {
    plugins: [
      '@babel/plugin-transform-typescript',
      ['@babel/plugin-proposal-decorators', { legacy: true }],
      ['@babel/plugin-proposal-class-properties', { loose: true }],
      'macros'
    ]
  }
```

thekip:
try this: 

```ts
// lingui.config.js
extractorParserOptions: {
  decoratorsBeforeExport: true
}
```

thekip:
Not very popular case for React projects. I will add a fix.

capaj:
oh we cannot use linguijs with legacy decorators ATM, certainly not 4.0.0.next-1

because this array
https://github.com/lingui/js-lingui/blob/580e7027032044c927f3b77b0c414e3f7950670e/packages/cli/src/api/extractors/babel.ts#L81-L94

cannot be extended from `lingui.config.js`

and this https://github.com/lingui/js-lingui/blob/580e7027032044c927f3b77b0c414e3f7950670e/packages/cli/src/api/extractors/babel.ts#L33

[...]

PR #1517, opened 2023-03-14
feat(extractor): support TS experimental decorators

thekip:
# Description

Babel parser since 7.21.0 understand both before and after export decorators, so `decoratorsBeforeExport` parser option is not needed. 

But for typescript's legacy decorators which could be applied for class method parameters we need to use different parser plugin. 

The problem is that these two plugins are not compatible with each other, so user should choose one. Use standardized decorators in theirs typescript code (default behaviour) or experimental (legacy). For JS code standardized decorators are always applied. 

This is more or less aligned with upcoming typescript 5 release https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#decorators

## Types of changes

- [ ] Documentation update

Fixes https://github.com/lingui/js-lingui/issues/1512

## Checklist


packages/cli/src/api/fixtures/collect-typescript-jsx/jsx-syntax.jsx
```diff
@@ -5,34 +5,38 @@ const jsx = <div>Hello!</div>
...
+  b = 1
 
   // classPrivateProperties
-  #b = 1;
+  #b = 1
 }
 
 // dynamicImport
-import('./guy').then(a)
+import("./guy").then(a)
```
thekip, line 27: Prettier change the formatting on save / before commit actions even if this files added as ignored 😬

andrii-bodnar (approved):

Issue #1522, opened 2023-03-16
variable only cases in select is omitted


taozhou-glean:
**Describe the bug**

pure variable variant is omitted from the extracted message, resulting in invalid ICU format

```
select(variant, {
    other: localeDateAndTime,
    today: `Today at ${localeTime}`,
    tomorrow: `Tomorrow at ${localeTime}`,
    yesterday: `Yesterday at ${localeTime}`,
  })
```

right now, `other` is omitted from the extracted message 

```
 {variant, select, today {Today at {localeTime}} tomorrow {Tomorrow at {localeTime}} yesterday {Yesterday at {localeTime}}
```

and according to https://format-message.github.io/icu-message-format-for-translators/editor.html, this format is invalid as missing `other` there.

**Expected behavior**

expect `other` is included

`{variant, select, other {{localeDateAndTime}} today {Today at {localeTime}} tomorrow {Tomorrow at {localeTime}} yesterday {Yesterday at {localeTime}}`

**Additional context**

not sure if this is v4 only, only tested on v4 tho

PR #1523, opened 2023-03-16
fix: expression only choices / select options

taozhou-glean:
# Description

to fix https://github.com/lingui/js-lingui/issues/1522

## Types of changes

- [ ] Documentation update

Fixes # (issue) https://github.com/lingui/js-lingui/issues/1522

## Checklist

### DIFF
--- a/packages/macro/src/macroJs.test.ts
+++ b/packages/macro/src/macroJs.test.ts
@@ -262,56 +262,16 @@ describe("js macro", () => {
       })
     })
 
-    it("plural with variable only cases", () => {
-      const macro = createMacro()
-      const exp = parseExpression(
-        "plural(count, { one: `# glass of ${drink}`, other: drink})"
-      )
-      const tokens = macro.tokenizeChoiceComponent(exp as CallExpression)
-      expect(tokens).toEqual({
-        type: "arg",
-        name: "count",
-        value: expect.objectContaining({
-          name: "count",
-          type: "Identifier",
-        }),
-        format: "plural",
-        options: {
-          one: [
-            {
-              type: "text",
-              value: "# glass of ",
-            },
-            {
-              type: "arg",
-              name: "drink",
-              value: expect.objectContaining({
-                name: "drink",
-                type: "Identifier",
-              }),
-            },
-          ],
-          other: {
-            type: "arg",
-            name: "drink",
-            value: expect.objectContaining({
-              name: "drink",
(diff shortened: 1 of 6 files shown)


packages/macro/src/macroJs.ts
```diff
@@ -352,8 +352,12 @@ export default class MacroJs {
           value = this.tokenizeTemplateLiteral(attrValue)
         } else if (this.types.isCallExpression(attrValue)) {
           value = this.tokenizeNode(attrValue)
+        } else if (this.types.isStringLiteral(attrValue)) {
+          value = attrValue.value
+        } else if (this.types.isExpression(attrValue)) {
```
taozhou-glean, line 357: StringLiteral is also considered as Expression, so I moved StringLiteral check above. From ts,  after all the checks, the types fall through are limited to `babelTypes.RestElement | babelTypes.AssignmentPattern | babelTypes.ArrayPattern | babelTypes.ObjectPattern`,  not sure exactly how we want to handle them, so I kept the previous fallback case untouched.

taozhou-glean: the `unknown` conversion is just to make ts happy, please let me know if you have other suggestions to handle the remaining cases.

packages/macro/src/macroJs.test.ts
```diff
@@ -262,6 +262,47 @@ describe("js macro", () => {
...
+            name: "drink",
+            value: expect.objectContaining({
+              name: "drink",
+              type: "Identifier",
+            }),
+          },
+        },
+      })
+    })
+
```
thekip, line 305: You can safely drop unit tests from macroJs.test.ts file. They don't have a lot of sense, it's perfectly covered from fixtures with js code and this just create more work for maintenance.

thekip (changes requested):
I think the same is true for JSX <Plural> and <Select> components. Could you also check add that? 

And it also should be checked for SWC Plugin. I think it should work already but at least the same test cases should be added to have test suite aligned.

taozhou-glean:
swc-plugin change in https://github.com/lingui/swc-plugin/pull/40 

is there a guide how to run tests etc ? I have zero experience with rust before... I can read more about it but would be great if there is a command that I can just copy and paste to run the tests ;)

thekip (approved):

andrii-bodnar (approved):

PR #1546, opened 2023-03-22
fix: fix type mismatch between formatter and conf/CatalogFormatter

taozhou-glean:
# Description

fix type mismatch: 

image

## Types of changes

- [ ] Documentation update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/conf/__typetests__/index.test-d.tsx
+++ b/packages/conf/__typetests__/index.test-d.tsx
@@ -5,6 +5,7 @@ import {
   LinguiConfig,
 } from "@lingui/conf"
 import { expectAssignable } from "tsd"
+import { formatter } from "@lingui/format-po"
 
 // only required props
 expectAssignable<LinguiConfig>({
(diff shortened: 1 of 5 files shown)


packages/conf/src/types.ts
```diff
@@ -62,12 +62,12 @@ export type CatalogFormatter = {
   templateExtension?: string
   parse(
     content: string,
-    ctx: { locale: string | null; sourceLocale: string; filename: string }
+    ctx: { locale: string; sourceLocale: string; filename: string }
```
taozhou-glean, line 65: from the call site, it looks like the `locale` is always present, while `existing` may not.

thekip: locale may not be present in case if you generating a template.

packages/conf/__typetests__/index.test-d.tsx
```diff
@@ -61,6 +62,12 @@ expectAssignable<LinguiConfig>({
   ],
 })
 
+// formatter
+expectAssignable<LinguiConfig>({
```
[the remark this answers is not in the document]
thekip: this test should be in `format-po` package.

packages/conf/src/types.ts
thekip: please return `| null` back

thekip (commented):
this is incorrect fix, let me do this properly in another PR.

thekip:
supressed by https://github.com/lingui/js-lingui/pull/1547

PR #1549, opened 2023-03-22
refactor(cli/catalog): decouple catalog type from Po message shape

thekip:
# Description

While working on a new xliff formatter realized that current shape of CatalogType and ExtractedMessage are tightly coupled to shape of the message of Pofile. There some properties which is related to po formatter only and should not be in the core type. 

The same with other formatters. For example xliff may have additional attributes such as `state=final` which we also should store somewhere to not lost when extracting.  So there should be a way to add some `extra` fields which would be preserved by cli and returned back to formatter. 

In scheme it something like

formatter.parse -> extra: {special field} -> cli catalog processing -> formatter.serialize (get the same extra: {special field}) 

Notes: 

i renamed extractedComments -> comments.  this naming was because Pofile has it. When it's uncoupled, we don't 
 need it anymore and just `comments` looks much cleaner in the api.

## Types of changes

- [ ] Documentation update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/cli/src/api/__snapshots__/catalog.test.ts.snap
+++ b/packages/cli/src/api/__snapshots__/catalog.test.ts.snap
@@ -10,8 +10,8 @@ exports[`Catalog POT Flow Should get translations from template if locale file n
 exports[`Catalog collect should extract messages from source files 1`] = `
 {
   Component A: {
+    comments: [],
     context: undefined,
-    extractedComments: [],
     message: undefined,
     origin: [
       [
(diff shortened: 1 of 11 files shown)


andrii-bodnar (approved):

andrii-bodnar:
@thekip, just noticed - https://github.com/lingui/js-lingui/blob/next/website/docs/guides/custom-formatter.md?plain=1#L53 😬

Issue #1573, opened 2023-03-30
v4 .po files cannot be saved with POEdit


ptmkenny:
**Describe the bug**

I've been using POEdit as a GUI to translate my app.

My app uses explicit ids for all strings.

However, after updating to v4, after opening a file output by `lingui extract` in POEdit, editing it, and saving it, when I re-run `lingui extract --clean --overwrite` to update the strings, **all the changes in POEdit are removed**.

**To Reproduce**

1. Generate a lingui PO file that contains strings with explicit IDs.
2. Edit the file in POEdit. Save.
3. Run `lingui extract`.  The edits are deleted (if using `--clean --overwrite`) or new strings are generated (if just running `extract` with no options).

**Expected behavior**

Edits should be kept.

**Additional context**

I think what is going on is that since v4, lingui now adds a flag to PO files: `#, explicit-id`.  However, POEdit doesn't know how to handle this, and it removes the flag on save.  So this causes Lingui to recreate the strings fresh and remove the strings saved by POEdit (that lack a `#, explicit-id`).

- jsLingui version ``lingui --version``
4.0-next4

thekip:
Confirmed, POEditor removes flags which could not handle. I see only one way to fix that. Move this flags to the comments, as it's done in `po-gettext` formatter. It seems POEdit preserve comments in po files.

PR #1574, opened 2023-03-30
fix(po-formatter): save explicit id flag in comments instead of flags

thekip:
# Description

fixes https://github.com/lingui/js-lingui/issues/1573

### DIFF
--- a/packages/format-po/src/po.ts
+++ b/packages/format-po/src/po.ts
@@ -60,7 +60,7 @@ function getCreateHeaders(language: string): PO["headers"] {
   }
 }
 
-const EXPLICIT_ID_FLAG = "explicit-id"
+const EXPLICIT_ID_FLAG = "js-lingui-explicit-id"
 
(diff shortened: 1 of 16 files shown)


packages/format-po/src/po.ts
```diff
@@ -60,7 +60,7 @@ function getCreateHeaders(language: string): PO["headers"] {
   }
 }
 
-const EXPLICIT_ID_FLAG = "explicit-id"
+const EXPLICIT_ID_FLAG = "js-lingui-explicit-id"
```
andrii-bodnar, line 63: Maybe the constant name should be changed as well?

```suggestion
const EXPLICIT_ID_COMMENT = "js-lingui-explicit-id"
```

thekip: i suppose it still a flag but saved in comment?  for me naming is ok

andrii-bodnar (approved):

Issue #1575, opened 2023-03-31
`mjs` is used instead of cjs in jest


taozhou-glean:
```
/node_modules/@lingui/detect-locale/dist/index.mjs:112
    export { detect, detectFromCookie as fromCookie, detectHtmlTag as fromHtmlTag, detectFromNavigator as fromNavigator, detectFromPath as fromPath, detectFromStorage as fromStorage, detectFromSubdomain as fromSubdomain, detectFromUrl as fromUrl, multipleDetect };
    ^^^^^^

    SyntaxError: Unexpected token 'export'
```
saw this when trying to upgrade to next-4
looks like somehow it switched to import mjs instead of cjs after 3->4, and only happens in jest tests

PR #1576, opened 2023-03-31
fix: require should use cjs instead of mjs

taozhou-glean:
# Description

this is breaking our jest as the default jest resolver ([`resolve.exports`](https://github.com/lukeed/resolve.exports/) with default conditions (`['require', 'default' ...]`)) will try to pull in asset listed in the package, and the mapping on `require` for detect-locale was pointing to `mjs` instead of `cjs`.

I have double checked and this seems to be the only one across all lingui packages ;)

## Types of changes

- [ ] Documentation update

Fixes https://github.com/lingui/js-lingui/issues/1575

## Checklist

### DIFF
--- a/packages/detect-locale/package.json
+++ b/packages/detect-locale/package.json
@@ -35,7 +35,7 @@
   },
   "exports": {
     ".": {
-      "require": "./dist/index.mjs",
+      "require": "./dist/index.cjs",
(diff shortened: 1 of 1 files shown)


andrii-bodnar (approved):
@taozhou-glean thanks!

PR #1583, opened 2023-04-05
refactor(cli/api): update pseudolocale package

thekip:
HOTFIX: as the author of `pseudolocale` pushed breaking changes without a major release bump, now it breaks lingui if user will install from "scratch" (or delete lock file)


thekip:
i think i will just fix the version of the package for `main` branch.

j4hr3n (approved):
Nice. Ran into this issue myself recently as well

andrii-bodnar (approved):

PR #1622, opened 2023-04-26
chore: import type testing APIs from `tsd-lite`

mrazauskas:
# Description

Glad to see you found it helpful to use `jest-runner-tsd`. As you may have noticed, instead of `tsd` this runner is using `tsd-lite` (I am maintaining this package). Just wanted to draw your attention, that it would be better to import type testing APIs from `tsd-lite`, because both implementations differ. For example, `tsd-lite` does not include few API which are not related with type testing.

Also the setup becomes somewhat more optimal. Take a look at changes in `yarn.lock`. `tsd-lite` makes your repo lighter (;

## Types of changes

- [ ] Documentation update
- [x] Dependency update

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "jest": "^29.4.3",
     "jest-environment-jsdom": "^29.4.3",
     "jest-environment-node-single-context": "^29.0.0",
-    "jest-runner-tsd": "^4.0.0",
+    "jest-runner-tsd": "^5.0.0",
     "lerna": "^6.5.1",
     "lint-staged": "^13.1.0",
     "memory-fs": "^0.5.0",
--- a/packages/conf/__typetests__/index.test-d.tsx
+++ b/packages/conf/__typetests__/index.test-d.tsx
@@ -4,7 +4,7 @@ import {
   FallbackLocales,
   LinguiConfig,
 } from "@lingui/conf"
-import { expectAssignable } from "tsd"
+import { expectAssignable } from "tsd-lite"
 
 // only required props
 expectAssignable<LinguiConfig>({
@@ -33,13 +33,13 @@ expectAssignable<LinguiConfig>({
     tsExperimentalDecorators: false,
   },
   fallbackLocales: {} as FallbackLocales,
-  format: "po",
+  format: "po" as const,
   formatOptions: { origins: true, lineNumbers: true },
   locales: [],
-  orderBy: "messageId",
+  orderBy: "messageId" as const,
   pseudoLocale: "",
   rootDir: ".",
-  runtimeConfigModule: ["@lingui/core", "i18n"],
+  runtimeConfigModule: ["@lingui/core", "i18n"] as [string, string],
   sourceLocale: "",
   service: { name: "", apiKey: "" },
 })
@@ -64,13 +64,13 @@ expectAssignable<LinguiConfig>({
 // runtimeConfigModule
 expectAssignable<LinguiConfig>({
   locales: ["en", "pl"],
-  runtimeConfigModule: ["./custom-i18n-config", "i18n"],
+  runtimeConfigModule: ["./custom-i18n-config", "i18n"] as [string, string],
 })
 
 expectAssignable<LinguiConfig>({
   locales: ["en", "pl"],
   runtimeConfigModule: {
-    i18n: ["./custom-config", "i18n"],
-    Trans: ["./custom-config", "Trans"],
+    i18n: ["./custom-config", "i18n"] as [string, string],
+    Trans: ["./custom-config", "Trans"] as [string, string],
   },
 })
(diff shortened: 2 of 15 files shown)


package.json
```diff
@@ -64,7 +64,7 @@
     "jest": "^29.4.3",
     "jest-environment-jsdom": "^29.4.3",
     "jest-environment-node-single-context": "^29.0.0",
-    "jest-runner-tsd": "^4.0.0",
+    "jest-runner-tsd": "^5.0.0",
```
mrazauskas, line 67: You can always use latest version, because `@tsd/typescript` is peer dependency. Makes it easy to test on any TS version and have latest bugs fixed.

packages/conf/__typetests__/index.test-d.tsx
```diff
@@ -33,13 +33,13 @@ expectAssignable<LinguiConfig>({
     tsExperimentalDecorators: false,
   },
   fallbackLocales: {} as FallbackLocales,
-  format: "po",
+  format: "po" as const,
```
mrazauskas, line 36: The inferred type of `"po"` is `string` and the inferred type of `"po" as const` is `"po"`. APIs of `tsd-lite` distinguishes between those two. For my understanding, this is more precise approach.

packages/conf/__typetests__/index.test-d.tsx
```diff
@@ -33,13 +33,13 @@ expectAssignable<LinguiConfig>({
...
   formatOptions: { origins: true, lineNumbers: true },
   locales: [],
-  orderBy: "messageId",
+  orderBy: "messageId" as const,
   pseudoLocale: "",
   rootDir: ".",
-  runtimeConfigModule: ["@lingui/core", "i18n"],
+  runtimeConfigModule: ["@lingui/core", "i18n"] as [string, string],
```
mrazauskas, line 42: Same here. The inferred type of `["@lingui/core", "i18n"]` is `Array<string>`.

yarn.lock
```diff
@@ -3765,20 +3765,13 @@ __metadata:
...
 
-"@tsd/typescript@npm:^4.9.5, @tsd/typescript@npm:~4.9.5":
+"@tsd/typescript@npm:^4.9.5":
   version: 4.9.5
   resolution: "@tsd/typescript@npm:4.9.5"
   checksum: f9f01ecf2094e0ac83e56cc57430bdf81da82aa20424280002b44ea29cdc171388a9777262257f550800bf76567774521728e1ef6e191633ae6f723ee64d396a
   languageName: node
   linkType: hard
 
-"@tsd/typescript@npm:~5.0.2":
```
mrazauskas, line 3775: For example, this is a copy of the whole `typescript` package, which you don’t use. It was here just because `tsd` ships it.

packages/conf/__typetests__/index.test-d.tsx
thekip: Hi, thanks for help. In this case what we have in test is different from what users will write in theirs code. Isn't the reason of these tests is to check that all user-scenarios would work as much as closer to real life? 

I mean, what if it will pass type tests only because of explicit casting to `const` but will not work in the user's codebase without it?

mrazauskas: Good point. I see at least two scenarios on user side:

```ts
export const config: LinguiConfig = {
  locales: ["en", "pl"],
  orderBy: "messageId", // no need for `as const`, that’s how `tsd` interprets it
}
```

```ts
let config: LinguiConfig = {
  locales: ["en", "pl"],
}

// simplified for demonstration, this could be returned by some function or so
const extendedConfig = {
  locales: ["en", "pl"],
  orderBy: "origin" as const, // requires `as const` to pass assignability check later
}

config = extendedConfig
```

Through my eyes, first case is declaration of a variable. Assignability happens in the second example. That’s why requiring `as const` makes more sense for me.

Would be good to cover both, or? I can add the declaration case to the type test file. If that is acceptable.

mrazauskas: Also try this with `tsd`:

```ts
import { LinguiConfig } from "@lingui/conf"
import { expectAssignable, expectNotAssignable } from "tsd"

// pass
expectAssignable<LinguiConfig>({
  locales: ["en", "pl"],
  format: "po",
})

// also pass! wait? how?!
expectNotAssignable<LinguiConfig>({
  locales: ["en", "pl"],
  format: "po",
})
```

mrazauskas: Ah.. Actually there is no need for extra test case, because the two below are equivalent:

```ts
const format: "po" = "po"

const format = "po" as const
```

Having `"po" as const` in an assignability test means that user will be allowed passing bare `"po"` while declaring a variable, but `"po" as const` will be required in case of assignment.

---

Sorry about this stream of thoughts. Type testing turns me on ;D

mrazauskas:
Never mind.

PR #1623, opened 2023-04-27
fix(vue-extractor): support components with no <template>

Touffy:
# Description

An improvement to the recently merged Vue single-file component extractor : A SFC may have no `<template>` (functional components), in which case the extract should just move on with the script.

## Types of changes

## Checklist

- [x] I have added necessary documentation (if appropriate) (N/A)

### DIFF
--- a/packages/extractor-vue/src/vue-extractor.ts
+++ b/packages/extractor-vue/src/vue-extractor.ts
@@ -18,12 +18,14 @@ export const vueExtractor: ExtractorType = {
       ignoreEmpty: true,
     })
 
-    const compiledTemplate = compileTemplate({
-      source: descriptor.template.content,
-      filename,
-      inMap: descriptor.template.map,
-      id: filename,
-    })
+    const compiledTemplate =
+      descriptor.template &&
+      compileTemplate({
+        source: descriptor.template.content,
+        filename,
+        inMap: descriptor.template.map,
+        id: filename,
+      })
 
     const isTsBlock = (block: SFCBlock) => block?.lang === "ts"
 
--- a/packages/extractor-vue/src/__snapshots__/extractor.test.ts.snap
+++ b/packages/extractor-vue/src/__snapshots__/extractor.test.ts.snap
@@ -1,5 +1,21 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`vue extractor should extract message from functional component 1`] = `
+[
+  {
+    comment: undefined,
+    context: undefined,
+    id: Render function message,
+    message: undefined,
+    origin: [
+      functional.vue,
+      10,
+      33,
+    ],
+  },
+]
+`;
+
 exports[`vue extractor should extract message from vue file 1`] = `
 [
   {
--- a/packages/extractor-vue/src/extractor.test.ts
+++ b/packages/extractor-vue/src/extractor.test.ts
@@ -61,4 +61,26 @@ describe("vue extractor", () => {
 
     expect(messages).toMatchSnapshot()
   })
+
+  it("should extract message from functional component", async () => {
+    const filePath = path.resolve(__dirname, "fixtures/functional.vue")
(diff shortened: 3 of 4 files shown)


packages/extractor-vue/src/vue-extractor.ts
```diff
@@ -12,13 +12,16 @@ export const vueExtractor: ExtractorType = {
...
       sourceMap: true,
       filename,
       ignoreEmpty: true,
     })
+    // this handles older SFC compilers (2.6) which return the descriptor directly
+    // @ts-ignore of course we're using typings for more recent SFC compiler so TS disagrees
+    const descriptor = parsed.descriptor || parsed as SFCDescriptor
 
-    const compiledTemplate = compileTemplate({
+    const compiledTemplate = descriptor.template && compileTemplate({
```
thekip, line 24: `"@vue/compiler-sfc": "^3.2.47"` written as dependency of `@lingui/extractor-vue` don't understand how this extractor would use older version and why.

Touffy: Yeah, me neither. My client uses a monorepo for most of the front-end (which is a couple of Vue 2.6 apps and some libraries), and I don't really understand how lerna and yarn decide what to install (we don't have a forced resolution for those packages) but the result is that when I run `lingui extract`, it ends up using compiler-sfc@2.6 and it crashes without this code.

Touffy: Actually the culprit may be the babel config (which lingui uses implicitly) rather than yarn/lerna.

thekip: I think it shouldn't be in the library code, because it's rather monorepo configuration issue. You need to check that all packages explicitly states all dependencies and there no "phantom dependecies" - for example `lingui extract` is used in some packages but not directly installed there.

thekip: > Actually the culprit may be the babel config (which lingui uses implicitly) rather than yarn/lerna. 

that is not the case, babel config in extractor is sealed and not affected by environment configuration.

Touffy: lingui and related packages are all a direct, explicit dependency of the app I'm using it for.

I agree that this shouldn't happen, but I didn't hallucinate the stack trace.

I don't mind removing that change. I can't use this package as-is anyway because of how the base extractor is imported (another issue with babel, maybe). But please consider that it's just a few more bytes and it may help others with messed-up configs.

thekip: >  because of how the base extractor is imported (another issue with babel, maybe).

Could you elaborate a bit on that?

Touffy: Like I said in the other discussion, when this is transpiled to commonjs, the result is `const babel = require('@lingui/cli/api/extractors/babel');` but what actually works (for me) is `const babel = require('@lingui/cli/api/extractors/babel').default;`.

thekip: well, i know why this happend. I will try to fix that. 

Regarding `const descriptor = parsed.descriptor || parsed as SFCDescriptor` i would suggest removing it because it brings maintenance burden (we are not responsible for misconfigured monorepos whatsoever) and this would be good to be merged

Touffy: OK. I removed that commit.

thekip (approved):

andrii-bodnar (approved):

thekip:
Fixed issue with `.default` https://github.com/lingui/js-lingui/pull/1626#issuecomment-1525656124

andrii-bodnar:
@Touffy thanks for the contribution! It seems like there are some Prettier issues in the code. Could you please take a look?

Touffy:
Sorry. Fixed.

Issue #1631, opened 2023-05-01
Translation.io service problem after upgrade `@lingui/cli` from `v4.0.0-next.5` to `v4.0.0`


K1DV5:
**Describe the bug**
I have the `TranslationIO` service enabled and works fine when `@lingui/cli` is at `v4.0.0-next.5`. But when `v4.0.0` comes along it says:
```
Can't load service module translationIO Error: Cannot find module '...node_modules/@lingui/cli/dist/services/translationIO' imp
orted from node_modules/@lingui/cli/dist/lingui-extract.js
```
And when I checked, the file `node_modules/@lingui/cli/dist/services/translationIO.js` does exist.

**To Reproduce**
1. Have any working project that has the `TranslationIO` service enabled, with `@lingui/cli v4.0.0-next.5` 
2. Update `@lingui/cli` to `v4.0.0` 
3. Run `npm run sync`

**Expected behavior**
`npm run sync` runs without a problem.

**Additional context**
My project is a private one with `"type": "module"`, if that's relevant.

- jsLingui version ``lingui --version``: 4.0.0
- Babel version ``npm list @babel/core``: 7.21.4
- Macro support: Enabled
 - [ ] I'm using SWC with `@lingui/swc-plugin`
 - [ ] I'm using Babel with `babel-macro-plugin`
 - [ ] I'm not using macro
- Your Babel config (e.g. `.babelrc`) or framework you use (Create React App, NextJs, Vite): Preact with Vite.

ZhengYuTay:
Yea, was able to reproduced with `TranslationIO` in lingui config file as well

thekip:
try to pass name of service with extension 

```js
 service: { name: "translationIO.js", apiKey: "" },
```

ZhengYuTay:
I tried but it still doesn't work and would result in

```
Can't load service module translationIO TypeError: module.default is not a function
```

to make it work would need to be like `module.default.default(xxx)`

ZhengYuTay:
I opened a PR with `require` solution, but feel free to fix it with any other solutions

PR #1647, opened 2023-05-10
fix: just use require to import services

ZhengYuTay:
- update services to be async and can be awaited

# Description

from #1631, there is error importing TranslationIO service with `import`. It result in `Cannot find module`. The problem is that it need the js extension.

One solution is to add js extension `import('xxx' + .js)` , but the result would still need to be
```
import('xxx'.js).then(module => module.default.default(xxx))
```
Not sure why the double default there, but using require would solve at least the double default.

## Types of changes

- [ ] Documentation update

Fixes # (issue)
#1631 

## Checklist


thekip (approved):
Thanks, that should help.

andrii-bodnar (approved):

PR #1671, opened 2023-05-24
fix(webpack + vite): fix dependency watching in loader

thekip:
# Description

Here we go again. The deps watching still doesn't work and cause problems with cache bursting on my project. 

Found that i erroneously assumed that path in the Catalog is a relative, where it actually is absolute. Based on this wrong assumption, unnecessary path concatenation was made, resulting in incorrect paths for files. 

I also write a test for webpack loader checking that webpack actually watching all files and react on the changes in them.

### DIFF
--- a/packages/cli/src/api/catalog/getCatalogDependentFiles.test.ts
+++ b/packages/cli/src/api/catalog/getCatalogDependentFiles.test.ts
@@ -49,7 +49,6 @@ describe("getCatalogDependentFiles", () => {
       config
     )
 
-    console.log(catalog.path)
     const actual = await getCatalogDependentFiles(catalog, "pt-PT")
     mockFs.restore()
 
--- a/packages/cli/src/api/catalog/getCatalogDependentFiles.test.ts
+++ b/packages/cli/src/api/catalog/getCatalogDependentFiles.test.ts
@@ -16,7 +16,7 @@ describe("getCatalogDependentFiles", () => {
 
   it("Should return list template + fallbacks + sourceLocale", async () => {
     mockFs({
-      "src/locales": {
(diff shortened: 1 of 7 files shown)


andrii-bodnar (approved):

taozhou-glean:
maybe related, we start seeing following errors after the latest release:

```
Invalid dependencies have been reported by plugins or loaders for this module. All reported dependencies need to be absolute paths.
Invalid dependencies may lead to broken watching and caching.
As best effort we try to convert all invalid values to absolute paths and converting globs into context dependencies, but this is deprecated behavior.
Loaders: Pass absolute paths to this.addDependency (existing files), this.addMissingDependency (not existing files), and this.addContextDependency (directories).
Plugins: Pass absolute paths to fileDependencies (existing files), missingDependencies (not existing files), and contextDependencies (directories).
Globs: They are not supported. Pass absolute path to the directory as context dependencies.
The following invalid values have been reported:
 * "data/locales/web/en/messages.po"
```

thekip:
@taozhou-glean yes, this is related, but could you create a miminal repro?  Don't understand the case where relative paths happened

taozhou-glean:
@thekip sorry for the late reply, was busy with something else. I just checked, i think the reason is that the `catalogs.path` is relative in our lingui.config.ts, so it has been relative path to loader all the time. Let me see if I can fix it on our side, otherwise I think maybe we can change this behavior to check `isAbsolute` and continue the resolving if it's not. (we use `bazel` for our builds and some symlinks is involved here, so it might not be that easy to change it to absolute.

gpessa:
I still have the same issue despite the fix

Issue #1673, opened 2023-05-26
nextjs-swc is unusable


yunsii:
**Describe the bug**
First render by SSR:

`/es`
![image](https://github.com/lingui/swc-plugin/assets/18096089/c723635d-597e-4afb-8098-bfe997011862)

`/`
![image](https://github.com/lingui/swc-plugin/assets/18096089/ec12c52b-dc30-4fdb-8796-447da60bc305)

AboutText componnent do not translate completely. 

---

After change path by CSR: `/` => `/es`

![image](https://github.com/lingui/swc-plugin/assets/18096089/1305c3e7-c8d4-491c-aab9-a5c9b6465c22)

Developers component translation is still English translation. after CSR `/es` => `/` path it use Spanish translation. It looks like CSR always use previous translation. It is so ridiculous...

**To Reproduce**

https://github.com/lingui/js-lingui/tree/main/examples/nextjs-swc

**Expected behavior**

Use correct tranlation.

**Additional context**

- jsLingui version 4.1.2
- Macro support:
 - [x] I'm using SWC with `@lingui/swc-plugin`
 - [ ] I'm using Babel with `babel-macro-plugin`
 - [ ] I'm not using macro
- Your framework you use NextJs 13.3.3

thekip:
https://github.com/lingui/js-lingui/pull/1674

yunsii:
Thanks for your work, the example works like a charm. By the way, when i update these deps
![image](https://github.com/lingui/js-lingui/assets/18096089/1fa3b3d0-36ed-4a0e-9007-6fddfe271f71)

i should update the usage too?

yunsii:
https://github.com/lingui/js-lingui/blob/v4.0.0-next.7/website/docs/releases/migration-4.md
https://github.com/lingui/js-lingui/blob/main/website/docs/releases/migration-4.md
The document update too 😂

thekip:
i don't think i've got the question. 
There are difference between v3 and v4. But i think there were no major changes between `-next.7` and final release.

PR #1674, opened 2023-05-26
docs: migrate examples to v4

thekip:
# Description

Some of examples get outdated after updating to v4. This PR migrates them to latest version. 

https://github.com/lingui/js-lingui/issues/1673

## Types of changes

- [ ] Documentation update

Fixes # (issue)

## Checklist

### DIFF
--- a/examples/nextjs-swc/package.json
+++ b/examples/nextjs-swc/package.json
@@ -7,13 +7,13 @@
     "debug": "NODE_OPTIONS='--inspect' next dev",
     "build": "yarn lang:extract && next build",
     "start": "next start",
-    "lang:extract": "lingui extract --clean",
+    "lingui:extract": "lingui extract --clean",
     "test": "yarn build"
   },
   "dependencies": {
     "@lingui/core": "^4.1.2",
     "@lingui/react": "^4.1.2",
-    "next": "13.2.3",
+    "next": "13.3.1",
     "react": "18.2.0",
     "react-dom": "18.2.0"
   },
--- a/examples/nextjs-swc/src/components/Developers.tsx
+++ b/examples/nextjs-swc/src/components/Developers.tsx
@@ -1,5 +1,5 @@
 import { useState } from 'react'
-import { Trans, plural } from '@lingui/macro'
+import { Trans, Plural } from '@lingui/macro'
 
 export default function Developers() {
   const [selected, setSelected] = useState('1')
@@ -17,10 +17,7 @@ export default function Developers() {
           <option value={'2'}>2</option>
         </select>
         <p>
-          {plural(selected, {
-            one: 'Developer',
-            other: `Developers`
-          })}
+          <Plural value={selected} one={'Developer'} other={`Developers`} />
         </p>
       </div>
     </div>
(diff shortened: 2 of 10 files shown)


examples/nextjs-swc/src/components/Developers.tsx
```diff
@@ -17,10 +17,7 @@ export default function Developers() {
           <option value={'2'}>2</option>
         </select>
         <p>
-          {plural(selected, {
-            one: 'Developer',
-            other: `Developers`
-          })}
+          <Plural value={selected} one={'Developer'} other={`Developers`} />
```
thekip, line 20: @yunsii whenever non-react macro (t, plural, select, etc) is used in react component, `useLingui()` hook should be added as well. OR just use JSX macro instead. 

That caused a translation mismatch you noticed in the issue. 

Here i've just changed it to JSX macro, because it's more appropriate usage.

examples/nextjs-swc/src/locales/en.po
```diff
@@ -13,34 +13,35 @@ msgstr ""
...
+msgid "{selected, plural, one {Developer} other {Developers}}"
+msgstr "{selected, plural, one {Developer} other {Developers}}"
+
+#: src/components/Switcher.tsx:10
 msgid "English"
 msgstr "English"
 
+#. js-lingui-explicit-id
+#: src/components/AboutText.tsx:6
+msgid "next-explanation"
```
thekip, line 26: @yunsii for explicit custom ids `lingui@4` requires `#. js-lingui-explicit-id` comment to be added. This was a root cause why this section was not translated.

examples/nextjs-swc/package.json
```diff
@@ -7,13 +7,13 @@
...
     "start": "next start",
-    "lang:extract": "lingui extract --clean",
+    "lingui:extract": "lingui extract --clean",
     "test": "yarn build"
   },
   "dependencies": {
     "@lingui/core": "^4.1.2",
     "@lingui/react": "^4.1.2",
-    "next": "13.2.3",
+    "next": "13.3.1",
```
thekip, line 16: The release version of `@lingui/swc-plugin` was built with a newer version of the `swc-core`, which is not compatible with `next@13.2.3"` this caused a "panic" during build. I put a compatible version and added a compatibility notice into readme.

andrii-bodnar (approved):

andrii-bodnar:
@thekip thank you!

examples/nextjs-swc/src/components/Developers.tsx
yunsii: It means if i don't use i18n, should still call useLingui hook like below?

```tsx
import { useLingui } from "@lingui/react"
import { plural } from '@lingui/macro'

function App() {
  useLingui()
  //...

  return (
    <p>
      {plural(selected, {
        one: 'Developer',
        other: `Developers`
      })}
    </p>
  )
}
```

thekip: yes. plural / t / select / selectOrdinal macro is simply a syntax sugar over i18n, it is expanded to the `i18n._` call during build.

I recommend to use react-related macros whenever possible, because it auto subscribe for a changes and less error-prone.

yunsii: Theaks for your detail explanation 👌

Issue #1717, opened 2023-06-26
Experimental Syntax 'importAttributes' Not Enabled


zvictor:
**Description**

When running the `lingui extract` command, the system encounters a syntax error due to the experimental syntax 'importAttributes' not being enabled. This error prevents the extraction of messages from source files that use the importAttributes syntax, such as:

```ts
import data from './data.json' assert { type: 'json' };
```

**To Reproduce**
1. Start a new Lingui project (https://lingui.dev/tutorials/setup-react)
2. Make use of 'importAttributes':
```shell
echo "import data from './data.json' assert { type: 'json' };" > src/bugged.js
```
3. Run the command `./node_modules/.bin/lingui extract --verbose`
4. Observe the error message related to the experimental syntax 'importAttributes'

```haskell
Extracting messages from source files…
⠋ Cannot process file /private/tmp/lingui-bug/src/bugged.js /private/tmp/lingui-bug/src/bugged.js: Support for the experimental syntax 'importAttributes' isn't currently enabled (1:32):

> 1 | import data from './data.json' assert { type: 'json' };
    |                                ^
  2 |

Add @babel/plugin-syntax-import-attributes (https://github.com/babel/babel/tree/main/packages/babel-plugin-syntax-import-attributes) to the 'plugins' section of your Babel config to enable parsing.
SyntaxError: /private/tmp/lingui-bug/src/bugged.js: Support for the experimental syntax 'importAttributes' isn't currently enabled (1:32):

> 1 | import data from './data.json' assert { type: 'json' };
    |                                ^
  2 |

Add @babel/plugin-syntax-import-attributes (https://github.com/babel/babel/tree/main/packages/babel-plugin-syntax-import-attributes) to the 'plugins' section of your Babel config to enable parsing.
    at instantiate (/private/tmp/lingui-bug/node_modules/@babel/parser/lib/index.js:63:32)
    at constructor (/private/tmp/lingui-bug/node_modules/@babel/parser/lib/index.js:358:12)
    at JSXParserMixin.raise (/private/tmp/lingui-bug/node_modules/@babel/parser/lib/index.js:3255:19)
    at JSXParserMixin.expectOnePlugin (/private/tmp/lingui-bug/node_modules/@babel/parser/lib/index.js:3301:18)
    at JSXParserMixin.maybeParseImportAttributes (/private/tmp/lingui-bug/node_modules/@babel/parser/lib/index.js:14243:14)
[...]

PR #1719, opened 2023-06-28
feat(extractor): expose extractFromFileWithBabel function for better flexibility

thekip:
# Description

This PR brings improvement for those who wants more flexibility in what files extractor could understand. It's covering https://github.com/lingui/js-lingui/issues/1717

When user has a Ecmascript feature which is not enabled in the parser by default there is only thing it could do about it.  Override extractor completely and pass all needed plugins manually. This is not future-proof solution, as user have to copy-paste lingui's default implementation and adjust it by its needs. 

Solution: I extracted from the default implementation a low level function `extractFromFileWithBabel` which is now a part of a public api and could be used in the userland.  I choosed this approach among others, because `parserPlugins` is not a static list and could vary depending on a file passed for extractor. So this would give to users enough flexibility to express any required logic.

## Types of changes

- [ ] Documentation update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/cli/src/api/extractors/babel.ts
+++ b/packages/cli/src/api/extractors/babel.ts
@@ -1,12 +1,16 @@
-import { DEFAULT_EXTENSIONS, transformAsync } from "@babel/core"
+import { DEFAULT_EXTENSIONS, transformAsync, ParserOptions } from "@babel/core"
 
-import type { ExtractPluginOpts } from "@lingui/babel-plugin-extract-messages"
+import type {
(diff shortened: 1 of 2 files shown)


andrii-bodnar:
Thanks @thekip! What do you think about adding this to the documentation? However, I have doubts about where exactly it can be added. Maybe [Lingui Configuration](https://lingui.dev/ref/conf)?

andrii-bodnar (approved):

PR #1743, opened 2023-08-15
feature(react): separate entry point for RSC

thekip:
# Description

Implementation step for: https://github.com/lingui/js-lingui/issues/1698

It turned out that RSC implementation in Nextjs complains when it meets any `creacteContext` function call in module import graph. So having a separate component which is not using context is not enough. It should be two completely separated entry points (files) and server entry point should not be polluted with `createContext` calls otherwise it shows an error: 

>  createContext only works in Client Components. Add the "use client" directive at the top of the file to use it.

This PR brings additional entrypoint: 

```ts
import { TransNoContext } from '@lingui/react/server';
```

Which is not intended to be used directly by end-users, but as part of future nextjs integration. 

## Types of changes

- [ ] Documentation update

Fixes # (issue)

## Checklist


vonovak (approved):
looks good to me!

PR #1777, opened 2023-10-04
docs: add explicit vs generated id doc. Move to non-global string-only translations

thekip:
# Description

The main goal was to remove inconsistency in string-only macro usage (`t` and etc) 

Because of historical reasons we have many different ways to achieve the same result. Some of these approaches a considered harmful and it's better to avoid them and not advertise in the docs. 

For example these are gives mostly the same result: 

```ts
t`Hello`
t(i18n)(`Hello`)
i18n.t(msg`Hello`)
i18n._(msg`Hello`)

// in react context
const { _ } = useLingui();
_(msg`Hello`)

const { i18n } = useLingui();
i18n.t(msg`Hello`)
i18n._(msg`Hello`)
```

Where usage of global `t` is very confusing and bring "delayed" issues, which then hard to fix. 

So I propose to gradually removing global usage from docs, and encourage developers to always use i18n methods on a particular instance. 

While working on achieving this goal, i've found that  "explicit vs generated id" could be extracted from "React Patterns" because it's more general.  And also could be a good place to gather related information which is currently spread across the whole docs.

## Types of changes

- [ ] Documentation update

Fixes # (issue)

## Checklist

### DIFF
--- a/website/docs/ref/macro.md
+++ b/website/docs/ref/macro.md
@@ -778,11 +778,11 @@ const message = /*i18n*/ {
 
 ### `Trans`
 
-| Prop name | Type   | Description                                                                                     |
-| --------- | ------ | ----------------------------------------------------------------------------------------------- |
-| `id`      | string | Custom message ID                                                                               |
-| `comment` | string | Comment for translators                                                                         |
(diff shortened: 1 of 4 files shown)


andrii-bodnar (approved):
@thekip thank you!

PR #1796, opened 2023-10-28
feat: `id` syntax for t macro

krizka:
```
t('some.id')`message`;
```

New short syntax to set id and message in the same time. In our project we giving ids to every message we have, and all the time I lack short `t` syntax with id.

# Description

## Types of changes

- [ ] Documentation update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/macro/src/macroJs.ts
+++ b/packages/macro/src/macroJs.ts
@@ -111,6 +111,34 @@ export default class MacroJs {
       return false
     }
 
+    // t(`id`)`message` -> i18n._({ id: "id", message: "message" })
+    if (
+      this.types.isCallExpression(path.node) &&
+      this.types.isTaggedTemplateExpression(path.parent) &&
+      (this.types.isStringLiteral(path.node.arguments[0]) ||
+        this.types.isTemplateLiteral(path.node.arguments[0])) &&
+      this.isLinguiIdentifier(path.node.callee, "t")
+    ) {
+      const idTokens = this.types.isTemplateLiteral(path.node.arguments[0])
+        ? this.tokenizeTemplateLiteral(path.node.arguments[0])
+        : [
+            {
+              type: "text",
+              value: this.expressionToArgument(path.node.arguments[0]),
+            } as TextToken,
+          ]
+      const msgTokens = this.tokenizeTemplateLiteral(path.parent.quasi)
+
+      const descriptor = this.createMessageDescriptorFromIdMsg(
+        idTokens,
+        msgTokens,
+        path.node.loc
+      )
+
+      path.replaceWith(descriptor)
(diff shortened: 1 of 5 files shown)


packages/macro/src/macroJs.ts
```diff
@@ -111,6 +111,34 @@ export default class MacroJs {
...
+      this.isLinguiIdentifier(path.node.callee, "t")
+    ) {
+      const idTokens = this.types.isTemplateLiteral(path.node.arguments[0])
+        ? this.tokenizeTemplateLiteral(path.node.arguments[0])
+        : [
+            {
+              type: "text",
+              value: this.expressionToArgument(path.node.arguments[0]),
+            } as TextToken,
+          ]
```
thekip, line 129: You don't need to parse id node into tokens, tokens are used only for message. You need to just check if it's string literal or tpl literal and pass this node directly to message descriptor factory.

packages/macro/src/macroJs.ts
```diff
@@ -418,6 +446,31 @@ export default class MacroJs {
...
+      this.createValuesProperty(values),
+    ]
+
+    return this.createMessageDescriptor(
+      properties,
+      // preserve line numbers for extractor
+      oldLoc
+    )
+  }
+
```
thekip, line 473: Let's DRY it, and instead of creating a separate function, just extend `createMessageDescriptorFromTokens` by adding an optional `id?: Node`

```ts
  createMessageDescriptorFromTokens(tokens: Tokens, oldLoc?: SourceLocation, idNode?: Node) {
    const { message, values } = buildICUFromTokens(tokens)
    
        const properties: ObjectProperty[] = [
          id ? this.createObjectProperty(ID, idNode) : this.createIdProperty(message),
    
```

website/docs/ref/macro.md
```diff
@@ -139,6 +139,20 @@ i18n._(
 );
 ```
 
+```js
+t('attachment.saved')`Attachment ${name} saved`;
```
thekip, line 143: Does it allow id to be only StringLiteral ot TplLiteral as well?  If Template literal also possible it should be in tests as well.

thekip (changes requested):
I left few important comments, but my general fillings about that - i don't really like adding one more syntax to the already complex and complicated set. This brings a maintenance burden and not really aligned with future plans. 

The idea that global `t` function (and it's siblings as `plural`) don't really play well with modern React (and not only) ecosystem which moves towards SSR + RSC. And even with SPA it brings many misusage and questions from the users.  

So i'm going to make global `t` usage discouraged in favor of using a particular instance of i18n. 

```ts
t(myI18n)`Hello`
```

PR #1808, opened 2023-11-27
fix(cli): sorting issue when different locales are used on host machines

Kepro:
# Description

`localeCompare` produces different sort output based on user machine locales

e.g. colleague in Denmark `da-DK`, me `en-US`
examples:
```
"Opening Entry".localeCompare("Opening entry"); // 1 (en-US)
"Opening Entry".localeCompare("Opening entry"); // -1 (da-DK)
```
this was causing different orders of messages committed by different users (we are using format-json minimal)
we have translations like examples above when difference is only in casing
`Opening Entry` vs `Opening entry`

## Types of changes

- [ ] Documentation update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/cli/src/api/catalog.ts
+++ b/packages/cli/src/api/catalog.ts
@@ -342,7 +342,7 @@ export function orderByMessage<T extends ExtractedCatalogType>(messages: T): T {
     .sort((a, b) => {
       const aMsg = messages[a].message || ""
       const bMsg = messages[b].message || ""
-      return aMsg.localeCompare(bMsg)
+      return aMsg.localeCompare(bMsg, "en-US")
     })
     .reduce((acc, key) => {
       ;(acc as any)[key] = messages[key]


packages/cli/src/api/catalog.ts
```diff
@@ -342,7 +342,7 @@ export function orderByMessage<T extends ExtractedCatalogType>(messages: T): T {
     .sort((a, b) => {
       const aMsg = messages[a].message || ""
       const bMsg = messages[b].message || ""
-      return aMsg.localeCompare(bMsg)
+      return aMsg.localeCompare(bMsg, "en-US")
```
Kepro, line 345: source language would be preferably always English in many projects, but still point is to have "stable" order in output

thekip: Is there any drawbacks of hardcoding "en-US" as default collator locale? I'm trying to come up with something, but only thinks that could produce not-accurate sorting for some users which is fine comparing to unstable sort.

pkasarda: yeah, we were investigating the code, and testing more versions, but this is the simplest as we don't have any information for example about "sourceLocale" in that function without bigger refactoring, but not it's really unstable and after upgrading to v4 we had a lot of complains about this issue :X

Kepro: yeah, we were investigating the code, and testing more versions, but this is the simplest as we don't have any information for example about "sourceLocale" in that function without bigger refactoring, but not it's really unstable and after upgrading to v4 we had a lot of complains about this issue :X

thekip:
Good catch. I also refresehd documentation of `localeComapre` method for myself and thinking may be it's better to implement what they say there:

>When comparing large numbers of strings, such as in sorting large arrays, it is better to create an [Intl.Collator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator) object and use the function provided by its [compare()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/compare) method.

Catalogs could be quite large, so instead of creatiing collator inside sort(), you could create it outside of the loop.

Kepro:
@thekip should I update PR? it is okay to use `en`?

packages/cli/src/api/catalog.ts
thekip: I actually could make this "big" refactoring myself and pass a locale to the compare function from processed catalog, but the question - is it worth it?

Kepro: my 2c people will don't care about small sorting issues, if the order stay stable, so probably not worth it...

thekip:
I think it's fine. Just add a comment with explanation why we hard-code the en-US and probably link to this PR, just to quicker find the reason in future.

andrii-bodnar (approved):
@Kepro thank you!

pkasarda:
YW! but looks like your size testing has also order problems :trollface:

PR #1809, opened 2023-11-27
test: debug

andrii-bodnar:


andrii-bodnar:
@vonovak oh, it seems like removing the parallel flag has fixed the issue 🎉 

 For some reason, the build order of packages was changed in the GH Actions environment

PR #1820, opened 2023-12-17
fix: allow parentheses in catalog pathnames

### LINKED ISSUE #1819: [Vite] Parentheses in catalog pathnames causes them to be unrecognized
**Describe the bug**
Importing a catalog that has `parentheses` in the pathname throws the following error in the Vite integration:

`Error: Requested resource <pathname_with_parentheses> is not matched to any of your catalogs paths specified in "lingui.config".`

**To Reproduce**

`await import("./(foo).catalog.po")`

Minimal reproduction in a Remix.run app (using the unstable Vite integration): [repl](https://github.com/jozefmery/lingui-issue-repr)

**Expected behavior**
The catalog should be recognized correctly regardless of special characters and imported  successfully.

**Additional context**

Why do I care about parentheses in pathnames? I would like to generate a separate catalog for each route, so that each of them can load their own translations with minimal network overhead. However, Remix uses parentheses in pathnames for defining optional parameters, causing the imports to fail.

Finally, I'd like to thank the authors for creating the sexiest I18N solution there is. Thank you.

jozefmery:
# Description

Escape parentheses in the pathname pattern according to the following issue: https://github.com/micromatch/micromatch/issues/123
The solution uses the simplest possible approach: the String#replace method. I considered the `escape-string-regexp` package, but it seems that `parentheses` are the only problematic characters.

## Types of changes

- [ ] Documentation update

Fixes #1819

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/lingui/js-lingui/blob/main/CONTRIBUTING.md) and [CODE_OF_CONDUCT](https://github.com/lingui/js-lingui/blob/main/CODE_OF_CONDUCT.md) docs (NOTE: I managed to run `yarn test`, but `yarn release:test` fails with the following message: `The filename, directory name, or volume label syntax is incorrect. command not found: test:tsd`. I believe this is unrelated to my changes and may warrant its own issue.)

### DIFF
--- a/packages/cli/src/api/catalog/getCatalogs.test.ts
+++ b/packages/cli/src/api/catalog/getCatalogs.test.ts
@@ -407,4 +407,22 @@ describe("getCatalogForFile", () => {
       catalog,
     })
   })
+
+  it("should allow parentheses in path names", async () => {
+    const catalog = new Catalog(
+      {
+        name: null,
+        path: "./src/locales/(asd)/{locale}",
+        include: ["./src/"],
+        format,
+      },
+      mockConfig({ format: "po", rootDir: "." })
+    )
+    const catalogs = [catalog]
+
+    expect(getCatalogForFile("./src/locales/(asd)/en.po", catalogs)).toEqual({
+      locale: "en",
+      catalog,
+    })
+  })
 })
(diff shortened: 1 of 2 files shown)


thekip (approved):

andrii-bodnar (approved):
@jozefmery thank you!

PR #1850, opened 2024-02-14
fix: don't replace octothorpe coming from variable

### LINKED ISSUE #1826: Using Trans inside a Plural macro unexpectedly replaces "#" found in variables
**Describe the bug**
Using `Trans` with the `Plural` macro can lead to variables being rendered in a modified way. It should be possible for a variable to contain "#" without it being replaced with the `value` of the `Plural` macro.

**To Reproduce**
Define a component, which uses a nested `Trans` within the `Plural` macro:
```jsx
import {Plural, Trans} from '@lingui/macro'

function Notification({count, documentTitle}) {
  return (
    <p>
      <Plural
        value={count}
        one={<Trans>There is a notification in <span>{documentTitle}</span></Trans>}
        other={<Trans>There are # notifications in <span>{documentTitle}</span></Trans>}
      />
    </p>
  )
}
```

**Expected behavior**
Rendering the component using this code:
```jsx
<Notification
  count={1}
  documentTitle="Title #1"
/>
```
Should result in
```html
<p>There is a notification in <span>Title #1</span></p>`
```
but it generates
```html
<p>There is a notification in <span>Title 11</span></p>
``` 
As you can see, the "#" found in the `documentTitle` variable is replaced with the count "1". It renders as `Title 11` instead of `Title #1`.

I think the expected behavior is to leave variables untouched and render them as is. Only hashtags that are directly part of the translation string should be replaced.

**Additional context**

- jsLingui version `4.6.0`
- Babel version `7.23.3`
- I'm using Babel with `babel-macro-plugin`

thekip:
resolves #1826

# Description

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/core/src/interpolate.test.ts
+++ b/packages/core/src/interpolate.test.ts
@@ -58,6 +58,25 @@ describe("interpolate", () => {
     expect(plural({ value: 2 })).toEqual("2 Books")
   })
 
+  it("should not replace `#` symbol passed in the variable in the jsx expression", () => {
+    const plural = prepare(
+      "{value, plural, one {There is a notification in <1>{documentTitle}</1>} other {There are # notifications in <1>{documentTitle}</1>}}"
+    )
+
+    expect(plural({ value: 1, documentTitle: "Title #1" })).toEqual(
+      "There is a notification in <1>Title #1</1>"
+    )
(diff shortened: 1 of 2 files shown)


andrii-bodnar (approved):
@thekip thank you!

Issue #1852, opened 2024-02-19
Improve React syntax with useLingui


dan-dr:
**Is your feature request related to a problem? Please describe.**
I'm implementing lingui for my company's product, with DX being among the top priority.
The syntax when using the lingui hook is quite clunky.

First, there are 2 ways to translate (correct me if i'm wrong):
```typescript
import { t, msg } from '@lingui/macro';

// react component
() => {
  const { i18n, _ } = useLingui();
  const one = _(msg`First way to translate`);
  const two = t(i18n)`Second way to translate`;
}
```

Second, each of these ways are somewhat "special" syntax with 2 parts each (`_` and `msg`, `t` and `i18n`).

I find the DX and readability sub-par with the high quality of this library and it's tooling.

**Describe proposed solution**
Basically I'd like to do something like:
```typescript
import { useLingui } from '@lingui/macro';

// react component
...
  const { i18n, _, t } = useLingui();
  const justOneWay = t`Very Cool`;
...
```

I'm guessing the macro should transform this to:
```typescript
import { useLingui } from '@lingui/react';

const { i18n, _ } = useLingui();
const justOneWay = _({ ...message descriptor... });
```

Not sure exactly about the naming or the macro API. maybe it's better to separate the hooks
```typescript
const { t } = useTranslate()
t`String`;
↓ becomes
const { _ } = useLingui();
_({...descriptor});
```
but then you probably need to account for other use of useLingui and merge the declaration, which isn't impossible.

**Describe alternatives you've considered**
I tried to play around with a custom hook to wrap, but I don't think it's possible with the macro implementation.
neither of these hacks are possible (and probably for a good reason:
```typescript
  const t = React.useCallback(
    (literals: TemplateStringsArray, ...placeholders: any[]) => {
      return _(defineMessage(literals, placeholders));
      nor
     return t(i18n)(literals, placeholders);
    },
    [.. dependencies],
  );
```

**Additional context**
I'm not familiar with macro implementation, else I'd give it a whirl myself...

thekip:
The current design here for a reason. Indeed, the design proposed by you is better and I even investigated to possibility to do something like that before. However, there are technical restrictions on the babel-macro side and also this kind of transpilation opens a wide variety of usages user's could do with these functions which we could not know ahead of time and it would be a big maintenance burden to handle all possible cases in transpilation. 

[...]

Issue #1797, opened 2023-10-30
`lingui extract-experimental` bugs


yunsii:
**Describe the bug**

- Extracted message id incorrect when `Trans` children has plain text follow JSX element, space is losted

![image](https://github.com/lingui/js-lingui/assets/18096089/6d2547fd-7dcc-4e51-9f20-24e7d852cb4a)

- Dependency tree related files (AboutText.tsx/Developers.tsx) do not extracted in reproduction, but Switcher.tsx did

**To Reproduce**

https://github.com/yunsii/lingui-examples-nextjs-swc

**Expected behavior**

- Extract correct message id
- Extract full dependency tree related files

- jsLingui version `v4.5.0`
- Macro support:
 - [x] I'm using SWC with `@lingui/swc-plugin`
 - [ ] I'm using Babel with `babel-macro-plugin`
 - [ ] I'm not using macro

thekip:
Well, i've found a problem why some files not get processed by extractor. Unfirtunately it's not easy to fix, because it's need to change literally everthying in how this is processed now. 

The simple explanation of the process now: 

1. Files bundled together with esbuild, JSX is preserved
2. Regular JS extractor is applied to the bundles
   1. Macro applied to whole bundle
   2. Extractor extract from all usages as it was written without macro 

The problem here is that esbuild while bundling doing a lot of work related to module imports. It renames, deduplicate and etc.  Macro is able to understand renamed imports, but macro is also responsible of turning macro input into runtime import. And here where problems appears. Macro is not smart enough to analyze what import already inserted in the file and in what order (the esbuild could generate very complicated setups, because what actually bundlers do)

So after macro expannding some imports got deleted and extractor did not recognize these nodes as belonging to the Lingui
Here is the output result after intermediate steps: 

<details>
  <summary>Esbuild bundle</summary>
```js
// src/pages/index.page.tsx
import { Plural as Plural2, t as t2, Trans as Trans3 } from "@lingui/macro";
import path from "path";
import Head from "next/head";

// src/components/AboutText.tsx
import { Trans } from "@lingui/macro";
function AboutText() {
  return <p>
    <Trans>Hello, world</Trans>
    <br />
    <Trans id="message.next-explanation">Next.js is an open-source React front-end development web framework that enables functionality such as server-side rendering and generating static websites for React based web applications. It is a production-ready framework that allows developers to quickly create static and dynamic JAMstack websites and is used widely by many large companies.</Trans>
  </p>;
}

[...]

PR #1859, opened 2024-02-22
feat(macro): add useLingui macro

timofei-iatsenko:
#1852

Adds new macro `useLingui` which is drastically simplify working with non-jsx messages in react components: 

```ts
import { useLingui } from '@lingui/macro';
function MyComponent() {
  const { t } = useLingui();
  const a = t`Text`;
}
```
becomes: 
```ts
function MyComponent() {
  const { _: _t } = useLingui();
  const a = _t(
    /*i18n*/
    {
      id: "xeiujy",
      message: "Text",
    }
  );
}`
```

All syntax flavors of original `t` function is also supported. 

Also, it's possible to use returned `t` as dependency in react hooks: 

```ts
import { useLingui } from '@lingui/macro';
function MyComponent() {
  const { t } = useLingui();
  const a = useMemo(() => t`Text`, [t]);
}

// becomes:
function MyComponent() {
  const { _: _t } = useLingui();
  const a = useMemo(
    () =>
      _t(
        /*i18n*/
        {
          id: "xeiujy",
          message: "Text",
        }
      ),
    [_t]
  );
}
```

The implementation still has some not covered cases, but I believe it should cover 95% regular cases and  it's already good enough to give it a try. 

# Description

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

### DIFF
--- a/website/docs/tutorials/react.md
+++ b/website/docs/tutorials/react.md
@@ -672,7 +672,7 @@ The last message in our component is again a bit specific:
 `lastLogin` is a date object, and we need to format it properly. Dates are formatted differently in different languages, but we don't have to do this manually. The heavy lifting is done by the [Intl object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl), we'll just use [`i18n.date()`](/docs/ref/core.md#i18n.date) function. The `i18n` object can be accessed by [`useLingui`](/docs/ref/react.md#uselingui) hook:
 
 ```jsx title="src/Inbox.js"
-import { useLingui } from "@lingui/macro";
+import { useLingui } from "@lingui/react";
 
 export default function Inbox() {
   const { i18n } = useLingui();


packages/macro/src/macroJs.ts
```diff
@@ -467,8 +562,10 @@ export default class MacroJs {
     return (
       this.types.isTaggedTemplateExpression(node) &&
       (this.isLinguiIdentifier(node.tag, "t") ||
+        this.isLinguiIdentifier(node.tag, "_") ||
         (this.types.isCallExpression(node.tag) &&
-          this.isLinguiIdentifier(node.tag.callee, "t")))
+          (this.isLinguiIdentifier(node.tag.callee, "t") ||
+            this.isLinguiIdentifier(node.tag.callee, "_"))))
```
thekip, line 568: This solution is temporary here to make recursive macro matching work.  What is really needed is extend `isRootPath` to understand calls of function returned by `useLingui` 

Other option would be use babel-macro only as "backdoor" which will allow running arbitrary transformation code and then write whole functionality as regular babel plugin with visitor pattern. 

A little bit of context here: 

babel's visitor traverse the AST in depth first algorithm. That completely eliminates a problem with nested macro calls such as: 

```js
t`Hello ${plural(...)}`
```

Because we can process them from root and expand recursively. 

However, with babel macro plugin we receive list of "references" which is flat and in undefined order. 
For example above we will receive 2 references `t` and `plural` and we should process only `t` and ignore `plural` since it's a part of outer `t` expression.

dan-dr:
Seems to have the same issue I had when extracting:
`{t`What ${plural(3, { zero: 'yes', other: 'no' })} Up`}`
image

maybe it's a problem with the extractor, but i'm pretty sure its a more a tokenizer/macro issue.
Also, there's (maybe rare) cases where you want the i18n from the useLingui, which currently isn't supported. i.e.:
```typescript
const { i18n, _ } = useLingui();
const { t } = useLinguiMacro();
const [locale, setLocale] = React.useState(i18n.locale);
```

dan-dr:
thinking about it, I think a better way may be when the references are processed in the index file, add the `t` nodes thatt are from `useLingui` to the `jsNodes` array.

dan-dr:
refactored your PR per my last comment:
https://github.com/dan-dr/js-lingui/tree/feature/use-lingui-macro

In my implementation:
- I prefer that the macro imitates the standard `useLingui` but replaces the `_` function. so if you need `i18n` you can use the macro as well
- pre-process useLingui in index file
  - set needsUseLinguiImport to true
  - report incorrect macro syntax
  - add non-array expression references to destructured `t` function to `jsNodes`
    - this eliminates any recursive macro issues
  - add "normal" conditions to process these references in `macroJs`
  - file-scope rename the `t` destruction to a unique identifier and add to nameMap to support multiple components in same file

thekip:
@dan-dr i cherry-picked your changes to the branch of this PR. 

I'm going to re-check your changes, add more tests, and mark PR as ready when I finish. Thanks for help and contributing.

thekip:
1. Added `usLingui` to `LinguiConfig.runtimeConfigModules` to be consistent with other macro
2. Added tests for validation errors
3. Fixed typings and add typetest for `i18n` usage from `useLingui`
4. added extractor test cases for `useLingui`

There still issue if the renamed import of useLingui is existing in the file and has renamed specifier, but the same error would be with any other macro, and i dont rembmer any user have issue with that.  

```ts
import { useLingui as useLinguiMacro } from '@lingui/macro';
import { useLingui as useLinguiRenamed  } from '@lingui/react';

function MyComponent() {
  const { _ } = useLinguiRenamed();

  console.log(_);
  const { t } = useLinguiMacro();
  const a = t`Text`;
}
```

I'm going to fix it in next iterations in another PR's. I'm already working on refactoring which will help with [experimental extractor](https://github.com/lingui/js-lingui/issues/1797) and fix edge cases with renamed imports as well.

thekip:
I updated documentation for useLingui macro. I feel that documentation could be better, but i did my best honestly.

dan-dr:
sweet, waiting for release :)

website/docs/ref/react.md
```diff
@@ -130,6 +130,23 @@ const CurrentLocale = () => {
 };
 ```
 
+:::tip
+There is a [macro version](/ref/macro#uselingui) of `useLingui` hook. Which supports all features of [`t` macro](/docs/ref/macro.md#t) and uses runtime `useLingui` hook under the hood.
```
vonovak, line 134:
```suggestion
There is a [macro version](/ref/macro#uselingui) of the `useLingui` hook which supports all features of the [`t` macro](/docs/ref/macro.md#t) and uses the runtime `useLingui` hook (from `@lingui/react`) under the hood.
```

website/docs/ref/react.md
```diff
@@ -130,6 +130,23 @@ const CurrentLocale = () => {
...
+
+const CurrentLocale = () => {
+  const { t } = useLingui();
+
+  const userName = "Tim";
+  return <span>{t`Hello ${userName}`}</span>;
+};
+```
+
+You also can safely use returned `t` function in a dependency array of React hooks.
```
vonovak, line 147:
```suggestion
You also can safely use the returned `t` function in a dependency array of React hooks.
```

website/docs/tutorials/javascript.md
```diff
@@ -53,7 +53,7 @@ i18n.activate("en");
 
 ## Localizing your app
 
-Once that is done, we can go ahead and use it! Wrap you text in [`t`](/docs/ref/macro.md#t) macro and pass it to [`i18n._()`](/docs/ref/core.md#i18n._) method:
+Once that is done, we can go ahead and use it! Wrap you text in [`t`](/docs/ref/macro.md#t) macro:
```
vonovak, line 56:
```suggestion
Once that is done, we can go ahead and use it! Wrap your text in [`t`](/docs/ref/macro.md#t) macro:
```

website/docs/tutorials/react-patterns.md
```diff
@@ -84,6 +96,8 @@ function MyComponent() {
 }
 ```
 
+Note that we import `useLingui` component from `@lingui/macro`. There is also a runtime version of `useLingui` hook exported from `@lingui/react`. In the case above, it doesn't matter what version to choose since we use only `i18n` object which is presented in both.
```
vonovak, line 99:
```suggestion
Note that we import `useLingui` from `@lingui/macro`. There is also a runtime version of `useLingui` hook exported from `@lingui/react`. In the case above, it doesn't matter what version to choose since we use only `i18n` object which is presented in both.
```

website/docs/tutorials/react-patterns.md
```diff
@@ -272,3 +243,22 @@ export function Welcome() {
   return <div>{welcome}</div>;
 }
 ```
+
+:::note
+Note on [`useLingui`](/ref/macro#uselingui) macro usage. The `t` function destructured from this hook, behaves the same way as `_` from runtime [`useLingui`](/ref/react#uselingui) counterpart, so you can safely use it in the dependency array.
```
vonovak, line 248:
```suggestion
Note on [`useLingui`](/ref/macro#uselingui) macro usage. The `t` function destructured from this hook, behaves the same way as `_` from the runtime [`useLingui`](/ref/react#uselingui) counterpart, so you can safely use it in the dependency array.
```

website/docs/tutorials/react.md
```diff
@@ -304,6 +304,45 @@ It isn't necessary to extract/translate messages one by one. This usually happen
...
+
+We have this piece of code in our example:
+
+```js
+const markAsRead = () => {
+  alert("Marked as read.");
+};
+```
+
+To translate it we will use a `useLingui` macro hook:
```
vonovak, line 319:
```suggestion
To translate it, we will use the `useLingui` macro hook:
```

packages/macro/test/js-useLingui.ts
```diff
@@ -0,0 +1,380 @@
...
+      message: "Text",
+    }
+  );
+}
+`,
+  },
+
+  {
+    // todo: implement this
+    skip: true,
```
vonovak, line 275: just flagging this

packages/macro/src/index.ts
```diff
@@ -110,54 +121,132 @@ function macro({ references, state, babel, config }: MacroParams) {
...
+          t.isIdentifier(property.value) &&
+          property.key.name == "t"
+      )
+    : null
+
+  if (!_property) {
+    reportUnsupportedSyntax(
+      path.parentPath.parentPath,
+      new Error(
+        `Must destruct \`t\` when using \`useLingui\` macro, i.e:
```
vonovak, line 192:
```suggestion
        `You have to destructure \`t\` when using the \`useLingui\` macro, i.e:
```

can we print the usage that triggered this error, to make the message more useful?

packages/conf/src/migrations/normalizeRuntimeConfigModule.test.ts
```diff
@@ -1,62 +1,77 @@
...
+  test("all defaults", () => {
+    const actual = normalizeRuntimeConfigModule({ runtimeConfigModule: {} })
 
-    expect(actual.runtimeConfigModule).toStrictEqual({
-      TransImportModule: "@lingui/react",
-      TransImportName: "Trans",
-      i18nImportModule: "../my-i18n",
-      i18nImportName: "myI18n",
-    })
+    expect(actual.runtimeConfigModule).toMatchInlineSnapshot(`
```
vonovak, line 7: In this whole file, I'd prefer that we revert to using `toStrictEqual` instead of `toMatchInlineSnapshot` because it's both more readable and more "writable".

snapshots are strings. In your IDE, they will display as a bunch of green lines and they are hard to read. 

And have you ever tried writing a snapshot? Again, because it's just a string, you get no help from the language grammar, syntax highlights...

It's not a big deal so feel free to ignore but I think the right use cases for snapshots are elsewhere.

packages/macro/src/index.ts
```diff
@@ -110,54 +121,132 @@ function macro({ references, state, babel, config }: MacroParams) {
...
  Error: ${e.message}`
   )
 }
 
+/**
+ * Pre-process useLingui macro
+ * 1. Get references to destructed t function for macro processing
+ * 2. Transform usage to non-macro useLingui
+ *
+ * @returns Array of paths to useLingui's t macro
```
vonovak, line 151: I don't understand the `returns` comment TBH. Could you expand a little on what "path" is in this context?

packages/macro/src/index.ts
```diff
@@ -110,54 +121,132 @@ function macro({ references, state, babel, config }: MacroParams) {
...
+ Example:
+
+ const { t } = useLingui()
+`
+      )
+    )
+    return null
+  }
+
+  const varDec = path.parentPath.parentPath.node
```
vonovak, line 172: The code in this file is the "meat" of the macro transform.

I'd appreaciate if we could make this as readable as possible for future developers.



`varDec`: what does that mean? Can we give this a clearer name?

packages/macro/src/index.ts
```diff
@@ -110,54 +121,132 @@ function macro({ references, state, babel, config }: MacroParams) {
...
+ const { t } = useLingui()
+ or
+ const { t: _ } = useLingui()
+ `
+      )
+    )
+    return null
+  }
+
+  if (t.isIdentifier(path.node)) {
```
vonovak, line 202: here, I'd like to ask again for more help to anyone reading this. What does this condition cover? Can we give a name to `t.isIdentifier(path.node)` so that the condition is less "magical"?

packages/macro/src/index.ts
```diff
@@ -110,54 +121,132 @@ function macro({ references, state, babel, config }: MacroParams) {
...
+    path.scope.rename(path.node.name, "useLingui")
+  }
+
+  // rename to standard useLingui _
+  _property.key.name = "_"
+  path.scope.rename(_property.value.name, newIdentifier)
+
+  return path.scope
+    .getBinding(_property.value.name)
+    ?.referencePaths.filter(
```
vonovak, line 213: if there's a way to break this into several logical steps that'd be appreciated but if not, ignore. I don't want to bother too much :D

packages/macro/src/macroJs.ts
```diff
@@ -148,13 +151,62 @@ export default class MacroJs {
       this.isLinguiIdentifier(path.node.callee, "t")
     ) {
       this.replaceTAsFunction(path as NodePath<CallExpression>)
+      this.needsI18nImport = true
+
       return true
     }
 
+    // { t } = useLingui()
```
vonovak, line 159: I'm sorry, I don't really follow these comments (and the conditions neither)

if the comment were a little longer that'd be nice.

vonovak (commented):
very exciting! thank you!

Question: what about the version for SWC?

I left a few annoying comments 🙈 😄

packages/macro/src/index.ts
```diff
@@ -110,54 +121,132 @@ function macro({ references, state, babel, config }: MacroParams) {
...
+): NodePath[] | null {
+  if (!path.parentPath.parentPath.isVariableDeclarator()) {
+    reportUnsupportedSyntax(
+      path,
+      new Error(
+        `\`useLingui\` macro must be used in variable declaration.
+        
+ Example:
+
+ const { t } = useLingui()
```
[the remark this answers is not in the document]
thekip: Yes, this happened automatically. This error is caught and re-thrown with babel-code-frame

packages/macro/test/js-useLingui.ts
thekip: 😬 thanks, i missed that one.

packages/macro/src/index.ts
thekip: the same answer as previously

packages/conf/src/migrations/normalizeRuntimeConfigModule.test.ts
thekip: I'm a big fan of snapshots. You don't need to touch it manually ever. I changed it to snapshots because it is much easier to maintain, when i added couple new lines into expected output i don't need to go and change every test manually, i just run it with `-u`. That saves a lot of time.

packages/macro/src/macroJs.ts
thekip: It's easier to follow the logic when you read whole file not only diff, there is a list of possible syntaxes:

image

This comments is extensions to this list. Purpose of this comment is to find an entry point of specific syntax to transform. 

I could agree that the code is not very much readable, but this is how usually transforms looks like. They a very wordy and not readable, there is not much we can do about it.

thekip:
> Question: what about the version for SWC?

I'm going to start working on that when we stabilize this API in babel version. I'm not a pro Rust developer, so still feel myself very uncomfortable there. Don't want to re-write this few times.  Also, babel's API is much more feature-rich as opposite to SWC. All this `getBinding()` or `getReferences()` does not exist in SWC, and i expect implementing this feature would take much more effort that in babel.

packages/macro/test/js-useLingui.ts
thekip: Yes, i wrote about this case in the comments. I'm working on next PR which will fix it

packages/macro/test/js-useLingui.ts
```diff
@@ -0,0 +1,380 @@
...
+
+function MyComponent() {
+  const { t } = useLingui();
+  const a = t\`Text\`;
+}
+`,
+    expected: `
+import { myUselingui } from "@my/lingui-react";
+function MyComponent() {
+  const { _: _t } = useLingui();
```
thekip, line 367: just found that this is incorrect, it imports `myUselingui` but uses `useLingui`

packages/macro/src/index.ts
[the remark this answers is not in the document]
thekip: image

packages/macro/src/index.ts
dan-dr: Mostly of the `isXXXX` from babel types act as type guard, giving us a way to get access to properties of a concrete type. In this case we need to access `name` of `path.node`, to which you wouldn't have had access to without this check.

Not sure how we can "name" it.

packages/macro/src/index.ts
```diff
@@ -110,54 +121,132 @@ function macro({ references, state, babel, config }: MacroParams) {
...
 
-  const linguiImport = state.file.path.node.body.find(
+  const linguiImport = body.find(
     (importNode) =>
       t.isImportDeclaration(importNode) &&
       importNode.source.value === module &&
       // https://github.com/lingui/js-lingui/issues/777
       importNode.importKind !== "type"
-  ) as ImportDeclaration
+  ) as t.ImportDeclaration
```
dan-dr, line 237:
```suggestion
  const linguiImport = body.find(
    (importNode) : importNode is t.ImportDeclaration =>
      t.isImportDeclaration(importNode) &&
      importNode.source.value === module &&
      // https://github.com/lingui/js-lingui/issues/777
      importNode.importKind !== "type"
  )
```

thekip:
@andrii-bodnar when it can be merged? I'm thinking of merging it to the main and creating a pre-release (with `-next` or `-beta` suffix) Want to tests it thoroughly in the ecosystem first.

packages/macro/src/index.ts
thekip: it's a matter of taste

packages/conf/src/migrations/normalizeRuntimeConfigModule.test.ts
vonovak: okay :). glad to have my opinion challenged 👍

packages/macro/src/index.ts
vonovak: well, a condition always represents some scenario. If the scenario is given a name or a comment I think more people will understand it.

// rename to standard useLingui

and what are we renaming?

Feel free to ignore though

thekip:
So let's merge it, so i can concentrate on this https://github.com/lingui/js-lingui/pull/1867

vonovak (approved):

website/docs/tutorials/react-native.md
```diff
@@ -112,12 +112,14 @@ We're importing the default `i18n` object from `@lingui/core`. Read more about t
 
 Translating the heading is done. Now, let's translate the `title` prop in the `<Button title="mark messages as read" />` element. In this case, `Button` expects to receive a `string`, so we cannot use the [`Trans`](/docs/ref/macro.md#trans) macro here!
 
-The solution is to use the `t` macro together with the `i18n` object which we can obtain from the `useLingui` hook. We use the two like this to get a translated string:
+The solution is to use the `t` macro which we can obtain from the `useLingui` hook. We use it like this to get a translated string:
```
andrii-bodnar, line 115: `useLingui` macro?

website/docs/tutorials/react-native.md
```diff
@@ -128,21 +130,21 @@ The interplay of `I18nProvider` and `useLingui` is shown in the following simpli
 
 ```tsx
 import { I18nProvider } from "@lingui/react";
-import { t, Trans } from "@lingui/macro";
+import { Trans, useLingui } from "@lingui/macro";
```
andrii-bodnar, line 133: The sentence above looks like not relevant anymore

```
If we cannot use `Trans`, we can use the `useLingui` hook to get hold of the `i18n` instance ourselves and get the translations from there.
```

at least, there we talk about hook, but use the macro in the example

website/docs/tutorials/react.md
```diff
@@ -304,6 +304,45 @@ It isn't necessary to extract/translate messages one by one. This usually happen
 
 For more info about CLI, checkout the [CLI tutorial](/docs/tutorials/cli.md).
 
+## Non-JSX Translation
+
+Until now learned how to translate string inside a JSX elements, but what if we want to translate something that is not in a JSX? Or pass a translation as prop to another component?
```
andrii-bodnar, line 309:
```suggestion
So far we learned how to translate string inside a JSX element, but what if we want to translate something that is not inside a JSX? Or pass a translation as a prop to another component?
```

website/docs/tutorials/react.md
```diff
@@ -304,6 +304,45 @@ It isn't necessary to extract/translate messages one by one. This usually happen
...
+
+We have this piece of code in our example:
+
+```js
+const markAsRead = () => {
+  alert("Marked as read.");
+};
+```
+
+To translate it, we will use the `useLingui` macro hook:
```
andrii-bodnar, line 319:
```suggestion
To translate it, we will use the `useLingui` macro:
```

website/docs/tutorials/react.md
```diff
@@ -304,6 +304,45 @@ It isn't necessary to extract/translate messages one by one. This usually happen
...
+const { t } = useLingui();
+
+const markAsRead = () => {
+  alert(t`Marked as read.`);
+};
+```
+
+Now the `Marked as read.` message would be picked up by extractor, and available for translation in the catalog.
+
+You also could pass variables and use any other macro in the message.
```
andrii-bodnar, line 335:
```suggestion
Now the `Marked as read.` message would be picked up by extractor, and available for translation in the catalog.

You could also pass variables and use any other macro in the message.
```

website/docs/tutorials/react.md
```diff
@@ -633,7 +672,7 @@ The last message in our component is again a bit specific:
 `lastLogin` is a date object, and we need to format it properly. Dates are formatted differently in different languages, but we don't have to do this manually. The heavy lifting is done by the [Intl object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl), we'll just use [`i18n.date()`](/docs/ref/core.md#i18n.date) function. The `i18n` object can be accessed by [`useLingui`](/docs/ref/react.md#uselingui) hook:
 
 ```jsx title="src/Inbox.js"
-import { useLingui } from "@lingui/react";
+import { useLingui } from "@lingui/macro";
```
andrii-bodnar, line 675: the sentence above also need to be updated, it still describes the react hook

andrii-bodnar (commented):
@thekip @dan-dr @vonovak thanks for your efforts and contributions!

Overall it looks good to me, but I was concerned about the confusion it might cause for developers trying to understand the difference between the `useLingui` macro and the `useLingui` React hook. Even now, it's a bit confusing at the documentation level.

The `use<something>` is a typical naming convention for React hooks. In our case, the `useLingui` macro might be confusing because of this naming convention for hooks.

Perhaps we need to better document the difference or consider choosing a different name for the macro. 

What do you think guys?

website/docs/tutorials/react-native.md
thekip: `useLingui` macro is a hook as well.  Could be used in the same context where runtime's `useLingui` used. 

This similar to Trans, which has two versions, macro and runtime. 

I think naming is fine, but documentation could be improved.

website/docs/tutorials/react-native.md
thekip: Agree this paragraph should be rephrased. Could you help with it? The idea, i think, if we are not in JSX we could use `const { t } = useLingui()` macro

website/docs/tutorials/react.md
thekip: why not? it's a valid react hook

website/docs/tutorials/react.md
thekip: i think we can revert here import from react, since it doesn't matter in this case what flavor to use.

thekip:
> The `use<something>` is a typical naming convention for React hooks. In our case, the `useLingui` macro might be confusing because of this naming convention for hooks.
> 
> Perhaps we need to better document the difference or consider choosing a different name for the macro.

`useLingui` macro is a hook as well. Could only be used in the context where runtime's useLingui used. So `use<something>` is a valid name for it. 

This similar to Trans, which has two versions, macro and runtime. 

I think naming is fine, but documentation could be improved.

thekip:
Guys, i believe we can improve documentation any time later, that should not block merging the feature.

website/docs/tutorials/react-native.md
andrii-bodnar: Will check after merging

andrii-bodnar:
@thekip thanks for the clarification! Could you please address the 3 remaining discussions?

https://github.com/lingui/js-lingui/pull/1859#discussion_r1503411665
https://github.com/lingui/js-lingui/pull/1859#discussion_r1507258926
https://github.com/lingui/js-lingui/pull/1859#discussion_r1507850748

thekip:
@andrii-bodnar @vonovak regarding naming, i think we can rise the discussion. I believe the initial idea for macro was a "subset" of the original params from runtime versions (as original author [said](https://github.com/lingui/js-lingui/issues/1433#issuecomment-1467778710)).  So they signatures originally had to match to each other.

But since then react/js ecosystem shifted towards typescript and quality typings and high type safety become a more valuable than ever. That resulted in changes in lingui, where runtime and macro version got they own typings with different signatures to highlight for developers different usages of macro/runtime versions. 

So now we have Trans from /react and /Trans from macro which are diffrent components. This PR will add useLingui from /macro in additional to /react. 
This is indeed confusing. It's also confusing for IDEs, they can automatically add import to incorrect symbol and user may spend hours to understand why it doesn't work as expected. 

I think, maybe it's time to change the naming to something more explicit? 

`Trans` / `TransM` and `useLingui` / `useLinguiM`

The Plural Select and SelectOrdinal doesn't have runtime counterparts, they are traspiled to Trans, but i believe they should be renamed to follow the same naming convention: 

`PluralM` `SelectM` and `SelectOrdinalM`

What do you think?

thekip:
@andrii-bodnar 

> Then, I am going to merge this to the next branch and release 4.8.0 with the next tag.

Could next release be created from "main" branch? I don't see a reason for branching here

packages/macro/src/index.ts
thekip: its variable declarator, please ignore this in favor of https://github.com/lingui/js-lingui/pull/1867

thekip:
@andrii-bodnar 
>@thekip thanks for the clarification! Could you please address the 3 remaining discussions?
>
>https://github.com/lingui/js-lingui/pull/1859#discussion_r1503411665
>https://github.com/lingui/js-lingui/pull/1859#discussion_r1507258926
>https://github.com/lingui/js-lingui/pull/1859#discussion_r1507850748

Reverted the docs, and just resolved rest, there are nothing to do about them.

andrii-bodnar:
> I think, maybe it's time to change the naming to something more explicit?

Sounds interesting. Let's move this to a separate discussion and consider it in planning the v5 release. Currently, I don't have other ideas, but the `M` suffix looks a bit weird to me.

> Could next release be created from "main" branch? I don't see a reason for branching here

When merging into the `main` branch, the documentation will be automatically deployed to lingui.dev. Also, if I remember correctly, the `next` release should be triggered from the `next` branch according to the Lerna configuration.

andrii-bodnar:
I was going to release it today but GitHub is currently experiencing an outage with Actions 😬 

Probably better to do it on Monday.

andrii-bodnar:
@thekip https://github.com/lingui/js-lingui/releases/tag/v4.8.0-next.0

PR #1874, opened 2024-03-06
feat(macro): support transforming member expressions to named args

dan-dr:
# Description

Helps extract complex member expressions to named arguments with `_` delimiter.
so `` t`Very {result.cool} Stuff` ``  results in `Very {result_cool} Stuff`

I wanted to add this because we have a lot of results from APIs that I really don't want to break down to different variables.

WIP, wanted to see thoughts of maintainers

TODO:
- Add opt-in option in the lingui config, because it will break previous extractions. name suggestion?
- Update tests
- Docs

## Types of changes

- [ ] Documentation update
- [ ] Examples update

## Checklist

### DIFF
--- a/packages/macro/src/macroJs.ts
+++ b/packages/macro/src/macroJs.ts
@@ -10,6 +10,7 @@ import {
   SourceLocation,
   StringLiteral,
   TemplateLiteral,
+  MemberExpression,
 } from "@babel/types"
 import { NodePath } from "@babel/traverse"
 
@@ -367,11 +368,31 @@ export default class MacroJs {
     }
   }
 
+  memberExpressionToArgument(exp: MemberExpression): string {
+    let parts: string[] = []
+
+    if (this.types.isThisExpression(exp.object)) {
+      // ignore this
+    } else {
+      parts.unshift(this.expressionToArgument(exp.object))
+    }
+
+    parts.push(
+      this.expressionToArgument(
+        this.types.isPrivateName(exp.property) ? exp.property.id : exp.property
+      )
+    )
+
+    return parts.join("_")
+  }
+
   expressionToArgument(exp: Expression): string {
     if (this.types.isIdentifier(exp)) {
       return exp.name
     } else if (this.types.isStringLiteral(exp)) {
       return exp.value
+    } else if (this.types.isMemberExpression(exp)) {
+      return this.memberExpressionToArgument(exp)
     } else {
       return String(this._expressionIndex())
     }
--- a/packages/macro/src/macroJsx.ts
+++ b/packages/macro/src/macroJsx.ts
@@ -1,6 +1,7 @@
 import * as babelTypes from "@babel/types"
 import {
   ConditionalExpression,
+  MemberExpression,
   Expression,
   JSXAttribute,
   JSXElement,
@@ -285,6 +286,7 @@ export default class MacroJSX {
       if (exp.isJSXElement()) {
         return this.tokenizeNode(exp)
       }
+
       return [this.tokenizeExpression(exp)]
     } else if (path.isJSXElement()) {
       return this.tokenizeNode(path)
@@ -464,8 +466,35 @@ export default class MacroJSX {
     }
   }
 
+  memberExpressionToArgument(path: NodePath<MemberExpression>): string {
+    let parts: string[] = []
+
+    if (this.types.isThisExpression(path.node.object)) {
+      // ignore this
+    } else {
+      parts.unshift(this.expressionToArgument(path.get("object")))
+    }
+
+    const propertyPath = path.get("property")
+
+    parts.push(
+      this.expressionToArgument(
+        propertyPath.isPrivateName() ? propertyPath.get("id") : propertyPath
+      )
+    )
+    return parts.join("_")
+  }
+
   expressionToArgument(path: NodePath<Expression | Node>): string {
-    return path.isIdentifier() ? path.node.name : String(this.expressionIndex())
+    if (this.types.isIdentifier(path.node)) {
+      return path.node.name
+    } else if (this.types.isStringLiteral(path.node)) {
+      return path.node.value
+    } else if (path.isMemberExpression()) {
+      return this.memberExpressionToArgument(path)
+    } else {
+      return String(this.expressionIndex())
+    }
   }
 
   /**


thekip:
I thought about something more generic, what about adding a special macro `ph` or `var` to manually specify the name of the placeholder? 

```ts
t`Very ${ph(result.cool, 'myName')} Stuff` // -> Very {myName} Stuff
```

This would also work for evey other type of the expressions:
```ts
// call expression
t`Very ${ph(callExpression(), 'myName')} Stuff`

// this expression
t`Very ${ph(this.myVar, 'myName')} Stuff`
```

Does it make sense?

dan-dr:
I think that kinda ruins readability and DX, the variable identifier should already be descriptive  enough.
What we can do is have the option to specify depth. So for example depth 2 will output name_first when given user.name.first

thekip:
By the way, 

this is valid JS: 

```js
 {
  'app.foo.bar': "World"
}
```
As well as this is valid ICU:

```icu
Hello, {app.foo.bar}!
```

So may be if it's a MemberExpression just get it as is, without replacing `.` to `_` and other modifications? 

```
t`Very ${this.myVar} Stuff` // -> Very {this.myVar} Stuff
```

dan-dr:
> So may be if it's a MemberExpression just get it as is, without replacing `.` to `_` and other modifications?
right, can be done. maybe also should be configured. I was thinking translators will have easier time with underscore.

btw my implementation ignores the `this` part

thekip:
> right, can be done. maybe also should be configured. I was thinking translators will have easier time with underscore.

less configuration, less branching, less maintenance burden. I don't see any benefit of `_` instead of original `.` (dots) 
The translation platform should do quality checks and helps with variables for translators.

> btw my implementation ignores the this part
 
i saw this, also I think it's obvious and expected from developers point of view if this `{app.foo.bar}` extracted as is, why `{this.foo.bar}` is not? Should be consistently. 

> regarding the rule, I think it shouldn't effect it? developers should decide if they want it or not, especially since this will be opt-in due to breaking change.

It should affect. The purpose of this rule is enforcing developers NOT to write expressions (including member expressions) in the messages due to lack context information and potential poor quality of the translation. If member expression will be processed as proposed in this PR, this rule should be relaxed to skip such cases. Or configuration added, but i prefer to keep them in sync, so the latest version of eslint plugin is aligned with latest version of the lib. 

> what's the stance on swc implementations? it's already diverging due to the useLingui macro we did, but will there be an official release without swc implementation? Is there any knowledgable rust maintainer that can port it?

I'm the author of that swc implemetation. The useLingui is opt-in hook, where change from this PR will affect essential macros. 

I would prefer to not do a lot of configuration and just include it in lingui 5 as default. SWC implementation should also be ready till that moment.

andrii-bodnar:
@dan-dr thank you for the contribution!

I see a couple of threats here:

- What if we have a few of the same strings with different variables passed? Will this result in duplicate strings with different placeholders and be treated as separate strings during extraction? If so, this could lead to excessive translation effort, as translators will translate these strings separately.
- In case the variable changes for any reason (renaming, refactoring, etc.), the string will be different. As a result, the previous translations may be lost (TM pre-translate with placeholder auto-substitution should then be applied, otherwise this string will be translated again and it will lead to additional translation costs).

thekip:
@andrii-bodnar all these points are true for current implementation as well. This PR doesn't bring something new, just extending the cases wich macro / extractor could process.

There is a tradeoff between passing more information to translator and re-usability of the message. 

Angular's team, for example, solves this problem a bit differently, they replace arguments in the string into positional placeholders but also saves meta information how these placeholders are mapped. Something like:

```po
# ph1: app.foo.bar
# ph2: myFunction()
msgid: Hello {ph1} and {ph2}
```
So translators could get the point of what variable is in here by looking into metadata.  

Currently, if developer want to reuse the string, he should provide the same name for all placeholders, sometimes this might require to introduce an intermediate variable for that: 

```js
t`Hello ${userName}` // -> Hello {userName} 
t`Hello ${getUserName()}` // -> Hello {0}


// Developer may reuse the translation by introducing a variable:
const userName = getUserName();
t`Hello ${userName}`  // ->  Hello {userName}  match to the string from case 1
```

So, maybe uncouple variable name from placeholder name not that bad idea even considering worse DX. 

In my example if developer will specify a placeholder name it would be stable between refactorings and would be more explicit than using a variable name
```js
t`Hello ${ph(getUserName(), 'userName')}`
t`Hello ${ph(userName, 'userName')}`
```

By the way, Angular's team solves this issues as well, they have introduce a special magic syntax for that, which is for my test is worse than using additional macro function
```js
$localize`Hello ${getUserName()}:userName` // `:userName` after the interpolation is a special "magic" syntax used in Angular to give name for placeholders
```

dan-dr:
I like it a bit more, JS-y. will make it optional. I agree on the eslint rule btw.
So the general thought it to make this a v5 feature? Really don't want to maintain a fork :D

thekip:
I thought about that for a while and here is my conclusion:

- We can make what you propose, and include in v4 with an additional configuration flag, disabled by default. This flag would be true by default in lingui5 (and even flag itself might be deleted)
- What I proposed (`` t`Hello ${{userName: getName()}}` ``) is a nice to have, addition and could be implemented as a separate feature in other PR. 

In v5 we can make a general configuration which will turn on/off using variable name in placeholder: 

```tsx
// useVaraibleNameAsPlaceholder: true (default)
t`Hello ${userName}` // -> Hello {userName} 
t`Hello ${user.name}` // -> Hello {user.name} 

// useVaraibleNameAsPlaceholder: false
t`Hello ${userName}` // -> Hello {0} 
t`Hello ${user.name}` // -> Hello {0} 
``` 

This has more sense than enabling/disabling particular syntax processing. Those users who want as much string as possible to be reused could opt out from named placeholders, or use explicit placeholders syntax `` t`Hello ${{userName: getName()}}` ``


The member expression parts should go as-is to the placeholder, no need to replace dots to dashes or cutting levels. 
Test cases: 
```tsx
t`Very ${this.myVar} Stuff` // -> Very {this.myVar} Stuff
t`Hello ${user.name}` // -> Hello {user.name}

// any level deeper should be supported
t`Hello ${foo.bar.baz}` // -> Hello {foo.bar.baz}

// don't forget about JSX
<Trans>Hello {foo.bar.baz}</Trans>  // -> Hello {foo.bar.baz}
```

Also rebase your branch to the `next` base, there is massive refactoring of the macro code.

dan-dr:
Question: How do translation platforms usually deal with arg changes?
example: `Text with {0}` that turns to `Text with {var}` 

Because I think that in v5 you really should make sure people can upgrade and keep the same "behavior". in what you propose, if someone is currently using both `userName` and `user.name`, they will have no way to upgrade to v5 without it causing changes in their messages. That should be avoided imo.

thekip:
That's why it's planned for v5.  We could not go forward if we would need to keep all this backward compatibility. 

Translation platforms will usually treat these as two separate strings, but with help of Translation Memory the translation could be retrieved back pretty quick.

timofei-iatsenko:
Want to resurrect this issue and share few thoughts. I still like this feature, and #1965 doesn't replace this one. 

1. i just realized that introducing a flag such as `useVaraibleNameAsPlaceholder` would be a first time when a macro option would affect extraction. Existing options affect transformation of the macro but these transformations don't affect extraction. This means that extractor and transform options should be synchronized otherwise message keys would be mismatched. I'm pretty sure that there are already users which has these options out of sync and they even don't know about this. Couple of examples how that inconsistency may happen

- When user use swc plugin. Extractor will use options from lingui.config and SWC plugin from parameters passed directly to it. 
- When user applies macro in packages which doesn't use the same lingui config, or lingui config not properly discovered
- When user uses `babel-plugin-lingui-macro` directly (v5) and pass options directly to it instead of lingui config

So this is option should be carefully rethought. 

2. I tried to implement a placeholder name as proposed 
```js
t`Hello ${{userName: getName()}}`
```

But realized that this will not work for JSX, because JSX typings will not allow an object as a children (ReactNode) and we can extend a typing for `<Trans>` itself to accept `ReactNode | Record<string, string>`, but it will still not work for inner tags: 

```jsx
<Trans>Hello <strong>{{userName: user.name}}</strong></Trans>
```

So the only two options left is a: 

With a function: 

```js
t`Hello ${ph(user.name, 'userName')}`
```

Or with a magic comment: 

```js
t`Hello ${/* ph: userName */user.name}`
```

I did a little vote between developers, and 9 out of 9 developers voted for the function, because comment fills to magically, despite the verbosity of the function.

PR #1899, opened 2024-03-29
Update Trans.tsx to 'use client'

mhfaust:
This only works in client components anyway. Why not add 'use client' here so that it doesn't blow up when composed in server components?

# Description

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/react/src/Trans.tsx
+++ b/packages/react/src/Trans.tsx
@@ -1,3 +1,5 @@
+'use client'
+
 import React from "react"
 
 import { useLinguiInternal } from "./I18nProvider"


fromthemills:
Thank you for this! We should also add `"use client"` tot the `I18nProvider` as it is also considered a client component.

thekip:
So the thing is, NexJS team doesn't provide library authors with a clear guideline how to incorporate this "use client" in the usual library building pipeline. 

The problem is, code of the library transpiled and bundled into one file, so this "use client" will not appear at the top of the module, and will not work as you are expecting.

fromthemills:
Yes this a thing... bundler support for `"use client"` is subpar for now. In the Nextjs docs they talk about it and have [some examples](https://nextjs.org/docs/app/building-your-application/rendering/composition-patterns#advice-for-library-authors) of how to do it with esbuild I think.

thekip:
Well, it seems the "proposed way" is to just turn the whole bundle into "use client".  Since we have a separate entrypoint for `TransNoContext` that may work...

andrii-bodnar:
Closed in favor of #1902

PR #1937, opened 2024-05-15
fix(message-utils): correctly handle multi-digit complex argument cases

Bertg:
# Description

Current version of lingui will not handle following pluralisation correctly: `{count, plural, one {One} other {#} =13 {thirteen}}`. This PR fixes that.

I also added some tests, but maybe some dedicated test should be made, I'm not sure what the maintainer expect here.

Did not report the issue separately, as I already had the fix.

## Types of changes

- [ ] Documentation update
- [ ] Examples update

## Checklist

### DIFF
--- a/packages/core/src/interpolate.test.ts
+++ b/packages/core/src/interpolate.test.ts
@@ -39,10 +39,12 @@ describe("interpolate", () => {
 
   it("should interpolate plurals", () => {
     const plural = prepare(
-      "{value, plural, one {{value} Book} other {# Books}}"
+      "{value, plural, one {{value} Book} =4 {Four books} =99 { Books with problems } other {# Books}}"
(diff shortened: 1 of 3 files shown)


Bertg:
I also want to add an other suggestion here. 

I noticed that the cause of this bug was a wrong regex, and regexes are hard. We could replace that regex with this operation instead: `item.key[0] === '=' ? item.key.slice(1) : item.key`. I think this has several advantages:

* Not a regex! so easier to understand
* More intention revealing (now it is clear that this also handles cases not starting with a `=`
* A lot faster, like a lot... JSBench tells me the regex method is 92.48 % slower (based on micro test cases)

thekip (approved):
Nice job. Thanks

PR #1938, opened 2024-05-20
fix: (lingui/core) i18n error if id is undefined

### LINKED ISSUE #1926: i18n._(id) throws error if id param passed as undefined
**Describe the bug**

`i18n._(id)` throws error if id param passed as undefined

**To Reproduce**

This is just an example. 

```jsx

export default function ErrorComponent(error) {
   return (
        <p>{i18n._(error) || <Trans>There is an error</Trans>}</p>
   )
}
```

**Expected behavior**

- if error is type of string, it should return and show translated message
- if error is undefined, it should return undefined so that or condition message displays.

**Additional context**

It was working in `v2.7.2` 
- `typeof id === "object"` check - returns `false`, if id is undefined as expected

```js
  _(
    id: string | Object,
    values: Object = {},
    { defaults, formats = {} }: MessageOptions = {}
  ) {
    // Expand message descriptor
    if (id && typeof id === "object") {
      values = id.values
      defaults = id.defaults
      formats = id.formats
      id = id.id
    }
```

in version > `3.0.0`, this got changed. 
- `!isString(id)` check - returns `true` ,if id is undefined. so it further checks values and throws error

```ts
  _(descriptor: MessageDescriptor): string
  _(id: string, values?: Values, options?: MessageOptions): string
  _(
    id: MessageDescriptor | string,
    values?: Values,
    options?: MessageOptions
  ): string {
    let message = options?.message
    if (!isString(id)) {
      values = id.values || values
      message = id.message
      id = id.id
    }
```

- jsLingui version ``lingui --version`` 4.7.0
- Babel version ``npm list @babel/core`` 7.15.5
- Macro support:
 - [ ] I'm using SWC with `@lingui/swc-plugin`
 - [x] I'm using Babel with `babel-macro-plugin`
 - [ ] I'm not using macro
- Your Babel config (e.g. `.babelrc`) or framework you use (Create React App, NextJs, Vite)

abdul-kissflow:
# Description

Lingui Core - i18n._(id) throws error if id param passed as undefined

## Types of changes

- [ ] Documentation update
- [ ] Examples update

 Fixes #1926 

- Added id check before checking its object values

## Checklist

### DIFF
--- a/packages/core/src/i18n.test.ts
+++ b/packages/core/src/i18n.test.ts
@@ -290,6 +290,25 @@ describe("I18n", () => {
     expect(handler).toHaveBeenCalledTimes(2)
   })
 
+  it("._ should emit missing event for undefined id", () => {
+    const i18n = setupI18n({
+      locale: "en",
+      messages: { en: {} },
+    })
+
+    const messageID = undefined
+
+    const handler = jest.fn()
+    i18n.on("missing", handler)
+    // @ts-ignore
+    i18n._(messageID)
+    expect(handler).toHaveBeenCalledTimes(1)
+    expect(handler).toHaveBeenCalledWith({
+      id: undefined,
+      locale: "en",
+    })
+  })
+
   describe("params.missing - handling missing translations", () => {
     it("._ should return custom string for missing translations", () => {
       const i18n = setupI18n({
--- a/packages/core/src/i18n.test.ts
+++ b/packages/core/src/i18n.test.ts
@@ -296,15 +296,13 @@ describe("I18n", () => {
       messages: { en: {} },
     })
 
-    const messageID = undefined
-
     const handler = jest.fn()
     i18n.on("missing", handler)
     // @ts-expect-error 'id' should be of 'MessageDescriptor' or 'string' type.
-    i18n._(messageID)
+    i18n._()
     expect(handler).toHaveBeenCalledTimes(1)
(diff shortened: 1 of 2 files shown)


thekip:
Thanks for contribution. Could you add a test, otherwise it might be broken again in the future.

abdul-kissflow:
> Thanks for contribution. Could you add a test, otherwise it might be broken again in the future.

Added test case for this case @thekip.

Screenshot 2024-05-21 at 3 10 51 PM

packages/core/src/i18n.test.ts
```diff
@@ -290,6 +290,25 @@ describe("I18n", () => {
...
+      locale: "en",
+      messages: { en: {} },
+    })
+
+    const messageID = undefined
+
+    const handler = jest.fn()
+    i18n.on("missing", handler)
+    // @ts-ignore
+    i18n._(messageID)
```
thekip, line 304: if it's exepcted to have a TS error here, put `@ts-expect-error` with explanation why it's expected.

abdul-kissflow: changed to 
// @ts-expect-error 'id' should be of 'MessageDescriptor' or 'string' type.

thekip:
you have issues in the pipeline

abdul-kissflow:
> you have issues in the pipeline

Fixed it @thekip

thekip (approved):

PR #1946, opened 2024-05-29
docs(examples): add Remix Vite + Babel example

depsimon:
# Description

This PR adds a Remix example that works with Vite & the Babel plugin.

I integrated a modified version of [remix-i18n](https://github.com/sergiodxa/remix-i18next) that makes it easy to detect the locale from multiple sources in Remix.

This should probably live in its own package at some point.


examples/remix-vite-babel/app/root.tsx
```diff
@@ -0,0 +1,62 @@
...
+    { headers: {
+      'Set-Cookie': await localeCookie.serialize(locale)
+    } },
+  );
+}
+
+export async function loader({ request }: LoaderFunctionArgs) {
+  const locale = await linguiServer.getLocale(request);
+
+  return json(
```
ericchernuka, line 28: I wonder if an example of setting a title from the loader would be a good use case?

ericchernuka (commented):
@depsimon this is great! I liked the detection code. It's a nice tie back to how Remix i18next's detection is abstracted so it'll be easy to knowledge transfer. Just a single question regarding a title example.

examples/remix-vite-babel/app/root.tsx
depsimon: Good idea. I added this in the example, thanks for the suggestion

andrii-bodnar (approved):
@depsimon thank you! 🚀

PR #1961, opened 2024-06-17
docs: Update cli.md

maisano-patreon:
Update description for `--strict` to align with CLI output.

# Description

The docs are seemingly out of date here. I've updated the ref for `lingui compile --strict` to match that of the CLI.

## Types of changes

- [x] Documentation fix

## Checklist

### DIFF
--- a/website/docs/ref/cli.md
+++ b/website/docs/ref/cli.md
@@ -178,7 +178,7 @@ Overwrite source locale translations from source.
 
 #### `--strict` {#compile-strict}
 
-Fail if a catalog has missing translations.
+Disable defaults for missing translations.
 
 #### `--format <format>` {#compile-format}
 


thekip:
Looking into the code it seems whats in the docs is right. Because cli will exit with non-zero code if any translation is missing. 

https://github.com/lingui/js-lingui/blob/main/packages/cli/src/lingui-compile.ts#L47-L70

maisano-patreon:
@thekip I haven't spent much time actually debugging this, but this code path isn't ever reached in my testing – `onMissing` isn't firing despite having a number of empty translations. I'll see if I can look into this later today.

maisano-patreon:
Oh – it never fires because the fallback locale is used to resolve the translation in this case. Not sure what's really supposed to happen here, but it feels a bit loose. I would expect the `--strict` flag to fail if a catalog doesn't have the translation, regardless of the fallback.

maisano-patreon:
I don't mind this being closed – unfortunately I don't have the time to dig through the source to understand what the flag is actually doing, but the current form of the docs is misleading/broken as `--strict` does not fail when translations are missing.

andrii-bodnar:
@maisano-patreon Sorry for the inconvenience. Could you please create an issue describing this? I think it should be investigated further and addressed properly. Thank you!

PR #1962, opened 2024-06-19
fix: include type cast on ts compilation

tagleeson:
# Description
In typescript projects which are using [ts-reset](https://github.com/total-typescript/ts-reset) and `compileNamespace: "ts"`, the compiled output is invalid because the type of `JSON.parse` is changed to `unknown` from the normal `any`, meaning that you cannot just use a type annotation and must use an explicit cast. There doesn't seem to be any simple configuration that could fix this issue, so this PR with an explicit type cast seems to be a logical answer. 

It doesn't affect anyone who is not using `ts-reset`, and expands compatibility for those who are.

## Types of changes

- [ ] Documentation update
- [ ] Examples update

## Checklist


thekip:
So type of the const is not needed then:

```diff
- export const messages:Messages=JSON.parse();
+ export const messages=(JSON.parse()) as Messages;

```

tagleeson:
@thekip I was trying to make as minimal of a change as possible, happy to remove the unnecessary annotation. Updated.

thekip (approved):

Issue #1963, opened 2024-06-21
`lingui extract <..files>` overwrites obsolete flag


nicolas-cherel:
When using `lingui extract <some file>` the result overwrite the obsolete flag of all the messages in the catalog to `false`. This is unexpected because `lingui extract` only adds `obsolete: true` flags and we probably don't want to overwrite this.

Eg `lingui extract <file with obsolete translation>` will produce the following patch that overwrite the obsolete flag
```
         20
       ]
     ],
-    "obsolete": true
+    "obsolete": false
   },
```

Also Eg `lingui extract <file with changed translation>` will produce the following change on listed files
```
+    ],
+    "obsolete": false
```

**To Reproduce**
do a `lingui extract <some file>` on a project with several files and see that `"obsolete": false` are added to catalog to translation of unlisted files, or have their `"obsolete": true` changed to `"obsolete": false`

**Expected behavior**
`obsolete` obsolete flag should be kept pristine when a list of file is supplied to the command line

**Code fix proposal**
Modify the line at https://github.com/lingui/js-lingui/blob/v4.11.1/packages/cli/src/api/catalog/mergeCatalog.ts#L52
from `obsolete: !options.files,` to `...(!options.files && { obsolete: true }),`
in order to do nothing on obsolete flags for files not listed on command line.

PR #1964, opened 2024-06-21
fix(extract): when files are used, don't overwrite obsolete

nicolas-cherel:
# Description
When using `lingui extract <...files>` all translations in catalog that are not in the listed file get their obsolete flag to false, regardless if the obsolete flag was not set or already set to true.

## Types of changes

Fixes # (issue)
#1963

## Checklist

### DIFF
--- a/packages/cli/test/index.test.ts
+++ b/packages/cli/test/index.test.ts
@@ -4,8 +4,8 @@ import extractExperimentalCommand from "../src/lingui-extract-experimental"
 import { command as compileCommand } from "../src/lingui-compile"
 import fs from "fs/promises"
 import os from "os"
-import nodepath from "path"
 import glob from "glob"
+import nodepath from "path"
 import { makeConfig } from "@lingui/conf"
 import { listingToHumanReadable, readFsToJson } from "../src/tests"
 import { getConsoleMockCalls, mockConsole } from "@lingui/jest-mocks"
@@ -180,12 +180,6 @@ describe("E2E Extractor Test", () => {
         "extractor-experimental"
       )
 
-      await fs.cp(
-        nodepath.join(rootDir, "existing"),
-        nodepath.join(rootDir, "actual"),
-        { recursive: true }
-      )
-
       await mockConsole(async (console) => {
         const config = makeConfig({
           rootDir: rootDir,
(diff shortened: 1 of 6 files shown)


andrii-bodnar:
@nicolas-cherel thanks for the contribution! Please include tests that cover this use case

thekip:
i would recommend to add full e2e test here `packages/cli/test` 

1. Create a folder with test case name
2. In `fixtures` put test files to extract (you can copy from sibling test case)
3. in `existing` put catalogs with already extracted messages an obsolete flag
4. in `expected` put how catalogs should like after extracting
5. add separate describe into `packages/cli/test/index.test.ts` and run `extractCommand` with parameters you need

nicolas-cherel:
Note: it seem that only the lingui json format has support for the obsolete flag, I hope depending on it for the tests is not a issue

thekip:
Po files also support obsolete, i would prefer to have it with PO files, lingui json format is not recommended to use by docs.

nicolas-cherel:
Ok I'll I must have missed something on my first pass, I get right away now that I have the json version set up

nicolas-cherel:
Ok, note, we previous had an issue specific to the json files where a `"obsolete": true` was added to translations with partial extract, which does not happen with .po files due to the format specificities. I manage to witness the obsolete flag being removed with the partial extract though, so that fix is confirmed.

I renamed the test suite fixtures/exisiting/expected folder to `extract-partial-consitency` as the general problem was more a consistency issue with whole and partial extracts generating inconsistent outputs.

nicolas-cherel:
re-pushed for a typo fix on `extract-partial-consistency` folder

packages/cli/test/index.test.ts
```diff
@@ -22,13 +23,18 @@ async function prepare(caseFolderName: string) {
...
     recursive: true,
     force: true,
   })
 
-  return { rootDir, actualPath, expectedPath }
+  if (glob.sync(existingPath).length === 1) {
+    await fs.cp(existingPath, actualPath, { recursive: true })
+  }
+
+  return { rootDir, actualPath, existingPath, expectedPath }
```
thekip, line 37: since now its implemented for all test cases, remove

```ts
      await fs.cp(
        nodepath.join(rootDir, "existing"),
        nodepath.join(rootDir, "actual"),
        { recursive: true }
      )
```

from this spec:

```
it("should extract and clean obsolete", 
```

thekip (changes requested):

andrii-bodnar:
@nicolas-cherel could you please address the discussion above?

nicolas-cherel:
sorry, I was not available on july, and forgot to get back to the issue, I'll try to do it ASAP

andrii-bodnar:
Hi @nicolas-cherel, is there anything we can do to move this PR forward?

nicolas-cherel:
So so sorry this got off my radar, I took the quick route with a in github edit, not my proudest moment but it was the best way for me to advance that right now.

andrii-bodnar (approved):

andrii-bodnar:
@nicolas-cherel thank you!

PR #1965, opened 2024-06-24
feat: collect and print placeholders in Po files for better translation context

thekip:
# Description

This feature serves similar purpose as in this PR https://github.com/lingui/js-lingui/pull/1874.

If the message contains unnamed placeholders, such as `{0}` print theirs values into PO comments, so translators got an idea what this placeholder is about. 

```js
t`Hello ${user.name} ${value}`
```

This will be extracted as

Before: 
```po
msgid "Hello {0} {value}"
```

After: 
```po
#. ph: {0} = user.name
msgid "Hello {0} {value}"
```

This benefit of this solution is that developers can reuse more translations, because different placeholder name will not create different translations keys. 

Also now placeholder metadata is available for formatters, so users can implement theirs own logic for placeholders (for example replace them to values)

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/babel-plugin-extract-messages/src/index.ts
+++ b/packages/babel-plugin-extract-messages/src/index.ts
@@ -8,7 +8,6 @@ import {
   isObjectExpression,
 } from "@babel/types"
 import type { PluginObj, PluginPass, NodePath } from "@babel/core"
-import {} from "@babel/core"
 import type { Hub } from "@babel/traverse"
 
 type BabelTypes = typeof BabelTypesNamespace
--- a/packages/babel-plugin-extract-messages/src/index.ts
+++ b/packages/babel-plugin-extract-messages/src/index.ts
@@ -5,8 +5,10 @@ import {
   Node,
   ObjectExpression,
   ObjectProperty,
+  isObjectExpression,
 } from "@babel/types"
 import type { PluginObj, PluginPass, NodePath } from "@babel/core"
+import {} from "@babel/core"
 import type { Hub } from "@babel/traverse"
 
 type BabelTypes = typeof BabelTypesNamespace
@@ -19,6 +21,7 @@ export type ExtractedMessage = {
   origin?: Origin
 
   comment?: string
+  placeholders?: Record<string, string>
 }
 
 export type ExtractPluginOpts = {
@@ -30,6 +33,7 @@ type RawMessage = {
   message?: string
   comment?: string
   context?: string
+  placeholders?: Record<string, string>
 }
 
 export type Origin = [filename: string, line: number, column?: number]
@@ -52,6 +56,7 @@ function collectMessage(
     message: props.message,
     context: props.context,
     comment: props.comment,
+    placeholders: props.placeholders || {},
     origin: [ctx.file.opts.filename, line, column],
   })
 }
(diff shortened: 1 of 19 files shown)


andrii-bodnar:
Nice, I really like this feature, thanks!

What about renaming the `ph:` to the full word `placeholders:`? It's not so clear what the `ph` means. It would be better for translators to understand, as well as for AI (which will have it as context during pre-translation in Crowdin).

Also, can we have a test case for the strings with more than one placeholder?

thekip:
No problem, can do. I'm also thinking about the case where few messages in the codebase with diffrent placeholders. Currently, it will take the only last or first (i didn't test) but probably it makes sense to gather an array of them and print it? 

```js
t`Hello ${user.name}`
t`Hello ${author.name}`
t`Hello ${moderator.name}`
```

```po
#. ph: {0} = user.name | author.name | moderator.name
msgid "Hello {0}"
```

1. how to format it, to make most of it? (so ai/translators would understand it). Note, here could be more complex expression, not only member expression.  Ternaries, function calls, or even self invoking expressions (not sure anyone would do that, but it's possible)
2.  Does it really need to list all of them?  Obviously, the code should dedupe the same entries

Btw the next improvement could be a flag which will completely opt-out from inline placeholder name populated from varaible names in favor of this feature for maximum translation reusability.

thekip:
i don't really like "|" symbol, because the same symbol could be in the placeholder value itself. So it should more clear way to distinguish them. 

```js
t`Hello ${name || surname}`
t`Hello ${userName}`
```

```po
#. ph: {0} = name || surname | userName
msgid "Hello {0}"
```

Also with line breaks with ternary: 

```js
t`Hello ${user
    ? user.name
    : 'User' }`
t`Hello ${userName}`
```
It seems clearing line breaks is unavoidable, because Po format is not allowing multiline comments

```po
#. placeholder: {0} = user ? user.name : 'User' | userName
msgid "Hello {0}"
```


Maybe every entry on a new line? 

```po
#. ph: {0} = name || surname
#. ph: {0} = userName
#. ph: {1} = user.login
#. ph: {1} = user
```

Verbose but for 100% is clear where one placeholder value starts and ends.

thekip:
I'm also thinking that to give an overall context it's enough to show 2 - 3 placeholder values, so we can limit it to avoid bloating the file. From the other side, that could be a tool to quickly spot an issue with a message: 

```js
t`from ${myTrip.date}`
t`from ${myTrip.startCity}`
```

```po
#. placeholder {0}:  myTrip.date
#. placeholder {0}:  myTrip.startCity <-- probably it should be different message created by different context
msgid "from {0}"
```

andrii-bodnar:
Yes, I think limiting the output would also be ok. 

As for spotting the messages that should probably be used in different contexts, that would also be a nice feature. This is a common i18n problem. I think it could be presented in the CLI extract output, or it could be a completely separate CLI command, e.g. `lingui lint`. But this is definitely out of the current scope

thekip:
@andrii-bodnar this one is ready

andrii-bodnar (approved):
Thank you!

PR #1925, opened 2024-04-29
feat: add full support of Vue.js

JSteunou:
add Vue.js support with components, plugins, extractor and compiler

# Description

Title is kinda catchy because in reality there is still a lot to do, but this PR brings at least a full Vue.js experience that *just works*

[//]: # (Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.)

## Types of changes

[//]: # (What types of changes does your code introduce to Lingui?)
[//]: # (_Put an `x` in the boxes that apply_)

- [ ] Bugfix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [X] I have read the [CONTRIBUTING](https://github.com/lingui/js-lingui/blob/main/CONTRIBUTING.md) and [CODE_OF_CONDUCT](https://github.com/lingui/js-lingui/blob/main/CODE_OF_CONDUCT.md) docs
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added the necessary documentation (if appropriate)

To this checklist I will add 

- [x] use jest snapshot
- [x] try again to import vue enum types
- [x] need to re-export some of the core & macro to align with the react package
- [ ] documentation
  - [ ] tutorial
  - [ ] API descriptions
  - [ ] example
- [ ] support of Plural would be a good thing
- [ ] integrate fix of https://github.com/lingui/js-lingui/issues/1627 

---

For more context this is the README of my original repo https://github.com/JSteunou/vue-lingui

## Why?

https://github.com/lingui/js-lingui/discussions/1730

### in the setup part

You can use any of the current [@lingui/macro](https://lingui.dev/ref/macro#js-macros) helper it will *just* work as long as you follow the vite configuration instruction below.

Of course vanilla JS with [@lingui/core](https://lingui.dev/ref/core) also works fine

### in the template part

[...]

PR #1976, opened 2024-07-12
refactor(macro): split and expose internals to be used in Vue macro

timofei-iatsenko:
# Description

extract refactoring to js macro from https://github.com/lingui/js-lingui/pull/1925 

Also add a new macro called "arg" which unblock using native ICU syntax directly with macro such as 

```ts
t`{${arg(count)}, plural, one {}, other {}}`
```

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist


andrii-bodnar:
@thekip Thank you! Could you please describe the new `arg` macro in more detail? What was the motivation and why do we need it?

We should probably decouple exposing internals for Vue integration and the `arg` feature to not block Vue integration. I would like to avoid merging things that are not yet implemented in the SWC plugin since we already have the `useLingui` macro which is not supported in the SWC plugin and it blocks the v5 release.

thekip:
i don't want to split this into two, since this `arg` macro is also a part of Vue integration but could also be helpful in non Vue code as well. 

This macro prints argument placeholder as raw without any additional brackets. This is useful if you want to construct ICU expression manually but still want to use all benefits of macro. 

Say i want to construct a plural icu expression without using a plural macro:  

```js
// without `arg` macro
t`{${count}, plural, one {}, other {}}`
```

We will receive a broken icu because of brackets around a `count` argument: 

```js
i18n.t({
  id: "<hash>",
  message: "{{count}, plural, one {}, other {}}",
  values: {count: count}
})
```

With the `arg` macro brackets would be omitted:

```js
t`{${arg(count)}, plural, one {}, other {}}`

i18n.t({
   id: "<hash>",
  message: "{count, plural, one {}, other {}}",
  values: {count: count}
})
```

This is especially useful if you want to create complex ICU expressions and want to use ICU syntax directly. 

For Vue this is the only one way to implement Plural functionality in templates. There is no counterpart for `<Plural>` instead users have to use `<Trans>` and write ICU expression manually using `arg` macro. 

> SWC plugin since we already have the useLingui macro which is not supported in the SWC plugin and it blocks the v5 release.

I don't think useLingui or arg macro are blocking the release. They are not the core functionality and users could decide not to use them if they use SWC version. We can add a "feature parity" table on the SWC plugin page and note which features are implemented and which are not.

Implementing useLingui hook is extremely difficult in Rust actually, that's why I was against this feature originally considering all the consequences of porting it to the SWC

andrii-bodnar:
Okay, now I get the motivation.

> For Vue this is the only one way to implement Plural functionality in templates

We can accept this in the first release and implement the normal Vue `Plural` later after the release. Let's gather some feedback on the Vue integration first. I think it's the right way instead of creating a new macro that is not fully supported.

Such inconsistencies between Babel and SWC implementations will cause more and more confusion for users. It's better not to release some feature than to release it in half, especially when we have quite a large amount of developers using the SWC plugin. Also, full feature support makes it much easier to write documentation.

thekip:
> We can accept this in the first release and implement the normal Vue Plural later after the release. Let's gather some feedback on the Vue integration first. I think it's the right way instead of creating a new macro that is not fully supported.

There is no way to implement Plural in Vue the same way as in React JSX. So there are no plans to implement a "normal" Plural.

> Such inconsistencies between Babel and SWC implementations will cause more and more confusion for users. It's better not to release some feature than to release it in half, especially when we have quite a large amount of developers using the SWC plugin. Also, full feature support makes it much easier to write documentation.

Would you be interested in contributing to the SWC plugin? Unfortunately, I don't have enough motivation to port features that I don't personally use. Alternatively, you can create a bounty or find a sponsor to support this development effort.

andrii-bodnar:
@JSteunou please share your opinion on this PR and the discussion.

> Would you be interested in contributing to the SWC plugin? Unfortunately, I don't have enough motivation to port features that I don't personally use.

I don't have any expertise in Rust. By the way, you already contributed in #1859 and https://github.com/lingui/swc-plugin/pull/90.

What I'm trying to convey is that such mis-synchronization between Babel and the SWC implementation is not good, and might cause more confusion than the feature brings benefit.

timofei-iatsenko:
@andrii-bodnar This one is updated, please review it and consider merging. This makes working on Vue integration easier and PR would have fewer files.

andrii-bodnar (approved):
@timofei-iatsenko thank you!

PR #1983, opened 2024-07-22
feat(react): preserve newlines during string formatting

Garbanas:
# Description

This change will preserve newline characters also for translations in `@lingui/react` when `<tags>` are present. Newlines are already preserved for values with no tags or just simple placeholders.
This PR ensures that formatting is done the same way for both variants: with and without `<tags>`.

The benefit is that CSS rules or tags that rely on preservation of newlines now work with `<tags>`.
See https://developer.mozilla.org/en-US/docs/Web/CSS/white-space#pre-line

## Types of changes

Depending on your view, this is either a bugfix (ensuring that formatting works the same across variants), or a breaking change (changes the newline behaviour for a formatting variant).

- [ ] Documentation update
- [ ] Examples update

## Checklist

### DIFF
--- a/packages/react/src/format.test.tsx
+++ b/packages/react/src/format.test.tsx
@@ -33,6 +33,20 @@ describe("formatElements", function () {
     ).toEqual('<a href="/about">About</a>')
   })
 
+  it("should preserve newlines", function () {
+    expect(html(formatElements("<0>Inn\ner</0>", { 0: <strong /> }))).toEqual(
+      "<strong>Inn\ner</strong>"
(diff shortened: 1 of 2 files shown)


andrii-bodnar:
Thanks @Garbanas for the contribution!

As @thekip said, this has already been implemented in the `next` v5 release, so I will close this PR.

Feel free to try the [5.0.0-next.0](https://github.com/lingui/js-lingui/releases/tag/v5.0.0-next.0) release and let us know if it works as expected for you.

PR #1984, opened 2024-07-24
fix(extract): escape special regex characters used in file-based routing systems

jjonesrs:
# Description

We recently migrated to [TanStack Router](https://github.com/TanStack/router), which uses file-based routing with several [file naming conventions](https://tanstack.com/router/latest/docs/framework/react/guide/file-based-routing#file-naming-conventions). The `lingui extract --watch` command stopped working on any routes with these special characters. However, `lingui extract` continued to work.

Debugging revealed that `--watch` sets `options.files`, and these paths are formed into a regular expression. If any of these paths contains a character with special regex meaning, like `$`, then `paths` won't be correctly filtered.

```
if (options.files) {
    options.files = options.files.map((p) =>normalize(p, false))
    const regex = new RegExp(options.files.join("|"), "i")
    paths = paths.filter((path: string) => regex.test(path))
}
```

To fix this, I created a new utility that escapes characters allowed in file paths, but with special regex meaning. I'm not sure if I captured all possible characters used in file-based routing conventions, or if I over specified too many.

I added tests for both the utility itself and the extractor's `collect` function. If you remove `makePathRegexSafe` from `collect`, you should see components C, D, and E removed from the snapshot because the extractor fails to handle them.

## Types of changes

- [ ] Documentation update
- [ ] Examples update

## Checklist

### DIFF
--- a/packages/cli/src/api/__snapshots__/catalog.test.ts.snap
+++ b/packages/cli/src/api/__snapshots__/catalog.test.ts.snap
@@ -7,6 +7,44 @@ exports[`Catalog POT Flow Should get translations from template if locale file n
 }
 `;
 
+exports[`Catalog collect should extract files with special characters when passed in options 1`] = `
+{
+  Component C: {
+    comments: [],
+    context: undefined,
+    message: undefined,
+    origin: [
+      [
+        collect/(componentC)/index.js,
+        1,
+      ],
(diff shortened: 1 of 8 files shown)


thekip (approved):
LGTM, thanks for your work!

Issue #1793, opened 2023-10-26
`lingui extract-experimental` extract different file path comment between Windows and Linux


yunsii:
**Describe the bug**

`lingui extract-experimental` should patch [`source-map@0.8.0-beta.0`](https://github.dev/mozilla/source-map/blob/60adcb064bf033702d954d6d3f9bc3635dcb744b/lib/util.js#L201-L203) with:

```diff
function withBase(url, base) {
+  return new URL(url, `file:///${base.replace(/\\/g, '/')}`).toString();
-  return new URL(url, base).toString();
}
```

to make the command works on Windows. And then extract po files with different file path comment:

![image](https://github.com/lingui/js-lingui/assets/18096089/941b4262-3fc7-4420-ad03-addbca2ea3d5)

**To Reproduce**

https://github.com/yunsii/lingui-examples-nextjs-swc

- patch [`source-map@0.8.0-beta.0`](https://github.dev/mozilla/source-map/blob/60adcb064bf033702d954d6d3f9bc3635dcb744b/lib/util.js#L201-L203) manually
- `lingui extract-experimental`

**Expected behavior**

Windows and Linux extract same file path comment.

**Additional context**

https://github.com/lingui/js-lingui/blob/04b7cef9876168bc376b55424ecc0b0ebde976c1/packages/cli/src/lingui-extract-experimental.ts#L57-L62

Windows esbuild bundleResult:
![image](https://github.com/lingui/js-lingui/assets/18096089/69f086a0-a5cf-4a8d-b3e2-0e6b6713e57d)

Linux esbuild bundleResult:
![image](https://github.com/lingui/js-lingui/assets/18096089/b38e5324-5aa2-4a45-a476-ad60bf4ba66c)

- jsLingui version `4.5.0`
- Macro support:
 - [x] I'm using SWC with `@lingui/swc-plugin`
 - [ ] I'm using Babel with `babel-macro-plugin`
 - [ ] I'm not using macro

thekip:
Try to use WSL, there no sense to continue developing under native windows. I'm not personally interested in investing time in that.

yunsii:
> Try to use WSL, there no sense to continue developing under native windows. I'm not personally interested in investing time in that.

Yes, I have used WSL many years, this issue was found on native Windows of my colleagues. Is it inappropriate not to consider cross-system issues?

thekip:
If even microsoft gave up supporting native windows for developers and brought a better alternative, why we should invest in that? 😅 

[...]

PR #1996, opened 2024-08-09
fix: windows compatibility

thekip:
# Description

i installed Windows ARM on my Mac, and fixed all compatibility issues which were marked as "skipOnWindows" in tests.

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/cli/src/api/catalog/getCatalogDependentFiles.test.ts
+++ b/packages/cli/src/api/catalog/getCatalogDependentFiles.test.ts
@@ -5,7 +5,7 @@ import { FormatterWrapper } from "../formats"
 import mockFs from "mock-fs"
 import * as process from "process"
 import os from "os"
-const skipOnWindows = os.platform() === "win32" ? it.skip : it
+import path from "node:path"
 
 describe("getCatalogDependentFiles", () => {
   let format: FormatterWrapper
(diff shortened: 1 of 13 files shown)


yunsii:
Is it will fix this issue https://github.com/lingui/js-lingui/issues/1793?

thekip:
no

thekip:
I hope so. I found that path issues on windows become even more weird when inter-disc execution is involved. For example, code working perfectly when it stored on disk C, but then suddenly become broken when you just move it to the disk D. 

There is also no way to express a relative path between these two disks, and not every tool understand an absolute windows paths.

yunsii:
> but then suddenly become broken when you just move it to the disk D.

I always think the comment is relative path in git repo, why related the inter-disc execution? 😦

thekip:
Exactly, it should be relative to the rootDir specified in the lingui config. And that the goal i wanted to achieve, so no matter where files are stored result should be the same. But on windows, it turned out that it might be broken due to where files are stored, which is weird. 

The inter disc execution happened in experimental extractor. Previously, it used `os.tmpDir()` to store esbuild generated bundles. On windows it's always would be on the system disk (C:) and if you code stored on disk D: esbuild not able to generate a correct relative path to sources, and then lingui extractor could not make this path relative to rootDir.

yunsii:
Is it possible to make lingui extractor always use same disk (maybe C:) to parse relative path?

thekip:
That exactly what i did in this PR, among other things

andrii-bodnar (approved):
@thekip thank you!

bravo-kernel:
This is great, much appreciated 👏

PR #2006, opened 2024-08-19
feat: allow to use useLingui with only i18n

thekip:
# Description

related: https://github.com/lingui/swc-plugin/issues/106#issuecomment-2295951277

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist


AndrewIngram:
Does this also allow use of the hook without any assignment?

eg
```js
function MyComponent() {
  useLingui();
  return <p>Something</p>
}
```

We don't make much use of this, but it's a useful trick to subscribe a component to locale changes even if the component body itself doesn't have any translated strings -- but it might call utility function that do.

andrii-bodnar (approved):

PR #2008, opened 2024-08-19
feat: adds option for CLDR plurals in gettext

garikkh:
# Description

Gettext has its own plural rules that are many times different than CLDR plurals. See [gettext plurals](https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html) and [CLDR plurals](https://www.unicode.org/cldr/charts/44/supplemental/language_plural_rules.html)

This option is helpful if you need to use CLDR plurals, but also have gettext style PO files. 

In my case, my TMS does not support ICU in PO, but also uses CLDR plurals and not gettext plurals. 

## Types of changes

- [X] Documentation update
- [ ] Examples update

## Checklist

- [X] I have added the necessary documentation (if appropriate)


garikkh:
Example issue:

Czech in gettext has 3 plural forms, but CLDR defines 4. If you have a translation entry like this one below, the default gettext formatter will fail. 

```po
msgid ""
msgstr ""
"Language: cs\n"
#. js-lingui:icu=%7Bcount%2C+plural%2C+one+%7B%7Bcount%7D+day%7D+few+%7B%7Bcount%7D+days%7D+many+%7B%7Bcount%7D+days%7D+other+%7B%7Bcount%7D+days%7D%7D&pluralize_on=#
msgid "# day"
msgid_plural "# days"
msgstr[0] "# den"
msgstr[1] "# dny"
msgstr[2] "# dne"
msgstr[3] "# dní"
```

And it will be incorrectly parsed into:

`{#, plural, one {# den} few {# dny} other {# dne} undefined {# dní}}`

But with this new option set, it will correctly create:
`{#, plural, one {# den} few {# dny} many {# dne} other {# dní}}`

PR #2013, opened 2024-08-24
Extra details about i18next + better lingui examples

igorsantos07:
# Description

As a user of i18next for around 3 years, I noticed some parts of the comparison were lacking. I got specially baffled by the verbosity of interpolations, but I'm glad I checked the API docs before giving up :)

> BTW, do [`i18n.t()`](https://lingui.dev/ref/core#i18n.t) really need different examples from `i18n._()`? I would say that only causes more confusion; at first glance it looks like that it works differently, even though it's an alias.

An important fact to distinguish as well is that you need a lot of extra config in i18n to use natLang keys, which honestly is quite troublesome for people coming from gettext. This happens because i18next relies solely on JSON files, so they need some... _workarounds_... to shove different details in the same JSON key.

I also replaced the `t()` references with `_()`, as that's the main function (and not the alias).

Lastly, I'm not entirely sure about the syntax to include translation context in Lingui. I stumbled upon `comment` in some other example and in the API docs, but again, that's not exactly clear. I would advise add a description for all options that can be used in `_()`, as well as examples for context/comments.

## Types of changes

- [x] Documentation update
- [x] Examples update

## Checklist

### DIFF
--- a/website/docs/misc/i18next.md
+++ b/website/docs/misc/i18next.md
@@ -28,7 +28,7 @@ i18next.init({
 document.getElementById("output").innerHTML = i18next.t("key");
 ```
 
-Since the Lingui v4 release, there is a core function [i18n.t(...)](/docs/ref/core.md#i18n.t) that allows doing pretty much the same thing. The following example shows how this works with Lingui:
+Since the Lingui v4 release, there is a core function [i18n._(...)](/docs/ref/core.md#i18n._) that allows pretty much the same thing. There's also a more descriptive syntax, if desired. The following example shows how this works with Lingui:
 
 ```js title="lingui.config.{js,ts}"
 /** @type {import('@lingui/conf').LinguiConfig} */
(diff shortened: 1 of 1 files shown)


igorsantos07:
By the way, [this](https://lingui.dev/tutorials/react-patterns#lazy-translations) is something that i18next does NOT support, and caused a lot of subtle bugs on our application until I realized what was going on. I'm not sure how to better write about this in the comparison page, though :|

website/docs/misc/i18next.md
```diff
@@ -192,7 +200,19 @@ i18next.t("key", { count: 5 }); // -> "items"
...
+
+```js
+i18next.t("Hello!", { context: "This goes in the header" });
+```
+
+Lingui sample:
+
+```js
+i18n._("Hello!", { comment: "This goes in the header" });
+```
```
thekip, line 215: Unfortunately that will not work. Most lingui features build around macros, pieces of code which executed during build step not a runtime. This helps reduce size footprint of the translation catalogs as well as library itself.

In this case to specify a context for a message you MUST use a macro like this

```js
import {t} from "@lingui/macro";

t({message: `From: `, context: "direction"})
```

With plain runtime modules (i18n._) the only way is custom id.

andrii-bodnar:
@igorsantos07 @timofei-iatsenko is there anything we can do to help move this PR forward?

timofei-iatsenko:
This article should be proofread and lingui examples replaced to theirs `macro` alternatives.

andrii-bodnar:
Close in favor of #2064

Issue #2012, opened 2024-08-22
Allow missing translations to be compiled/interpolated on production


garikkh:
**Is your feature request related to a problem? Please describe.**

Currently, translations that are not extracted have issues when rendering on production. This only happens when `NODE_ENV=production`. There are two main issues:

1. the `message` attribute is removed on production. This seems to be intentional, and can be fixed by configuring the babel macro plugin, but it is really not clear and maybe just needs to be documented better
2. Values that are interpolated will NOT get rendered properly - the interpolation does not happen because of a `NODE_ENV` check prevents the message to be compiled properly.

[related check](https://github.com/lingui/js-lingui/blob/main/packages/core/src/i18n.ts#L242-L246)

With this, you are _forced_ to extract and compile when doing any production build in your JS package. Although many people probably do, I think its an unnecessary restriction on the library.

---

**Describe proposed solution**

A configuration option (probably in the actual `i18n` object) that allows for compiling the translation on production.

---

**Describe alternatives you've considered**

I know that `@lingui/macro` has an option called `extract` that keeps the default message as a parameter when creating the `i18n(...)` call, however this isn't enough because it only works with basic strings. Strings that need to be interpolated will not be correctly interpolated, and placeholder variables will show in production.

Without macro option:
![image](https://github.com/user-attachments/assets/9d8364ea-65fc-4f02-91f1-459e63e07289)

With the macro option:

![image](https://github.com/user-attachments/assets/93881bd9-e47b-4f8c-b115-a34f07bb56d8)

---

**Additional context**

garikkh:
Added option here: https://github.com/lingui/js-lingui/compare/main...garikkh:js-lingui:feat/prod-compile-opt

thekip:
Yes, this intentional by design. 

> Added option here: https://github.com/lingui/js-lingui/compare/main...garikkh:js-lingui:feat/prod-compile-opt

This will break tree shaking for message compiler. 

I believe making it configurable brings more maintenance burden and confusing for people than benefits. 

Step back and let's analyze WHY you need this. I bet it's because you are struggling with proper place for extraction command, so sometimes something don't get to catalogs. 

Try to put an additional extraction command just before build command:

```
build: "lingui extract && app build"
```

This would work as a last resort if something was not extracted / translated and you will not see ids in the bundle

garikkh:
> This will break tree shaking for message compiler.

Not sure what you mean by this, this is the runtime, why would it affect tree shaking.

[...]

PR #1913, opened 2024-04-10
fix: escaped curly-braces doesn't preserve in development mode 

ninesunsabiu:
# Description

#1901 

[//]: # (Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.)

## Types of changes

[//]: # (What types of changes does your code introduce to Lingui?)
[//]: # (_Put an `x` in the boxes that apply_)

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [x] I have read the [CONTRIBUTING](https://github.com/lingui/js-lingui/blob/main/CONTRIBUTING.md) and [CODE_OF_CONDUCT](https://github.com/lingui/js-lingui/blob/main/CODE_OF_CONDUCT.md) docs
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added the necessary documentation (if appropriate)

PR #2035, opened 2024-09-21
feat: add setMessagesCompiler method

timofei-iatsenko:
# Description

1. Allow users to use uncompiled catalogs on production (json)
2. have a proper working fallback to default message defined in the code (with the proper macro config)

Related user request: https://github.com/lingui/js-lingui/issues/2012

To achieve these goals, `i18n.setMessagesCompiler()` method was provided, which allow to manually register message compiler on production. 

Also, the format of compiled message was slightly changed. 

Previously, if the message doesn't have any ICU features inside it was printed as a string, array used for the rest of cases. That make distinguishing between compiled and uncompiled message particularly hard and sometimes lead to double compilation which was tried to be fixed here https://github.com/lingui/js-lingui/pull/1913

Now uncompiled message is always a string, and compiled is always an array. 

Some examples: 
```
"Hey '{name}'!" -> ["Hey {name}!"] // escaped placeholder
"Hey {name}!" -> ["Hey", "name", "!"] // icu placeholder
"Hey!" -> ["Hey!"] 
```

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist


timofei-iatsenko:
Had to exclude test from `@lingui/remote-loader`, because it's not published since version 3 and has outdated dependencies. That causes integration tests to fail, because snapshots are created using source code and new compiler format, but integration tests use packages from `node_modules` and for `@lingui/remote-loader` it takes stale versions.

timofei-iatsenko:
Something happened to Verdaccio. The WF uses:

`npm i -g {verdaccio,verdaccio-auth-memory,verdaccio-memory}`

Which is installing latest possible version, they published version 6 couple days ago and that might be the reason. I tried to fix v5 version, but it didn't help. The error:

> info attempt registry request try #1 at 10:04:12 AM
> http request PUT http://0.0.0.0:4873/-/user/org.couchdb.user:test
> info retry will retry, error on last attempt: Error: connect ECONNREFUSED 0.0.0.0:4873

However, the verdaccio itself is up and running:

image

.github/workflows/release-test.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       - name: Publish with Verdaccio
         run: |
           git switch -c "pull-request"
-          npm i -g {verdaccio,verdaccio-auth-memory,verdaccio-memory}
+          npm i -g {verdaccio@5,verdaccio-auth-memory,verdaccio-memory}
```
andrii-bodnar, line 34: This is no longer needed. The Verdaccio issue is already fixed

packages/core/src/i18n.ts
```diff
@@ -125,6 +133,26 @@ export class I18n extends EventEmitter<Events> {
...
+   *
+   * Example usage:
+   *
+   * ```ts
+   * import { compileMessage } from "@lingui/message-utils/compileMessage";
+   *
+   * i18n.setMessagesCompiler(compileMessage);
+   * ```
+   */
+  setMessagesCompiler(compiler: MessageCompiler) {
```
andrii-bodnar, line 151: Please add the new API to the documentation - https://lingui.dev/ref/core#i18n

vonovak (approved):
Hi, apologies for late reply. I have the same feedback as Andrii :)

vonovak (changes requested):
Hi, apologies for late reply. I have the same feedback as Andrii - docs :)

timofei-iatsenko:
@andrii-bodnar @vonovak i added docs for that method. The better overall tutorial will be in next iterations after this [epic](https://github.com/lingui/js-lingui/issues/2043) finished

website/docs/ref/core.md
```diff
@@ -115,16 +115,32 @@ Formatting of messages as strings (e.g: `"My name is {name}"`) works in developm
...
 
 :::
 
+### `i18n.setMessagesCompiler(compiler)` {#i18n.setMessagesCompiler}
+
+Registers a `MessageCompiler` to enable the use of uncompiled catalogs at runtime.
+
+In production builds, the `MessageCompiler` is typically excluded to reduce bundle size.
+
+By default, message catalogs should be precompiled during the build process. However, if you need to compile catalogs at runtime, you can use this method to set a message compiler.
```
vonovak, line 134: As a user reading the docs, I understand what the function does (the `setMessagesCompiler` is also quite self-explanatory) but I don't really understand why I'd need such feature in the first place. Could we have a paragraph about what problem it solves and how to use it? 

I know writing docs is annoying but I think it'd be great to not wait until a later epic is finished.
The PR mentions "the proper macro config" - it'd be great to know what that is, otherwise that can be quite frustrating for consumers since @timofei-iatsenko is probably the only one who knows that "proper config". 🙈 
Apologies and thanks :)

vonovak (changes requested):

website/docs/ref/core.md
timofei-iatsenko: There is still functionality needed to be done before this article could be written. I don't want to bring too much user attention to that while it would be 100% ready.  That was also the reason why I didn't write this doc in the first place. 

This is a breaking change, due to message catalog format changes, and i wanted this to be included in next major release. The rest features needed to make this work would not be breaking and could be done later.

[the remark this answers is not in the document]
timofei-iatsenko: I don't like the proposed change. It's bringing a confusion for the users, only `lingui compile`? What if i use Vite plugin, should I still use `lingui compile` ? 

I think we need a separate page describing the concept of compiling catalogs. With all question and answers for different options and link to that page might be added to this doc.

website/docs/ref/core.md
```diff
@@ -115,16 +115,32 @@ Formatting of messages as strings (e.g: `"My name is {name}"`) works in developm
...
+
+Example usage:
+
+```ts
+import { compileMessage } from "@lingui/message-utils/compileMessage";
```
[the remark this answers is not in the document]
timofei-iatsenko: It's coming from `@lingui/` scope, isn't it obvious enough? More over, this method wasn't built to make it possible to override a default implementation, so it's intended to use always with a default implementation.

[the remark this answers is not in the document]
vonovak: > this method wasn't built to make it possible to override a default implementation, so it's intended to use always with a default implementation.

That's a bit confusing to me because then I don't understand the use of this so much. But it's probably because other changes need to be done before this can be properly documented, as you said before.

> isn't it obvious enough?

Well, just a few moments ago it felt obvious to me that I can use also something else other than the default implementation, so the word "obvious" can be a bit tricky.

If no other implementation should be used, the docs should say that imho.

Anyways, I don't want to block this PR from merging. 🙂👍

vonovak (approved):

andrii-bodnar (approved):

PR #2064, opened 2024-10-23
docs: Resources section improvements

andrii-bodnar:
# Description

🖼️ [Deployment Preview](https://js-lingui-git-docs-resources-crowdin.vercel.app/)

## Types of changes

- [x] Documentation update
- [ ] Examples update

## Checklist

### DIFF
--- a/website/docs/misc/i18next.md
+++ b/website/docs/misc/i18next.md
@@ -5,11 +5,13 @@ description: Comparison of Lingui and i18next internationalization libraries
 
 # Comparison with i18next
 
-[i18next](https://www.i18next.com/) is an internationalization-framework written in and for JavaScript. i18next and Lingui are two popular internationalization (i18n) libraries used for translating and localizing JS-based applications. Both libraries have their strengths and weaknesses, and which one is best for a particular project depends on the project's specific needs.
+[i18next](https://www.i18next.com/) is a widely used internationalization framework designed specifically for JavaScript applications. Both i18next and Lingui are popular libraries for translating and localizing JavaScript-based projects, each offering unique strengths and features.
 
-## Basic comparison
+The choice between them ultimately depends on the specific needs of your project.
 
-Here is a basic example of the i18next usage:
+## Basic Comparison
+
+Here's a simple example of how to use i18next:
 
 ```js
 import i18next from "i18next";
@@ -24,11 +26,15 @@ i18next.init({
     },
   },
 });
-// ...
+```
+
+```js
+import i18next from "i18next";
+
 document.getElementById("output").innerHTML = i18next.t("key");
 ```
 
-Since the Lingui v4 release, there is a core function [i18n.t(...)](/docs/ref/core.md#i18n.t) that allows doing pretty much the same thing. The following example shows how this works with Lingui:
+The equivalent example with Lingui looks like this:
 
 ```js title="lingui.config.{js,ts}"
 /** @type {import('@lingui/conf').LinguiConfig} */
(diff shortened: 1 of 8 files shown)


website/docs/misc/i18next.md
```diff
@@ -45,48 +51,72 @@ module.exports = {
...
 
+If you prefer to use explicit IDs for your messages, you can utilize the `msg` macro:
+
+```js
+import { msg } from "@lingui/core/macro";
+
+msg({ id: "msg.greeting", message: `Hello World` });
+```
+
+Read more about [Explicit vs Generated Message IDs](/docs/guides/explicit-vs-generated-ids.md).
```
timofei-iatsenko, line 71: It also possible tow write explicit id with the `t` macro, 

```js
import { t } from "@lingui/core/macro";

document.getElementById("output").innerHTML = t({ id: "msg.greeting", message: `Hello World` });
```

The difference between `msg` and `t` is first one returns `MessageDescriptor`, second one wraps this message descriptor into global `i18n._` method, so returns a translated string directly. The signature of these to macros is almost the same.

website/docs/misc/i18next.md
```diff
@@ -45,48 +51,72 @@ module.exports = {
...
 Lingui sample:
 
 ```js
-i18n._({ id: "msg.name", message: "My name is {name}", values: { name: "Tom" } });
+import { msg, t } from "@lingui/core/macro";
+
+const name = "Tom";
+
+t`My name is ${name}`;
+msg({ id: "msg.name", message: `My name is ${name}` });
```
timofei-iatsenko, line 94: As commented earlier, more correct analog would be: 

```js
t`My name is ${name}`;
t({ id: "msg.name", message: `My name is ${name}` });
```

website/docs/misc/i18next.md
```diff
@@ -185,37 +218,78 @@ i18next handles plurals differently. It requires a separate key to be defined fo
...
+Lingui sample:
+
+```js
+import { msg } from "@lingui/core/macro";
+
+msg({
+  message: "Right",
+  context: "direction",
+});
+```
```
timofei-iatsenko, line 249: Would it be usefull to add a `po` example here to show what was written in the description above: 

> Lingui provides an automatic additional "context" by including in the `.po` file the locations where each message is used - i18next can't do this from its plain JSON files.

website/docs/misc/i18next.md
```diff
@@ -185,37 +218,78 @@ i18next handles plurals differently. It requires a separate key to be defined fo
...
+Lingui sample:
+
+```jsx
+import { Trans } from "@lingui/react/macro";
+
+const HelloWorld = () => {
+  return <Trans>Hello World!</Trans>;
+};
+```
+
```
timofei-iatsenko, line 274: I think the main diffrence between this two, and key selling point of using macro over runtime jsx elements is ability to use child components in the string and don't split message into chunks. I think it's worth to highlight it here.

```jsx
import { Trans } from "react-i18next";

const HelloWorld = () => {
  return <Trans i18nKey="link.check">Check</Trans>{' '}<a href="#"><Trans i18nKey="link.this">this!</Trans</a>;
};
```

Lingui sample:

```jsx
import { Trans } from "@lingui/react/macro";

const HelloWorld = () => {
  return <Trans>Check <a href="#">this!</a></Trans>;
};
```

timofei-iatsenko:
Reviewed the i18next page, will come back to the rest a bit later.

website/docs/misc/i18next.md
andrii-bodnar: Based on the documentation (https://react.i18next.com/latest/trans-component), it doesn't seem like it should be concatenated that way like in your example

timofei-iatsenko: Wow, you are right, I've believed that this is not possible with a plain JSX. Hence created this example.

Now I'm extremely interested how do they do that.

PR #2153, opened 2025-01-15
chore: remove integration tests

timofei-iatsenko:
# Description

I tried to fix "integration" tests and then decided to completely remove them. "integration"  taken in quotes, because it's actually not an integration tests. It is the same test but executed on the compiled output. This doesn't have much sense now and bring more maintenance burden for us. 

The unbuild is responsible to validate correctness of entrypoints in the `package.json`. And the way how we consume packages in tests is as close as possible to real use case with compiled packages, so we are more than on the safe side in this regard. 

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist


timofei-iatsenko:
Unblocking pipeline from integration tests revealed a problem with typings.  I did a fixin this PR, but wondered why it wasn't caught during `tsd` tests. I dug into tsd setup and found that our setup is obsolete. tsd-lite we use is deprecated and archived. The proposed alternative is https://tstyche.org/

Also with a newer version of tsd / tsd-lite and stricter checks it started complaining about ramda types, so i decided to [put an effort](https://github.com/lingui/js-lingui/pull/2155) to get rid of it.

I will do migration to tstyche in next PRs

andrii-bodnar (approved):

PR #2209, opened 2025-03-26
docs(examples): add Tanstack Start example

depsimon:
# Description

This PR adds a Tanstack Start example that works with Vite & the Babel plugin.

The locale is stored in a `locale` cookie but can be defined by an URL Search param `?locale=fr`.


examples/tanstack-start/app.config.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export default defineConfig({
+  tsr: {
+    appDirectory: 'src',
+  },
+  react: {
+		babel: {
+			plugins: ["@lingui/babel-plugin-lingui-macro"],
+		},
+	},
```
timofei-iatsenko, line 13: weird formatting here and in other places, did you apply prettier to these files?

depsimon: prettier is ignoring examples. I forced it to run on this example now

PR #2218, opened 2025-04-04
feat(macro): make select macro choices strictly typed

### LINKED ISSUE #2214: Stricter types for select macro
### Problem Description

Select macro doesn't validate choices other than `other` currently.

### Proposed Solution

Make the macro generic over value, and the choice keys required if the value is a union type.

Potential solution:
```tsx
 declare function select<T extends string | number>(
    value: T,
    options: {
      [key in T]: string
    } & {
      other: string
    }
  ): string
```

### Alternatives Considered

I can extend the types or patch the dependency locally, but it could make sense to contribute this change to benefit everyone

### Additional Context

_No response_

MikaelSiidorow:
# Description

Closes #2214, implementing stricter types for the options in the select macro.

## Types of changes

  - Could potentially be breaking change if you have runtime errors in your code at the moment, and are running strict typechecking? 
- [x] Documentation update
- [ ] Examples update

Fixes #2214 

## Checklist

### DIFF
--- a/packages/babel-plugin-lingui-macro/src/macroJsAst.ts
+++ b/packages/babel-plugin-lingui-macro/src/macroJsAst.ts
@@ -180,6 +180,8 @@ export function tokenizeChoiceComponent(
     format: format,
     options: {
       offset: undefined,
+      /** Default to fill other with empty value for compatibility with ICU spec */
+      other: "",
     },
   }
 
--- a/packages/core/macro/index.d.ts
+++ b/packages/core/macro/index.d.ts
@@ -154,11 +154,20 @@ export function selectOrdinal(
   options: ChoiceOptions
 ): string
 
-type SelectOptions = {
+type SelectOptionsExhaustive<T extends string = string> = {
+  [key in T]: string
+}
+
+type SelectOptionsNonExhaustive<T extends string = string> = {
   /** Catch-all option */
   other: string
-  [matches: string]: string
+} & {
+  [key in T]?: string
 }
+
+type SelectOptions<T extends string = string> =
+  | SelectOptionsExhaustive<T>
+  | SelectOptionsNonExhaustive<T>
 
 /**
  * Selects a translation based on a value
@@ -180,9 +189,9 @@ type SelectOptions = {
  * @param value The key of choices to use
  * @param choices
  */
-export function select(
-  value: string | LabeledExpression<string>,
-  choices: SelectOptions
+export function select<T extends string = string>(
+  value: T | LabeledExpression<T>,
+  choices: SelectOptions<T>
 ): string
 
 /**
--- a/packages/react/macro/__typetests__/index.test-d.tsx
+++ b/packages/react/macro/__typetests__/index.test-d.tsx
@@ -11,7 +11,8 @@ import {
 import React from "react"
 import { ph } from "@lingui/core/macro"
 
-const gender = "male"
+type Gender = "male" | "female"
+const gender = "male" as Gender
 const user = {
   name: "John",
 }
@@ -126,8 +127,20 @@ m = (
 // @ts-expect-error: `value` could be string only
 m = <Select value={5} other={"string"} />
 
-// @ts-expect-error: `other` required
-m = <Select value={"male"} />
+// @ts-expect-error: `other` required unless exhaustive
+m = <Select value={gender} />
+
+// @ts-expect-error: `other` required unless exhaustive
+m = <Select value={gender} _male="..." />
+
+// @ts-expect-error: `other` required unless exhaustive
+m = <Select value={gender} _female="..." />
+
+// non-exhaustive okay if other is defined
+m = <Select value={gender} _female="..." other="..." />
+
+// exhaustive okay without other
+m = <Select value={gender} _male="..." _female="..." />
 
 // @ts-expect-error: `value` required
 m = <Select other={"male"} />
--- a/packages/babel-plugin-lingui-macro/src/macroJsAst.test.ts
+++ b/packages/babel-plugin-lingui-macro/src/macroJsAst.test.ts
@@ -372,5 +372,34 @@ describe("js macro", () => {
         },
       })
     })
+
+    it("select without other", () => {
+      const exp = parseExpression(
+        `select(gender, {
+          male: "he",
+          female: "she",
+        })`
+      )
+      const tokens = tokenizeChoiceComponent(
+        (exp as NodePath<CallExpression>).node,
+        JsMacroName.select,
+        createMacroCtx()
+      )
+      expect(tokens).toMatchObject({
+        format: "select",
+        name: "gender",
+        options: expect.objectContaining({
+          female: "she",
+          male: "he",
+          offset: undefined,
+          other: "", // <- other should be filled with empty string
+        }),
+        type: "arg",
+        value: {
+          name: "gender",
+          type: "Identifier",
+        },
+      })
+    })
   })
 })
--- a/packages/babel-plugin-lingui-macro/src/macroJsx.test.ts
+++ b/packages/babel-plugin-lingui-macro/src/macroJsx.test.ts
@@ -322,5 +322,33 @@ describe("jsx macro", () => {
         },
       })
     })
+
+    it("Select without other", () => {
+      const macro = createMacro()
+      const exp = parseExpression(
+        `<Select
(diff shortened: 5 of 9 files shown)


timofei-iatsenko:
Thanks! Basically `other` is a required case from icu spec, it's a catch all option which should always be presented. I would like to see the test case where, input type does not specify the `other` in itself,  because this is most likely would be 99% of cases: 

```ts
type Gender = "male" | "female";

const genger: Gender = 'male'

expectType<string>(
  select(gender, {
    male: "he",
    other: "female",
  })
)
```

The cases for the select is not exhaustive, means you can skip one, and it should fallback to the `other`

MikaelSiidorow:
> Thanks! Basically other is a required case from icu spec, it's a catch all option which should always be presented. I would like to see the test case where input type does not specify the other in itself, because this is most likely would be 99% of cases:

I will add another test case, which doesn't contain other in the set of options!

> The cases for the select are not exhaustive, means you can skip one, and it should fallback to the `other`

If that's the case, should this instead be a new optional macro or option to enable exhaustiveness? I find the use case where it is exhaustive more common in our real world usage scenario than not being exhaustive

timofei-iatsenko:
> If that's the case, should this instead be a new optional macro or option to enable exhaustiveness? I find the use case where it is exhaustive more common in our real world usage scenario than not being exhaustive

I'm not sure about this one, all macro is just a syntax sugar about ICU expressions and should follow theirs semantics, adding a new macro with a different behavior contradict with this architecture rule. 

> The pattern string for SelectFormat allows user-defined keywords that match the pattern [a-zA-Z][a-zA-Z0-9_-]* as well as the predefined keyword "other". The keyword "other" must occur in the pattern string, or the error U_DEFAULT_KEYWORD_MISSING is returned. For comparison, the PluralFormat pattern allows as keywords only the set of tags defined in UTR 35, Unicode Locale Data Markup, section C. 11, Language Plural Rules.

https://icu.unicode.org/design/formatting/select

If you create a exhaustive rule that means the user will have to specify one of the options twice: 

```ts
type Gender = "male" | "female";

const genger: Gender = 'male'

expectType<string>(
  selectExhaustive(gender, {
    male: "he",
    female: "female",
    other: "female", // <-- still should be there even in the exhaustive version
  })
)
```

I understand your intention to make it 100% type safe and help to catch bugs when a new option added to the input type, but the underlying ICU Select was designed not considering this case.

Kinda from the DX perspective, I think ergonomically would be nice to implement exhaustive / non-exhaustive based if the `other` case presented or not. Similarly, how pattern matching works in languages as Rust

```ts
type SelectOptionsExhaustive<T extends string = string> = {
  [key in T]: string
}

type SelectOptionsNonExhaustive<T extends string = string> =  {
  /** Catch-all option */
  other: string
} & {
  [key in T]?: string
}

type SelectOptions<T extends string = string> = SelectOptionsExhaustive<T> | SelectOptionsNonExhaustive<T>
```

But again this will contradict with the underlying ICU Select.

timofei-iatsenko:
First of all, thanks for digging into the topic. Second, let's check how the message format parser would behave if you will not specify `other` case in the message. I'm not particularly fan of the diverging from the spec, but i see a value in that. If the message format parser will not complain about this one we could simply make a version where other case is not specified at all. Or simply left `other {}` empty for the exhaustive version.

MikaelSiidorow:
> First of all, thanks for digging into the topic. Second, let's check how the message format parser would behave if you will not specify `other` case in the message. I'm not particularly fan of the diverging from the spec, but i see a value in that. If the message format parser will not complain about this one we could simply make a version where other case is not specified at all. Or simply left `other {}` empty for the exhaustive version.

From quick testing ignoring the `other` value (leaving completely undefined) didn't seem to break anything at least when used with React through the Vite plugin. Should that be the approach then, adding conditional types to `select` and `<Select>` macro:
1. If `other` is defined, allow for loose typing and missing keys
2. If other is not defined strictly validate keys

timofei-iatsenko:
@MikaelSiidorow 

image

Even if the parser used in the lingui is not crashing, most likely that wouldn't be a case in TMS. Example above from the message format online editor. So omitting `other ` could result in breaking compatibility with different vendors. 

It is better to left it empty. Also, i suppose this change might be a breaking for the users, since typing become stricter.  But we can ignore it, possibly

MikaelSiidorow:
> It is better to left it empty. 

Makes sense. I guess this also needs runtime code changes, not only types. I can check this later

> Also, i suppose this change might be a breaking for the users, since typing become stricter. But we can ignore it, possibly

Yep, noted in the description. I guess it needs to be decided whether to treat typechecking errors as breaking or not

packages/babel-plugin-lingui-macro/src/macroJsAst.ts
```diff
@@ -180,6 +180,8 @@ export function tokenizeChoiceComponent(
     format: format,
     options: {
       offset: undefined,
+      /** Default to fill other with empty value for compatibility with ICU spec */
+      other: "",
```
timofei-iatsenko, line 184: oh, that means SWC plugins changes would also be required

timofei-iatsenko:
Looks good to me. We also need to update docs for these macros, just with few sentences about this new behaviors

packages/babel-plugin-lingui-macro/src/macroJsAst.ts
MikaelSiidorow: I see, I guess that's in a different repo 👀

timofei-iatsenko (approved):

timofei-iatsenko (changes requested):
Since there is a changes done to the macro, you need to add a test for the macro itself covering this change

MikaelSiidorow:
> Since there is a changes done to the macro, you need to add a test for the macro itself covering this change

Makes sense! Thanks for being super responsive. I have been a bit busy, but I'll update this PR later

MikaelSiidorow:
@timofei-iatsenko 
> Since there is a changes done to the macro, you need to add a test for the macro itself covering this change

Finally got back to this, sorry for the delay! I have added a test for the new macro functionality

vcarel:
Hello, not pushing, but I would be super interested by this PR. Any chance it could be merged soon? 🙏😊

packages/react/macro/__typetests__/index.test-d.tsx
```diff
@@ -11,7 +11,8 @@ import {
 import React from "react"
 import { ph } from "@lingui/core/macro"
 
-const gender = "male"
+type Gender = "male" | "female"
+const gender = "male" as Gender
```
vonovak:
```suggestion
const gender = "male" satisfies Gender
```

this is makes sure TS treats this as `male` but ensures it's a `Gender` at the same time

packages/core/macro/index.d.ts
```diff
@@ -154,12 +154,21 @@ export function selectOrdinal(
...
+  [key in T]: string
+}
+
+type SelectOptionsNonExhaustive<T extends string = string> = {
   /** Catch-all option */
   other: string
-  [matches: string]: string
+} & {
+  [key in T]?: string
 }
```
vonovak, line 166: can reuse the `SelectChoiceOptionsExhaustive` with

`& Partial<SelectChoiceOptionsExhaustive<T>>`

but no strong feelings about this being better

packages/react/macro/__typetests__/index.test-d.tsx
```diff
@@ -126,8 +127,20 @@ m = (
...
+m = <Select value={gender} _male="..." />
+
+// @ts-expect-error: `other` required unless exhaustive
+m = <Select value={gender} _female="..." />
+
+// non-exhaustive okay if other is defined
+m = <Select value={gender} _female="..." other="..." />
+
+// exhaustive okay without other
+m = <Select value={gender} _male="..." _female="..." />
```
vonovak, line 144:
```suggestion
// @ts-expect-error: Property _female does not exist on type...
m = <Select value={"male"} _male="..." _female="..." />

// `_female` can be specified for "male" if generic type is provided
m = <Select<Gender> value={"male"} _male="..." _female="..." />

```

just adding two type tests

vonovak (changes requested):
just some super minor stuff, thanks for the PR!

jtbandes:
I'm interested in this too :)

PR #2272, opened 2025-06-25
docs: Add note about react compiler

gustavoharff:
# Description

While setting up Lingui in my Vite project, I noticed that the macros were unstable — some worked, others didn't. After several tests, I realized that my `macros` plugin was listed after the `babel-plugin-react-compiler`. Once I swapped the order of the plugins, everything worked perfectly.

## Types of changes

- [x] Documentation update

## Checklist

- [x] I have added the necessary documentation

### DIFF
--- a/website/docs/installation.mdx
+++ b/website/docs/installation.mdx
@@ -56,6 +56,10 @@ Follow these steps to set up Lingui with Babel:
 When using a preset, first check if it includes the `macros` plugin. If so, then you don't need to install and set up `@lingui/babel-plugin-lingui-macro`. For example, `react-scripts` already includes the `macros` plugin.
 :::
 
+:::caution
+If you're using [React Compiler](https://react.dev/learn/react-compiler), make sure that `@lingui/babel-plugin-lingui-macro` comes **before** the React Compiler plugin in the Babel plugins list. This ensures macros are expanded correctly before the compiler processes the code.
+:::
+
 </TabItem>
 <TabItem value="swc" label="SWC">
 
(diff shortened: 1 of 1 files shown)


vonovak (approved):
nice catch

yslpn:
I think this message should be common for both tabs. 

1) Now we have the compiler only as a babel plugin. But soon it will be available as a swc plugin.
2) At the moment, users can use babel (react-compiler) + swc with lingui plugin.

cc @timofei-iatsenko @vonovak

gustavoharff:
> 2\. At the moment, users can use babel (react-compiler) + swc with lingui plugin.

@yslpn I tried setting up Babel + SWC but wasn’t successful, are you sure about this?

Next can do that and [here's how](https://nextjs.org/docs/app/api-reference/config/next-config-js/reactCompiler#how-it-works), but I don't know how to do that in Vite.

timofei-iatsenko:
> Next can do that and [here's how](https://nextjs.org/docs/app/api-reference/config/next-config-js/reactCompiler#how-it-works), but I don't know how to do that in Vite.

Speculating from this doc, most likely lingui macro will not work with nextjs when react compiler is enabled because the order of execution is not possible to change.

yslpn:
> > 2. At the moment, users can use babel (react-compiler) + swc with lingui plugin.
> @yslpn I tried setting up Babel + SWC but wasn’t successful, are you sure about this?

If you use [webpack](https://react.dev/learn/react-compiler#usage-with-webpack), it's extremely simple. Just changing the order of loaders in the array.

```js
const { reactCompilerLoader } = require("react-compiler-webpack");
...
    rules: [
      {
        test: /\.(js|jsx|ts|tsx)$/,
        exclude: /(node_modules)/,
        use: [
          {
            loader: "swc-loader",
          },
          {
            loader: reactCompilerLoader,
            options: {
              target: "18",
            },
          },
        ],
      },
...
```

gustavoharff:
> Speculating from this doc, most likely lingui macro will not work with nextjs when react compiler is enabled because the order of execution is not possible to change.

You're right, I tested here and the same problem happens.

We might consider adding an additional note [here](https://lingui.dev/tutorials/react-rsc#nextjs-config) or consolidating this information in the [Choosing a Transpiler](https://lingui.dev/installation#choosing-a-transpiler) section, and also sharing it with SWC.

What do you think?

gustavoharff:
I created a minimal reproduction for anyone who wants to visualize the problem.

https://github.com/gustavoharff/lingui-react-compiler

Instruction are in the readme.

timofei-iatsenko:
LGTM

timofei-iatsenko (approved):

jeanregisser:
Ah I just bumped into this issue, some random `<Trans>` were not rendering properly in production (just showing the msg id). I'm using Nextjs with react compiler enabled and the lingui swc-plugin.

Thanks for adding a note! 

I hope there will be a solution to this ordering issue in Nextjs 👀

PR #2313, opened 2025-08-14
feat: support compile with multi thread

Zxilly:
# Description

Offload the compile task to `worker_thread` to utilize the computing power of multiple cores.

## Types of changes

- [ ] Documentation update
- [ ] Examples update

This is a follow up of #2299

## Checklist

I'm not sure where I should add the documentation, but this patch introduces a new option `experiment.multiThread` .

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -81,7 +81,8 @@
     "picocolors": "^1.1.1",
     "pofile": "^1.1.4",
     "pseudolocale": "^2.0.0",
-    "source-map": "^0.8.0-beta.0"
+    "source-map": "^0.8.0-beta.0",
+    "threads": "^1.7.0"
   },
   "devDependencies": {
     "@lingui/jest-mocks": "*",
@@ -89,6 +90,7 @@
     "@types/micromatch": "^4.0.1",
     "@types/normalize-path": "^3.0.0",
     "mock-fs": "^5.2.0",
-    "mockdate": "^3.0.5"
+    "mockdate": "^3.0.5",
+    "ts-node": "^10.9.2"
   }
 }
--- a/packages/cli/src/api/catalog.test.ts
+++ b/packages/cli/src/api/catalog.test.ts
@@ -686,7 +686,7 @@ describe("writeCompiled", () => {
   ])(
     "Should save namespace $namespace in $extension extension",
     async ({ namespace, extension }) => {
-      const { source } = createCompiledCatalog("en", {}, { namespace })
+      const { source } = await createCompiledCatalog("en", {}, { namespace })
       // Test that the file extension of the compiled catalog is `.mjs`
       expect(await catalog.writeCompiled("en", source, namespace)).toMatch(
         extension
--- a/packages/cli/src/api/compile.test.ts
+++ b/packages/cli/src/api/compile.test.ts
@@ -181,8 +181,8 @@ describe("compile", () => {
 
 describe("createCompiledCatalog", () => {
   describe("options.namespace", () => {
-    const getCompiledCatalog = (namespace: CompiledCatalogNamespace) =>
-      createCompiledCatalog(
+    const getCompiledCatalog = async (namespace: CompiledCatalogNamespace) =>
+      (await createCompiledCatalog(
         "fr",
         {
           key: "Hello {name}",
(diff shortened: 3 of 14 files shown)


timofei-iatsenko:
I was very close to do it by myself, so thanks for the handling that. 

I have a few questions though: 

1. Why "threads" but not "priscina"? https://npm-compare.com/piscina,threads,workerpool it seems priscina much more popular

Also could you elaborate a bit about most notable changes, considerations, difficulties. to have more context for review. 

From my side, i believe it's possible to implement for custom extractor as well, i will spent some time to tackle this.

Zxilly:
1. Since I didn't know `priscina`, I initially wrote it directly using the native API of `worker_thread`. Then I realized that I needed a wrapper library, and after a quick search, I found `threads.js`, which seemed to meet my requirements.
2. I don't think there's anything special to note here. Just offload a function to a worker, design a serialization protocol that allows objects to be passed between threads, and that's all. The main issue is that we need to ensure that the results of multi-threaded execution are consistent with single-threaded execution, so I added a lot of extra sorting to get deterministic results.
3. I have no idea about custom extractors. How should the function be passed between threads? Perhaps `toString` and then `eval`, but if the extractor contains any additional imports, this won't work.

timofei-iatsenko:
@Zxilly i'll take it from here, thanks for contribution.

Zxilly:
I will also continue working on this patch. Perhaps we can standardize the way workers are handled? Or add more unit tests? I'll see what I can do.

timofei-iatsenko:
@Zxilly for some reason i could not push to you branch, so i created my own from yours and created a separate PR https://github.com/lingui/js-lingui/pull/2317

I will post updates in that PR

Zxilly:
@timofei-iatsenko I can enable `Allow edits from maintainers`, a new pr didn't allow me to continue working on this. Please reopen.

Zxilly:
@timofei-iatsenko I've create a new PR at #2318 with `Allow edits from maintainers`, you can try to push to that branch.

timofei-iatsenko:
I closed all other PR's and reopened this one. Here is my comment from other PR; 


> 1. I reverted multithreading changes for experemental extractor. The splitting work for worker pool was done in in the incorrect place. So instead of parallelezing work between workers, it actually create many parallel pools, with one worker only. This need more effort to make it work correctly, i would like to focus more on the stable extractor / compile first and then optimize new extractor. 
> 2. I added support for custom extractors as well, instead of passing a config as a parameter to a worker, i'm re-creating the config in every of them. I also updated jiti (a lib used to read esm/ts/js configs on the fly in nodejs) version to make sure that it using a cache for config transpilation, also i'm passing a resolved config path to a worker, to avoid extra config crawling. 
> 3. I simplified the code, and made it more aligned to the non-multithreading version. Instead of passing file content to the worker, worker reading fs by itself. 
> 
> I'm also going to review the compile approach, probably some changes would be done there as well. I'll keep you posted.

packages/cli/src/api/resolveWorkersOptions.ts
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  const multiThread = true
+
+  if (!opts.workers) {
+    if (cores <= 2) {
+      return { multiThread, poolSize: cores } // on tiny machines, use all
+    }
+    // on big machines cap to 8, to avoid trashing
+    return { multiThread, poolSize: Math.min(cores - 1, 8) }
```
Zxilly: Maybe `cpu() - 1` can be a better default value?

timofei-iatsenko: nodejs docs suggesting using `os.availableParallelism()` instead of `os.cpu()` exactly for this case. 

This is what llm says about it: 

### Key difference from `os.cpus().length`

* `os.cpus().length` simply enumerates all CPUs the OS knows about.
* `os.availableParallelism()` respects **process-level CPU affinity and quotas**:

  * If the process is restricted (e.g. via **cgroups**, **Docker container CPU limits**, or **cpuset affinity masks**), it returns the **actual number of CPUs available to the process**, not the whole machine.
  * Example: On a 64-core server, if your container is limited to `4` CPUs, `os.cpus().length` will still report `64`, but `os.availableParallelism()` will return `4`.

This is why the Node core team recommends it as the correct API for deciding thread pool sizes.

Zxilly: Good idea, I didn't know this before.

timofei-iatsenko:
Regarding the coverage - coverage is not gathered for worker files, it seems they are not get instrumented since they are spawned in a separate thread. 

The other files just extracted from existing places, so coverage level was not changed for them.

Zxilly:
https://github.com/bcoe/c8 This may be helpful for coverage collection.

timofei-iatsenko:
https://github.com/lingui/js-lingui/pull/2320

Issue #2336, opened 2025-10-06
Extracted ordering is not stable when same message occurs with multiple contexts


jtbandes:
### Description

The `extract` output is not stable when the same message text occurs with different contexts. I thought this was the same issue as https://github.com/lingui/js-lingui/issues/2306, but it may be different because this is about context, not placeholder.

### Verifications

- [x] I've checked [the docs](https://lingui.dev) and this isn't covered there.
- [x] I've searched existing issues on [GitHub](https://github.com/lingui/js-lingui/issues).

### Reproduction Steps

This test demonstrates the issue:

```diff
diff --git a/packages/cli/src/api/catalog.test.ts b/packages/cli/src/api/catalog.test.ts
index c8cc7ace..eeeacd3d 100644
--- a/packages/cli/src/api/catalog.test.ts
+++ b/packages/cli/src/api/catalog.test.ts
@@ -210,6 +210,8 @@ describe("Catalog", () => {
         mockConfig()
       )
 
+      expect(Object.keys(runA)).toEqual(Object.keys(runB))
+
       expect(Object.values(runA)[0].placeholders[0]).toStrictEqual(
         Object.values(runB)[0].placeholders[0]
       )
@@ -222,8 +224,6 @@ describe("Catalog", () => {
           ],
         }
       `)
-
-      // expect(messages).toMatchSnapshot()
     })
 
     it("should support experimental typescript decorators under a flag", async () => {
diff --git a/packages/cli/src/api/fixtures/collect-placeholders-sorting/a.ts b/packages/cli/src/api/fixtures/collect-placeholders-sorting/a.ts
index f046e741..b55fc549 100644
--- a/packages/cli/src/api/fixtures/collect-placeholders-sorting/a.ts
+++ b/packages/cli/src/api/fixtures/collect-placeholders-sorting/a.ts
@@ -4,4 +4,6 @@ function getUser() {
   return 'John'
 }
 
-t`Hello ${getUser()}`
\ No newline at end of file
+t`Hello ${getUser()}`
+
+t({ message: "With context", context: "one" })
diff --git a/packages/cli/src/api/fixtures/collect-placeholders-sorting/b.ts b/packages/cli/src/api/fixtures/collect-placeholders-sorting/b.ts
index 77e108df..696b85c8 100644
--- a/packages/cli/src/api/fixtures/collect-placeholders-sorting/b.ts
+++ b/packages/cli/src/api/fixtures/collect-placeholders-sorting/b.ts
[...]

PR #2337, opened 2025-10-06
fix: orderByMessage should also order by context

jtbandes:
# Description

Fixed a bug causing inconsistent ordering of extracted messages when messages differ only by context.

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes https://github.com/lingui/js-lingui/issues/2336

## Checklist

### DIFF
--- a/packages/cli/src/api/catalog.test.ts
+++ b/packages/cli/src/api/catalog.test.ts
@@ -617,7 +617,7 @@ describe("order", () => {
     expect(Object.keys(orderedCatalogs)).toMatchSnapshot()
   })
 
-  it("should order messages by message", () => {
+  it("should order messages by message and then by context", () => {
     const catalog = {
       msg1: makeNextMessage({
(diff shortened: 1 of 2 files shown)


vonovak (commented):
should also `comments` be taken into account in the ordering?

https://github.com/lingui/js-lingui/blob/319547b23e901f1feb0168e19d59b33f58d6ac82/packages/conf/src/types.ts#L31

jtbandes:
I don't know, perhaps a maintainer / someone more familiar with this code can provide insight on that? Would the same string with different comments produce multiple entries in the .po file, or would they be the same entry?

andrii-bodnar:
@jtbandes the same string, even with different comments, produces a single entry. Multiple entries for the same string are only created when the strings has a distinct `context` field.

jtbandes:
Ok, so perhaps this PR is fine as-is?

PR #2359, opened 2025-10-30
fix: improve type of values accepted by date/time/number formats

YarnSphere:
# Description

Modern ES versions support more types of values when calling `Intl.(DateTime|Number)Format.prototype.format` than those allowed by Lingui:
- Datetime formatting nowadays supports `Temporal` values (see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/format#parameters);
- Number formatting nowadays supports `BigInt`s and strings (see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/format#parameters).

This commit uses the types defined by TypeScript directly, rather than ad hoc types, which depend on the TypeScript target configured by the user, allowing users to pass to `date` and `number` any values that are supported by `Intl.(DateTime|Number)Format.prototype.format`.

**Example of code that currently fails to type check (even though it runs correctly):**

```ts
i18n.date(999999999999);
i18n.number(99n);
i18n.number("99.9");
```

## Types of changes

- [x] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/core/src/i18n.ts
+++ b/packages/core/src/i18n.ts
@@ -1,6 +1,12 @@
 import { interpolate } from "./interpolate"
 import { isString, isFunction } from "./essentials"
-import { date, defaultLocale, number } from "./formats"
+import {
+  date,
+  type DateTimeFormatValue,
+  defaultLocale,
+  number,
+  type NumberFormatValue,
+} from "./formats"
 import { EventEmitter } from "./eventEmitter"
 import { compileMessage } from "@lingui/message-utils/compileMessage"
 import type { CompiledMessage } from "@lingui/message-utils/compileMessage"
@@ -312,11 +318,14 @@ Please compile your catalog first.
    */
   t: I18n["_"] = this._.bind(this)
 
-  date(value: string | Date, format?: Intl.DateTimeFormatOptions): string {
+  date(
+    value?: string | DateTimeFormatValue,
+    format?: Intl.DateTimeFormatOptions
+  ): string {
     return date(this._locales || this._locale, value, format)
   }
 
-  number(value: number, format?: Intl.NumberFormatOptions): string {
+  number(value: NumberFormatValue, format?: Intl.NumberFormatOptions): string {
     return number(this._locales || this._locale, value, format)
   }
 }
--- a/packages/core/src/formats.ts
+++ b/packages/core/src/formats.ts
@@ -13,9 +13,14 @@ function normalizeLocales(locales: Locales): string[] {
 
 export type DateTimeFormatSize = "short" | "default" | "long" | "full"
 
+// These types vary depending on the TypeScript target (e.g., modern ES versions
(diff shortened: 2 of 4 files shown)


packages/core/src/i18n.ts
```diff
@@ -312,11 +318,11 @@ Please compile your catalog first.
    */
   t: I18n["_"] = this._.bind(this)
 
-  date(value: string | Date, format?: Intl.DateTimeFormatOptions): string {
+  date(value?: string | DateTimeFormatValue, format?: Intl.DateTimeFormatOptions): string {
```
timofei-iatsenko: Why value here become optional? 

Does the `DateTimeFormatValue` already include a `string`?

YarnSphere: > Why value here become optional?

I made it optional simply because it's supported, when the value is `undefined`, it returns the current date. Try out: `new Intl.DateTimeFormat().format()` (should return the current date), which should be similar to calling `i18n.date()`.

Let me know if you'd prefer it to **not** be optional, but note that it would still work in JavaScript, it simply wouldn't type check anymore.

> Does the `DateTimeFormatValue` already include a `string`?

No, see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/format#parameters (yet it does support numbers, even though MDN doesn't mention it).

---

By the way, you can test the changes in the Lingui repository itself just be changing the `"target": "es2017"` to something more modern in the root `tsconfig.json` (_e.g._, `"target": "ESNext"`).

YarnSphere: Just to give you a little bit more context: with `ESNext`, the latest `Intl.(DateTime|Number)Format.prototype.format` types are the following (omitting all non-relevant definitions):

```ts
declare namespace Intl {
    // Defined in `lib.es5.d.ts`:
    interface DateTimeFormat {
        format(date?: Date | number): string;
        // ...
    }

    // Defined in `lib.es2023.intl.d.ts`:
    type StringNumericLiteral = `${number}` | "Infinity" | "-Infinity" | "+Infinity";
    interface NumberFormat {
        format(value: number | bigint | StringNumericLiteral): string;
        // ...
    }
}
```

I believe `Temporal` isn't yet supported (this will most likely change in the future), and this PR should make it so that Lingui supports `Temporal` as soon as the TypeScript ES lib types are updated to support it.

Also, note that the `date` in `DateTimeFormat.prototype.format` is optional (as opposed to the `NumberFormat.prototype.format` value), which is why I changed Lingui's `i18n.date`'s value to be optional. Hope this helps. :)

timofei-iatsenko (approved):

YarnSphere:
Fixed Prettier issues in latest push.

PR #2377, opened 2025-12-05
feat(cli): add --output-prefix option to customize compiled file headers

Gamote:
# Description

This PR adds support for customizing the output prefix (header) in compiled message catalogs through a new `--output-prefix` CLI option.

Currently, Lingui hardcodes `/*eslint-disable*/` at the top of compiled files. This change allows projects to specify custom prefixes for generated files, making it easier to integrate with different tools like linters, coverage tools, or formatters that rely on header directives.

**Usage:**

```textmate
# Use default (/*eslint-disable*/)
lingui compile

# Use custom prefix for Oxlint
lingui compile --output-prefix "/*oxlint-disable*/"

# Use custom prefix for Biome
lingui compile --output-prefix "/*biome-ignore lint: auto-generated*/"

# No prefix at all
lingui compile --output-prefix ""
```

**Changes:**
- Added `outputPrefix` option to `CreateCompileCatalogOptions` type
- Updated `createCompiledCatalog` function to accept and use the custom prefix with default value `'/*eslint-disable*/'`
- Added CLI option `--output-prefix <prefix>`
- Passed the option through the compilation pipeline from CLI to the final output
- Updated documentation to reflect the new option name and usage

## Types of changes

- [ ] Documentation update
- [ ] Examples update

## Checklist

- [x] I have read the [[CONTRIBUTING](https://github.com/lingui/js-lingui/blob/main/CONTRIBUTING.md)](https://github.com/lingui/js-lingui/blob/main/CONTRIBUTING.md) and [[CODE_OF_CONDUCT](https://github.com/lingui/js-lingui/blob/main/CODE_OF_CONDUCT.md)](https://github.com/lingui/js-lingui/blob/main/CODE_OF_CONDUCT.md) docs

### DIFF
--- a/packages/cli/src/api/compile.test.ts
+++ b/packages/cli/src/api/compile.test.ts
@@ -288,15 +288,15 @@ describe("createCompiledCatalog", () => {
     })
   })
 
-  describe("options.lintDirective", () => {
-    const getCompiledCatalog = (lintDirective?: string) =>
+  describe("options.outputPrefix", () => {
+    const getCompiledCatalog = (outputPrefix?: string) =>
       createCompiledCatalog(
         "en",
         {
           Hello: "Hello",
         },
         {
-          lintDirective,
+          outputPrefix,
         }
       ).source
 
@@ -306,19 +306,24 @@ describe("createCompiledCatalog", () => {
     })
 
     it("should use oxlint-disable directive", () => {
-      const result = getCompiledCatalog("oxlint-disable")
+      const result = getCompiledCatalog("/*oxlint-disable*/")
       expect(result).toContain("/*oxlint-disable*/")
     })
 
-    it("should use custom lint directive when specified", () => {
-      const result = getCompiledCatalog("biome-ignore lint: auto-generated")
+    it("should use custom prefix when specified", () => {
+      const result = getCompiledCatalog("/*biome-ignore lint: auto-generated*/")
       expect(result).toContain("/*biome-ignore lint: auto-generated*/")
       expect(result).not.toContain("eslint-disable")
     })
 
-    it("should handle empty string directive", () => {
+    it("should handle empty string prefix (no header)", () => {
       const result = getCompiledCatalog("")
-      expect(result).toContain("/**/")
+      expect(
+        result.startsWith("export") ||
+          result.startsWith("module") ||
+          result.startsWith("window") ||
+          result.startsWith("global")
+      ).toBeTruthy()
     })
   })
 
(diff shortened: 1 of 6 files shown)


andrii-bodnar (commented):
Thank you for the contribution, @Gamote! It looks useful!

Could you also please update the [documentation](https://lingui.dev/ref/cli#compile)? It would be great to include value examples for popular linters like Oxlint in your case or Biome.

Gamote:
> Thank you for the contribution, @Gamote! It looks useful!
> 
> Could you also please update the [documentation](https://lingui.dev/ref/cli#compile)? It would be great to include value examples for popular linters like Oxlint in your case or Biome.

@andrii-bodnar Sure thing, good call. I just pushed clear documentation and working examples.
I have also pushed unit and integration tests for the new `--lint-directive`.

timofei-iatsenko:
IMO, looks a bit overcoupled to the usage. Would it be better to just have "prefix / suffix" ? Than it is not coupled and user could do whatever and don't break the semantics.

andrii-bodnar:
> IMO, looks a bit overcoupled to the usage. Would it be better to just have "prefix / suffix" ? Than it is not coupled and user could do whatever and don't break the semantics.

If think about a more generic solution, do some other use-case is possible except lint directive? Is there any use-case for having a suffix there as well?

timofei-iatsenko:
Ignoring by prettier, which is semantically not a linter. Instanbul coverage ignoring and so on. 

Not sure about "suffix", but something like `header` or `prefix` would be definetily more semantically correct solution then narrowing it to the `--lint-directive`

Gamote:
> Ignoring by prettier, which is semantically not a linter. Instanbul coverage ignoring and so on.
> 
> Not sure about "suffix", but something like `header` or `prefix` would be definetily more semantically correct solution then narrowing it to the `--lint-directive`

I can see how that can be useful, also the suffix could be of help if there are tools that need to mark the start and end.

How about `--output-prefix` and `--output-suffix`? With `--output-prefix` default being `/*eslint-ignore*/` to preserve the current behaviour. (@andrii-bodnar @timofei-iatsenko)

If this is something you want I can updated the code, tests and docs.

andrii-bodnar:
For now, I would suggest adding only the prefix. We can introduce a suffix later if necessary.

Gamote:
@andrii-bodnar I have pushed the refactoring, updated the docs, the tests and the PR title and description.

Can you please review the changes?

timofei-iatsenko:
@andrii-bodnar @Gamote LGTM

PR #2368, opened 2025-11-20
feat!: ESM-only conversion

timofei-iatsenko:
# Description

https://github.com/lingui/js-lingui/issues/2363

Most noticable changes: 

1. Packages migrated to ESM-only, except `@lingui/babel-plugin-lingui-macro`, `@lingui/loader`, `@lingui/macro` (deprecated). Those excluded packages still using ESM depdendecies, so it will affect theirs consumers. Those packages was not converted to esm because there is no sense in that, babel and webpack consume them using a `require` function. 
2. Jest replaced to Vitest.  I wasn't able to make Jest working with ESM modules without hacks. Vitest just worked out of the box with no issue
3. Every package now has it's own tsconfig and build pipeline, since packages has diffrent targets (one for nodejs without a bundler, other for a bundler)
4. The monorepo layout changed from using typescript path aliases to consume built artifacts. This is consequence of the point above. That ruined the whole dx inside of the monorepo, but this is temporarly step to make it work first, in next iterations i will try to improve development experience.

[//]: # (Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.)

## Types of changes

[//]: # (What types of changes does your code introduce to Lingui?)
[//]: # (_Put an `x` in the boxes that apply_)

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [ ] I have read the [CONTRIBUTING](https://github.com/lingui/js-lingui/blob/main/CONTRIBUTING.md) and [CODE_OF_CONDUCT](https://github.com/lingui/js-lingui/blob/main/CODE_OF_CONDUCT.md) docs
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added the necessary documentation (if appropriate)

PR #2320, opened 2025-08-22
feat(cli): support multithreading in all cli commands

timofei-iatsenko:
# Description

Since i rewrite almost everethying from the [original PR](https://github.com/lingui/js-lingui/pull/2313), i'm reopening this from myself. 

## Feature description

This PR adds support of thread pool (using [threads-js](https://threads.js.org/)) for compile, extract, extract-template and experimental extractor commands. 

Here is how tasks split between workers in commands: 

- `compile`: by locale. So en, ru, pl, etc will go each into it's own worker. Why not by files (catalogs)? Because of the mergeCatalogs feature (where all files merged into one) and because, as i believe the message compiling itself doesn't take too much time, but serializing this into the file can do (BTW, this is for the future iterations, that shouldn't take that long). In any way, splitting by locales should give very even distribution across workers because amount of job for every locale is the same. 
- `extract` / `extract-template`: by source files
- `experimental-extractor`: by bundles, each bundle after esbuild are processed by separate worker in the pool. So if experimental extractor used for pages in the MPA applications, every page first bundled by esbuild and then processed in parallel by worker pool

Multhreading is enabled by default. Users can disable or configure it's behavior using `--workers` argument for cli. Docs were added. 

Other notable changes: 

Due to the fact that you can send between workers / main thread only serializable data structures, you could not send a class with methods. So i did some refactorings to internal structures to use less classes/instances and more plain structures and pure functions. 

[//]: # (Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.)

## Types of changes

[//]: # (What types of changes does your code introduce to Lingui?)
[//]: # (_Put an `x` in the boxes that apply_)

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

[...]

Issue #2235, opened 2025-04-22
Extracted multiline PO format IDs diff


MikaelSiidorow:
### Description

Extracting with the `@lingui/format-po` formatter for multiline strings:
```tsx
t`Something\nor something else`
```

Generates `msgid`s like:
```po
msgid ""
"Something\n"
"or something else"
```

But platforms, such as Crowdin will strip this initial empty string from the id, resulting in :
```po
msgid "Something\n"
"or something else"
```

Which will cause lost connection between these source strings when extracting again. (Looks like GitHub code highlight also reads them differently, which might be hinting there's a problem 👀 )

### Verifications

- [x] I've checked [the docs](https://lingui.dev) and this isn't covered there.
- [x] I've searched existing issues on [GitHub](https://github.com/lingui/js-lingui/issues).

### Reproduction Steps

Have a source string as described above.

Extract according to docs with
`lingui extract`

Push and pull from Crowdin according to docs
`crowdin push && crowdin pull`

Extract again
`lingui extract`

See unexpected diff

### Expected Behavior

IDs would be generated in a logical way that there's no diff. If this is an issue with Crowdin, then it's likely out of scope for Lingui to fix, but it seems there's mutual interest to keep the interop as seamless as possible

### Macro Support

SWC with @lingui/swc-plugin

### Lingui Version

5.3.0

### Framework

Vite

MikaelSiidorow:
Note: It looks like there's a potential workaround if using the `printLinguiId` option, as it seems to re-connect these strings when extracting, but without that option it doesn't work.

```ts
format: formatter({
    printLinguiId: true,
  })
```

Regardless, it's a bit annoying to have this need to re-extract after syncing the translations, as the expected behavior would be to use similar ids for both

MikaelSiidorow:
Might or might not be related:
Using both explicit ID and extracted ID with same message, such as:
```tsx
// server-side, no macro:
i18n._("Cancel") // explicit id

// client-side macro:
{<Trans>Cancel</Trans>} // extracted
```

Generates two separate strings with the same `msgid` in the `messages.po` file, generating an invalid(-ish) file with duplicate `msgid`s

Let me know if this should be in a separate issue

timofei-iatsenko:
This is another one problem related to upstream [`pofile`](https://www.npmjs.com/package/pofile), honestly i'm not a pro in how the correct po file should looks like. @andrii-bodnar could you ask the chaps from crowdin on that topic? 

[...]

Issue #2098, opened 2024-11-28
PO translator comments must begin with # and space


mlocati:
**Describe the bug**

The .po files downloaded from Transifex start with some translator comments, for example:

```gettext
# 
# Translators:
# Translator 1, 2024
# Translator 2, 2024
# 
msgid ""
msgstr ""
```

Please remark that the first and the 5th lines contain a `#` comment followed by a space (`# `) and not just a `#`
When lingui reads and updates such .po files, the trailing space is stripped out (that is, the lines `# ` become `#).

[As per the gettext specs](https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html), translator comments *always* start with a `#` followed by a space:
> There are two kinds of comments: those which have some white space immediately following the # - the translator comments -
> [...]

So, lingui should keep those trailing spaces

**To Reproduce**

1. generate a .po file with `lingui extract`
2. add an empty translator comment (`#` followed by a space)
3. run `lingui extract` again
4. The line with the empty translator comment will become a `#` **NOT** followed by a space
`

**Expected behavior**

Lines `# ` should still be `# ` and not `#` when lingui regenerates them.

timofei-iatsenko:
Thanks for the report. It seems it's an issue in the underlaying `pofile` library. Another good reason to drop it in favor of own implementation

PR #2379, opened 2025-12-09
refactor: replace the `pofile` dependency with `pofile-ts`

swernerx:
Replaces the `pofile` dependency with [`pofile-ts`](https://github.com/sebastian-software/pofile-ts), a modernized fork with full TypeScript support.

**Why this change?**
- **TypeScript-first**: Native type definitions, no `@types/` package needed
- **Zero dependencies**: Smaller install footprint  
- **Modern API**: Promise-based `save()` method instead of callbacks
- **Maintained**: `pofile` last release was December 2022

**Affected packages:**
- `@lingui/format-po`
- `@lingui/format-po-gettext`
- `@lingui/cli`

**Code improvements:**
- Use exported types (`Item`, `Headers`) directly from `pofile-ts`
- Convert callback-based `po.save()` to Promise-based API
- Remove manual type definitions

**Other:**
- Add `pofile-ts` to Jest `transformIgnorePatterns` for ESM support
- Add `.lingui/` to `.gitignore` (test cache artifacts)
- Document required `yarn release:build` step in CONTRIBUTING.md

## Types of changes

- [x] Documentation update
- [ ] Examples update

## Checklist

### DIFF
--- a/packages/format-po-gettext/src/po-gettext.ts
+++ b/packages/format-po-gettext/src/po-gettext.ts
@@ -1,6 +1,12 @@
 import { parse as parseIcu, Select, SelectCase } from "@messageformat/parser"
 import pluralsCldr from "plurals-cldr"
-import { parsePo, stringifyPo, createItem, type PoItem, type SerializeOptions } from "pofile-ts"
+import {
+  parsePo,
+  stringifyPo,
+  createItem,
+  type PoItem,
+  type SerializeOptions,
+} from "pofile-ts"
 import gettextPlurals from "node-gettext/lib/plurals"
 
 import type { CatalogFormatter, CatalogType, MessageType } from "@lingui/conf"
@@ -552,7 +558,10 @@ export function formatter(
         )
       })
 
-      return formatter.parse(stringifyPo(po, serializeOptions), ctx) as CatalogType
+      return formatter.parse(
+        stringifyPo(po, serializeOptions),
+        ctx
+      ) as CatalogType
     },
 
     serialize(catalog, ctx): string {


timofei-iatsenko:
FYI https://github.com/timofei-iatsenko/pofile

swernerx:
> FYI https://github.com/timofei-iatsenko/pofile

Interesting... I was not aware of any other work on this... I assume it's related but mine might be even more up-to-date also covering a series of new test cases.

timofei-iatsenko:
@swernerx The background of this package is that I initially made the update for the same reasons you mentioned - to modernize it, fix the TypeScript typing issues, and generally bring it up to date. My intention, however, was to transfer it under the Lingui organization. This package is quite important for Lingui, and ideally it should be maintained by the community or the organization rather than by a random guy from the internet (in this case, me).

I faced some resistance from the Lingui maintainers on moving it into the organization, which is why the package hasn’t been published yet.

By the way, the original implementation has several known issues in Lingui repo (https://github.com/lingui/js-lingui/issues/2235 https://github.com/lingui/js-lingui/issues/2098). I still believe there’s real value in bringing this package into the organization and addressing everything together as a single project.

Another option would be completely change implementation for the https://www.npmjs.com/package/gettext-parser not sure how close this to the gettext spec though

swernerx:
@timofei-iatsenko Thanks for the context! I looked at the previous approach, but I'd argue against integrating PO parsing directly into Lingui for a few reasons:

1. **Scope creep**: Lingui already covers a lot of ground—SWC, Babel, macros, Vite, Metro, etc. Adding PO file parsing to that mix doesn't make it better, just bigger. Keeping concerns separated makes maintenance easier for everyone.

2. **Standalone value**: We have a genuine need for a standalone PO library outside of Lingui. Others likely do too. A separate package serves that use case.

3. **Minimal migration effort**: As you can see from this PR, switching to `pofile-ts` is trivial—just a few import changes and one API adjustment (`save()` → `writeFile()` + `toString()`).

4. **What's new in pofile-ts**: I've taken inspiration from your earlier work and modernized it further—removed the filesystem methods (pure parsing/serialization now, zero dependencies), cleaned up the source, added proper TypeScript support with exported types, and kept the same test coverage. The main goal here was simply to stop maintaining manual type definitions in Lingui.

I think keeping PO parsing as a separate, well-maintained library is the more pragmatic path forward.

timofei-iatsenko:
@swernerx it's not about integrating pofile into lingui, it's about hosting the package under the lingui org. So the package would be still separate, installable, but would be at full control of the lingui organization and with more trust in it from the consumers.

The problem with packages implemented by some random guy, is that, people usually don't realize how much work it would be for them after releasing something like that and burnout quite quickly leaving the package abandoned.

swernerx:
Fair point about maintainer burnout—I've been doing open source for years and know exactly what you mean. That said, I'd suggest a step-by-step approach here:

1. **This PR is low-risk**: Switching to `pofile-ts` is trivial and doesn't make anything worse. It just modernizes a dependency that hasn't been updated in two years.

2. **Future integration is possible**: If the Lingui team later decides to bring PO parsing under the organization—in whatever form—I'm open to that conversation. But that decision doesn't need to block this straightforward dependency update.

3. **Lingui already depends on many external packages**: The PO parser is important, sure, but it's one of dozens of dependencies. Babel, SWC, Vite plugins, Metro transformer, messageformat packages—all maintained by "some random person" at some level. The same trust argument could apply to any of them.

4. **It's all open source**: Since everything is MIT licensed, the Lingui team can fork, integrate, or adopt the code at any time—no alignment or permission needed. That decision is entirely yours to make whenever it makes sense.

Let's not let perfect be the enemy of good. This PR improves the status quo today. Organizational decisions about package ownership can happen separately.

packages/format-po-gettext/src/po-gettext.ts
```diff
@@ -1,6 +1,6 @@
 import { parse as parseIcu, Select, SelectCase } from "@messageformat/parser"
 import pluralsCldr from "plurals-cldr"
-import PO from "pofile"
+import PO, { Item as POItem } from "pofile-ts"
```
timofei-iatsenko: I recommend to get rid of default export completely, it's a pain point in CJS/ESM packages

jest.config.js
```diff
@@ -9,7 +9,7 @@ const tsConfigPathMapping = pathsToModuleNameMapper(
 )
 
 const testMatch = ["**/?(*.)test.(js|ts|tsx)", "**/test/index.(js|ts|tsx)"]
-const transformIgnorePatterns = ["node_modules/(?!@messageformat)"]
+const transformIgnorePatterns = ["node_modules/(?!@messageformat|pofile-ts)"]
```
timofei-iatsenko: I saw that your package is distributed as dual cjs/esm, if everything setup correctly for the dual package jest should consume CJS version and this should not be needed. 

If it's still needed - it means more serious problem here.

 it means commonjs code of linguijs would require a esm package which was not possible until nowdays, but fortunately [possible today](https://socket.dev/blog/require-esm-backported-to-node-js-20). However it will bump a minimal compatible version of nodejs for users to 20.16+, 22.19+, or 24+ which i would consider a breaking change

swernerx: Should really switch over to Vitest ;) But anyway... I think 20.16+ should be fine. I tried removing it and it seems to work. Previously it was ESM only... that's the original reason for the modification.

timofei-iatsenko: FYI https://github.com/lingui/js-lingui/pull/2368

swernerx: There it is ... Vitest :)

packages/format-po-gettext/src/po-gettext.ts
swernerx: I just released v3.0.0 with modular functions. No `PO` class anymore. No default exports. This is now also updated in this PR.

timofei-iatsenko:
@swernerx meanwhile maybe you can work on these problems in your fork:

---
Multiline formatting issue: https://github.com/lingui/js-lingui/issues/2235 

Here i would suggest to have a wrap or fold option similar to gettext-parser's https://www.npmjs.com/package/gettext-parser#:~:text=with%20possible%20values-,foldLength,-is%20the%20length

> foldLength is the length at which to fold message strings into newlines (default: 76). Set to 0 or false to disable folding. 

This later might be exposed as the formatter parameter. 

Also would be nice to dig into original source and specs to understand why `pofile` creates an empty string when multiline message: 

```po
msgid ""
"Something\n"
"or something else"
```

This change might be a disruptive a bit, i would not consider it's breaking because, files generated would be still consumable  and result should not be changed, but changes here will definitely produce a diff in consumers codebases.  

---

PO translator comments must begin with # and space https://github.com/lingui/js-lingui/issues/2098

This should be easy to fix. Will also produce diff but not breaking.

swernerx:
> @swernerx meanwhile maybe you can work on these problems in your fork:
> 
> Multiline formatting issue: #2235
> 
> Here i would suggest to have a wrap or fold option similar to gettext-parser's https://www.npmjs.com/package/gettext-parser#:~:text=with%20possible%20values-,foldLength,-is%20the%20length
> 
> > foldLength is the length at which to fold message strings into newlines (default: 76). Set to 0 or false to disable folding.
> 
> This later might be exposed as the formatter parameter.
> 
> Also would be nice to dig into original source and specs to understand why `pofile` creates an empty string when multiline message:
> 
> ```po
> msgid ""
> "Something\n"
> "or something else"
> ```
> 
> This change might be a disruptive a bit, i would not consider it's breaking because, files generated would be still consumable and result should not be changed, but changes here will definitely produce a diff in consumers codebases.
> 
> PO translator comments must begin with # and space #2098
> 
> This should be easy to fix. Will also produce diff but not breaking.

Indeed my fork had the same issue. I fixed that and also implemented two new options to tweak it as needed for different usage models.

Fixes: https://github.com/lingui/js-lingui/pull/2379 (hopefully) ;)

packages/format-po/src/po.ts
```diff
@@ -107,6 +108,40 @@ export type PoFormatterOptions = {
...
+   * msgid "First line\n"
+   * "Second line"
+   * ```
+   *
+   * When `false`, uses GNU gettext's traditional format with an empty first line:
+   * ```po
+   * msgid ""
+   * "First line\n"
+   * "Second line"
+   * ```
```
timofei-iatsenko, line 136: Do you have any links to read about this traditional format? I'm wondering was it a indeed a traditional and made intentionally by author of the pofile or it was done unintentionally and we keeping them unnecessarily

timofei-iatsenko:
@swernerx both po formatters has this ugly piece of code: 

```ts
       // accessing private property
        ;(po as any).headerOrder = Object.keys(po.headers)
```

The `as any` assertion is no longer needed with your fork. Could you change it, please?

packages/format-po/src/po.ts
swernerx: The empty string convention for multiline entries in PO files is documented in the GNU gettext manual, section "The Format of PO Files":

Link: https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html

The format uses C-style string literal concatenation — adjacent quoted strings are automatically joined. The empty string "" on the first line is a stylistic convention that:

- Improves alignment of subsequent lines 
- Enhances readability
- Has no semantic effect (empty string + rest = rest)

The underlying mechanism comes from the C specification (ISO C99 Section 6.4.5 "String literals" and 5.1.1.2 "Translation phases", Phase 6).

swernerx:
> @swernerx both po formatters has this ugly piece of code:
> 
> ```ts
>        // accessing private property
>         ;(po as any).headerOrder = Object.keys(po.headers)
> ```
> 
> The `as any` assertion is no longer needed with your fork. Could you change it, please?

Fixed

timofei-iatsenko:
@swernerx now you need to update and fix conflicts from the main branch.

swernerx:
> @swernerx now you need to update and fix conflicts from the main branch.

Done

packages/format-po/src/po.ts
timofei-iatsenko: For reference the exact example with detailed description is here  
https://www.gnu.org/software/gettext/manual/html_node/More-Details.html#Further-details-on-the-PO-file-format

What i'm thinking, is that if this pattern is stated in the official docs - this should be a default, not the "compact" style, wdyt? 

I would also put a link to this doc into the flag description.

swernerx: I was a bit undecided here. My thinking was mostly shaped by the Crowdin use case and other automated tooling I’ve worked with recently – for those tools the concrete PO formatting is effectively irrelevant. Since Crowdin is one of the bigger players, I considered its behavior a reasonable “de-facto” default. In our setup people almost never open the PO files manually and 99.9% of translations happen fully automatically, so readability of multiline entries hasn’t really been a concern so far.

That said, if the GNU-style empty-string convention is explicitly recommended in the official docs and improves readability for anyone who does look at the files, I’m perfectly fine with using that as the default and treating the compact form as an alternative style.

timofei-iatsenko:
@swernerx it's failing on the tests now. 

I'm wondering may be you can also look at the performance side of that pofile parsing? We had a discussion https://github.com/lingui/js-lingui/pull/2320#issuecomment-3244865575 some time ago which you might be interested to follow.  TL; DR; When you have a quite big catalog parsing is taking a significant time. 

I'm not sure is it something that could be improved a lot, but at least you could give it a try.

timofei-iatsenko:
@swernerx CI still failing

packages/format-po/src/po.ts
timofei-iatsenko: > Since Crowdin is one of the bigger players, I considered its behavior a reasonable “de-facto” default

I would consider a deafult example written in the gettext official docs. 

> In our setup people almost never open the PO files manually and 99.9% of translations happen fully automatically, so readability of multiline entries hasn’t really been a concern so far.

Same here.

timofei-iatsenko:
@swernerx Is there anything to add or we can review and move this forward?

timofei-iatsenko:
@swernerx ping, there are conflicts to be resolved. Thanks.

andrii-bodnar:
Hi @swernerx, @fastner is there any chance of moving this forward soon? Thanks!

Issue #2363, opened 2025-11-10
V6 Roadmap: ESM Only distribution


timofei-iatsenko:
The JavaScript ecosystem keeps moving forward, and more and more projects are switching to **ESM-only** distribution.

I’ve tried to make this change a few times already, but each time I was blocked by various incompatibilities. LinguiJS is a complex project consisting of multiple parts — runtime, CLI, and plugins for different tools.

Switching to ESM-only requires ensuring **interoperability** across all these tools. The main roadblock was usually supporting ESM in the `babel-plugin-macros`. I even contributed to Babel itself https://github.com/babel/babel/issues/16860 to make this possible in the future.

However, the `babel-plugin-macros` codebase is quite outdated, and before working on the original issue, a lot of maintenance work was needed just to bring everything up to date. Because of that, I had to postpone this task for a long time.

Fortunately, the ecosystem has evolved, and with recent Node.js improvements — such as [`require(esm)` backported to Node.js 20](https://socket.dev/blog/require-esm-backported-to-node-js-20) — this workaround is no longer necessary. That means we can now distribute LinguiJS as pure ESM.

I expect no major changes will be required from users unless they rely on very specific or non-standard setups. Still, since this is a significant shift, it should be released as a **new major version**, similar to what [Storybook 10 did](https://storybook.js.org/blog/storybook-10/).

I don’t plan to include any other major changes in this release — this breaking change will be big enough on its own.

Testing checklist: 

- Ensure support for webpack
- Ensure support for `babel-plugin-macros`
- Ensure support for `metro` 

Bump minimal nodejs version requirement to 20.16+, 22.19+, or 24+

## Blockers:

- [x] https://github.com/MartinCerny-awin/pseudolocale/pull/27
- [x] https://github.com/andywer/threads.js/issues/496

## Another fixes to be included in the V6 release

- [ ] Generate Lingui ID with URL and filename safe alphabet  #2207 
- [x] Replace cosmiconfig with lilconfig in configuration handling #2286

## To do List:

- [x] Merge ESM Conversion #2368
- [x] typecript's `strict: true` & `"strictNullChecks": false,` for most of packages (cli + core + react) #2412
- [x] rename `jest-mocks` package into `test-utils` #2413
- [ ] test on a REAL projects with all supported tooling integrations (webpack, metro, babel-macros)
- [ ] spike: extract cli api from cli package to separate cli-lib.
- [x] reconsider `@lingui` peerDependencies
[...]

PR #2386, opened 2025-12-16
chore: deprecate Intl wrapper helpers and document removal

### LINKED ISSUE #2265: Let's mark the wrappers over Intl methods as deprecated
### Problem Description

When I faced the need to use the `Intl.ListFormat` and `Intl.NumberFormatOptions` methods, I could not understand why the library had wrappers for `NumberFormatOptions`, but no `ListFormat`. I had to spend time to find https://github.com/lingui/js-lingui/pull/2223, but it seems that there is no final solution. New methods are not added, and old method wrappers remained in i18n for current version of lingui.

The main problem is wasting time on research and misunderstanding the best practices that should be used in your project.

### Proposed Solution

I suggest marking the wrapper methods over `Intl` as deprecated via `JSDoc` for a while, and removing them in the next major. Also leave a comment about this in the documentation, that the recommended way is to use Intl methods directly.

### Alternatives Considered

Maybe you could just add to the documentation an explanation of why only two `Intl` methods have wrappers.

### Additional Context

_No response_

yslpn:
# Description

Deprecate Lingui’s Intl wrappers (`i18n.date/number` plus the shared date/time/number helpers) ahead of removal, and add documentation callouts to direct users to native `Intl.DateTimeFormat`/`Intl.NumberFormat`. Related: https://github.com/lingui/js-lingui/issues/2265.

Open question for @timofei-iatsenko: do we also want a short blog for this deprecation, and is it OK to plan removal in the next major given the ESM-only focus discussed in https://github.com/lingui/js-lingui/issues/2363? Or it would be better to change the deprecated message.

## Background

Initially, Lingui aimed to be a Swiss Army knife for localization, trying to take on as much as possible: from DX to performance optimization. We strived to create convenient formatting wrappers with built-in memoization, normalization, and so on. Now we have functions like `i18n.date` and `i18n.number`. But this prevents us from moving forward, and we need to abandon them.

## Problems

Creating such wrappers around native APIs leads to the following problems:

This is additional code that needs to be supported and maintained. The library does too much. We need to do less, but do it well. Lingui JS should focus on message extraction and working with catalogs. Formatting dates, numbers, and so on is the job of the `Intl` API.

These wrappers increase bundle size. Not all users use them. For example, there may be applications where date formatting doesn't depend on locale and is configured manually by the user. Tree-shaking doesn't work here because these are class methods.

We can't create wrappers for all Intl methods—there are too many of them https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl. This leads to poor DX when it's unclear what's in the library and what's not. This is exactly what prompted me to create an issue when I couldn't find how to use Intl.NumberFormat in Lingui.

## Plan

In a minor release, we're publishing a blog post, updating documentation, and adding deprecated warnings for methods via JSDoc comments.

In one of the major releases, we will remove the methods, but before that, we'll give enough time for users to prepare their code.

## Migrating to native Intl

The migration principle is simple: we need to replace the use of `i18n.date` and `i18n.number` methods with direct use of `Intl.DateTimeFormat` and `Intl.NumberFormat` respectively.

- `i18n.date(date, options)` => `new Intl.DateTimeFormat(i18n.locale, options).format(date);`
- `i18n.number(n, options)` => `new Intl.NumberFormat(i18n.locale, options).format(n);`

Example without React:

```ts
import type { I18n } from "@lingui/core";

const dateOptions = { dateStyle: "medium" } as const;
const currencyOptions = {
  style: "currency",
  currency: "EUR",
} as const;

export function formatOrderSummary(i18n: I18n, date: Date, total: number) {
  const dateFormatter = new Intl.DateTimeFormat(i18n.locale, dateOptions);
  const numberFormatter = new Intl.NumberFormat(i18n.locale, currencyOptions);

  // "Jan 18, 2026 - €1,234.56"
  return `${dateFormatter.format(date)} - ${numberFormatter.format(total)}`;
}
```

In React you have to use `useMemo` to memoize formatters and react to locale changes:

```tsx
import { useMemo } from "react";

const dateOptions = { dateStyle: "medium" } as const;
const priceOptions = {
  style: "currency",
  currency: "USD",
  minimumFractionDigits: 2,
} as const;

function PriceLine({ date, total }: { date: Date; total: number }) {
  const { i18n } = useLingui();

  const dateFormatter = useMemo(
    () => new Intl.DateTimeFormat(i18n.locale, dateOptions),
    [i18n.locale]
  );
  const numberFormatter = useMemo(
    () => new Intl.NumberFormat(i18n.locale, priceOptions),
    [i18n.locale]
  );

  return (
    <span>
      {/* "Jan 18, 2026 - $1,234.56" */}
      {dateFormatter.format(date)} - {numberFormatter.format(total)}
    </span>
  );
}
```

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [x] Documentation update
- [ ] Examples update

Fixes #2265

## Checklist


timofei-iatsenko:
Hey @yslpn thanks for the contribution. I'm not sure that  developers would have enough time to transition and that will not block adoption ESM only linguijs. 

So it would be better to not remove it in the next major version.  Blog post would be much apreciated to explain what was the driver for the decision. 

Meanwhile, do you think it would be nice to have ready to use react hooks as an alternative? Because users will still have to create them on their own, may be we can create them and make available out of the box?

yslpn:
> Meanwhile, do you think it would be nice to have ready to use react hooks as an alternative? Because users will still have to create them on their own, may be we can create them and make available out of the box?

If we ship ready-made React hooks for date/number formatting, people will soon want the same helpers outside React, and we end up back where we were: maintaining wrappers around every `Intl` method. I'd prefer to keep recommending native `Intl.*` directly. If someone needs helpers, they can create their own functions or hooks.

Does that sound good to you?

> Blog post would be much apreciated to explain what was the driver for the decision.

I also drafted a blog post to explain the deprecation and migration; would love your take on it. The social image in the post was generated via AI (Nano Banana by Gemini).

website/blog/2025-12-17-deprecating-intl-wrappers/index.md
```diff
@@ -0,0 +1,113 @@
...
+- Special cases already covered. We do not need an `Intl.PluralRules` wrapper because plural selection lives inside the messageformat compiler (`@lingui/message-utils`). Exposing another public API only confuses the story.
+
+## What full wrapper parity would require
+
+To meet the expectations for a complete helper surface we would need wrappers and caching at least for `DateTimeFormat`, `NumberFormat`, `ListFormat`, `RelativeTimeFormat`, `DisplayNames`, `Collator`, `Segmenter`, `Locale`, `DurationFormat` (once it ships), plus `formatRange` and `formatToParts`. We are intentionally not going to maintain this matrix.
+
+## Plan
+
+- Now: mark the helpers as deprecated in code and docs without breaking existing apps.
+- Upcoming majors: keep them long enough to avoid blocking the ESM transition and give teams time to migrate; removal will happen in one of the next majors (not necessarily the very next one).
```
vonovak:
```suggestion
- Upcoming minor: runtime warning with instructions on how to migrate
- Upcoming majors: keep them long enough to avoid blocking the ESM transition and give teams time to migrate; removal will happen in one of the next majors (not necessarily the very next one).
```

website/blog/2025-12-17-deprecating-intl-wrappers/index.md
```diff
@@ -0,0 +1,113 @@
...
+- Removal: no separate warning blog is planned - please track the release notes for major versions.
+
+## Migrating to native Intl
+
+Pick the same locale precedence (`i18n.locales ?? i18n.locale`) and build your own formatters. Define helpers that read the current locale every time, so locale switches stay in sync:
```
vonovak:
```suggestion
In principle, the migration is simple: `i18n.date(dateInstance, formatOptions)` becomes `dateFormattingFn(i18n, dateInstance)`. Pick the same locale precedence (`i18n.locales ?? i18n.locale`) and build your own formatters. Define helpers that read the current locale every time, so locale switches stay in sync:
```

website/blog/2025-12-17-deprecating-intl-wrappers/index.md
```diff
@@ -0,0 +1,113 @@
...
+
+export function formatOrderSummary(date: Date, total: number) {
+  const dateFormatter = getDateFormatter({ dateStyle: "medium" });
+  const numberFormatter = getNumberFormatter({
+    style: "currency",
+    currency: "EUR",
+  });
+
+  return `${dateFormatter.format(date)} - ${numberFormatter.format(total)}`;
+}
```
vonovak: This suggested code has two problems:

1) it assumes that `i18n` instance comes from `@lingui/core`. We should recommend code that is agnostic of the `i18n` instance origin. It could come from `@lingui/core`, from `react context` or other place. We should not assume.
2) it's suboptimal because for each `formatOrderSummary` we create a new instance of the `Intl` objects. Let's aim to not create new objects if not really needed.

this is a sketch:

```
function getDateFormatter(locales: Locales | undefined) {
  const options: Intl.DateTimeFormatOptions = {
    year: "numeric",
    month: "long",
    day: "numeric",
  }
  return new Intl.DateTimeFormat(locales, options);
}

const memoizedGetDateFormatter: typeof getDateFormatter = memoize(getDateFormatter);

function getNumberFormatter(locales: Locales | undefined, options?: Intl.NumberFormatOptions) {
  return new Intl.NumberFormat(locales, options);
}

export function formatOrderSummary(i18n: I18n, date: Date, total: number) {
  const locales = i18n.locales ?? i18n.locale;
  const dateFormatter = memoizedGetDateFormatter(locales);
  // note - unlike memoizedGetDateFormatter, getNumberFormatter creates a new Intl instance on each call!
  // consider memoizing it if formatter options are stable
  const numberFormatter = getNumberFormatter(locales, {
    style: "currency",
    currency: "EUR",
  });

  return `${dateFormatter.format(date)} - ${numberFormatter.format(total)}`;
}

function useDateFormatter() {
  const { i18n } = useLingui();
  const locales = i18n.locales ?? i18n.locale;
  return memoizedGetDateFormatter(locales);
}
```

website/blog/2025-12-17-deprecating-intl-wrappers/index.md
```diff
@@ -0,0 +1,113 @@
...
+  const numberOptions = useMemo(
+    () => ({
+      style: "currency",
+      currency: "USD",
+      minimumFractionDigits: 2,
+    }),
+    []
+  );
+
+  const dateFormatter = useDateFormatter(dateOptions);
```
vonovak: if you have a hundred `PriceLine`s rendered, that's 

- 100 `dateOptions`
- 100 `numberOptions`
- 100 `Intl.NumberFormat`s
- 100 `Intl.DateTimeFormat`s

Where you could have 1 of each and it'd work the same - I know this may be just a contrived example, but it adds up and people tend to copy the examples.

website/blog/2025-12-17-deprecating-intl-wrappers/index.md
```diff
@@ -0,0 +1,113 @@
...
+}
+```
+
+## What about React hooks
+
+We do not plan to add `useDateFormat` or `useNumberFormat`. Shipping hooks on top of the wrappers would trigger requests for the same helpers outside React and across frameworks, putting us back into the role of wrapping all of `Intl`. Teams can keep local hooks or shared formatters without Lingui adapters.
+
+## FAQ
+
+- **What about performance?** Creating `Intl.*Format` is inexpensive, and you can keep formatters at module scope or memoize them. No special internal cache from Lingui is required.
```
vonovak:
```suggestion
- **What about performance?** Creating `Intl.*Format` is inexpensive, and you can memoize them. No special internal cache from Lingui is required.
```

creating at module scope isn't gonna work because then the formatter's locale is set, and cannot be changed

website/blog/2025-12-17-deprecating-intl-wrappers/index.md
```diff
@@ -0,0 +1,113 @@
...
+- **What about `PluralRules`?** Lingui already uses it inside message compilation, so another public formatter is unnecessary.
```
vonovak:
```suggestion
- **What about `Intl.PluralRules`?** Lingui already uses it inside message compilation, so another public formatter is unnecessary.
```

timofei-iatsenko:
I’ll be honest with you: this PR has been sitting around because the blog post feels very LLM-ish, and I don’t really want to merge something like that. I also don’t have enough energy to rewrite it myself, so I’ve been kind of dodging this PR.

Could you rewrite it in a more human way — how you would explain what changed and why — without going into too much detail? Ideally, it should be engaging to read and feel less like it was written by a robot.

yslpn:
> I’ll be honest with you: this PR has been sitting around because the blog post feels very LLM-ish, and I don’t really want to merge something like that. I also don’t have enough energy to rewrite it myself, so I’ve been kind of dodging this PR.
> 
> Could you rewrite it in a more human way — how you would explain what changed and why — without going into too much detail? Ideally, it should be engaging to read and feel less like it was written by a robot.

Yes, I used LLM when writing it. I'll mark the PR as a draft and rewrite it when I'm ready.

website/blog/2025-12-17-deprecating-intl-wrappers/index.md
```diff
@@ -0,0 +1,105 @@
...
+
+## What full wrapper parity would require
+
+To meet the expectations for a complete helper surface we would need wrappers and caching at least for `DateTimeFormat`, `NumberFormat`, `ListFormat`, `RelativeTimeFormat`, `DisplayNames`, `Collator`, `Segmenter`, `Locale`, `DurationFormat` (once it ships), plus `formatRange` and `formatToParts`. We are intentionally not going to maintain this matrix.
+
+## Plan
+
+- Now: mark the helpers as deprecated in code and docs without breaking existing apps.
+- Upcoming majors: keep them long enough to give teams time to migrate; removal will happen in one of the next majors (not necessarily the very next one).
+- Removal: no separate warning blog is planned - please track the release notes for major versions.
```
vonovak: We should think about offering a codemod, or asking community to contribute a codemod.

andrii-bodnar:
Hi @yslpn, thanks for the contribution!

I would suggest retargeting this PR to the `next` branch. Instead of a standalone blog post, we will include this information in the v6 migration guide and release notes. This type of change is a perfect fit for a major release, and keeping it bundled there helps avoid the overhead of a separate post.

yslpn:
> I would suggest retargeting this PR to the `next` branch. Instead of a standalone blog post, we will include this information in the v6 migration guide and release notes. This type of change is a perfect fit for a major release, and keeping it bundled there helps avoid the overhead of a separate post.

@andrii-bodnar Good idea. You can switch the target branch to the next one. There shouldn't be any conflicts.

Should I delete the blog post or do something else from my side?

website/blog/2025-12-17-deprecating-intl-wrappers/index.md
yslpn: We'll be accepting an i18 instance from the outside. I've added this to my examples.

There's no need to delve too deeply into premature optimization. Library users can decide if they need to.

I think creating formatting functions will be a simple operation and won't cause any performance issues.

website/blog/2025-12-17-deprecating-intl-wrappers/index.md
yslpn: I've extracted the options from React, but to keep the example simple, I suggest creating formatting functions for each component instance.

Adding unnecessary complexity can confuse users. They might think they have to optimize this, but I'm not so sure.

website/blog/2025-12-17-deprecating-intl-wrappers/index.md
yslpn: It seems like it would be difficult to make any kind of reliable codemod here.

andrii-bodnar:
@yslpn yes, I will reuse its content in #2424 and #2427 (perhaps in more compact and compressed form)

yslpn:
Hi @vonovak  Can I ask you to review this again please? I made some changes. The most significant ones are:

1) Deleted the blog post; @andrii-bodnar is now responsible for that for v6.

2) I changed the examples.

Was:
```ts
new Intl.NumberFormat(i18n.locales ?? i18n.locale).format(12345.678);
```
Now:
```ts
new Intl.NumberFormat(i18n.locale).format(12345.678);
```

The old version didn't make sense because NumberFormat doesn't throw an error and will always find some locale if the one passed is invalid. Main backup en-US

3) I didn't optimize the code too much to keep it simple for understanding the examples.
4) Moved most of the text from the blog to the PR description.

PR #2387, opened 2025-12-18
chore: new jsx transform

yslpn:
# Description

Switch React packages to the automatic JSX runtime: Babel/esbuild now emit JSX without explicit `React` imports, and the React tsconfig uses `react-jsx`. Raised the minimum supported `react` peer to `16.14.0` to align with the automatic runtime requirements. This follows the guidance in the new JSX transform announcement: https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

- No more import React — you don't need to import React in every file that uses JSX
- Smaller bundle size — slightly reduces compiled output

## Types of changes

- [x] Documentation update
- [ ] Examples update

Fixes # N/A

## Checklist

### DIFF
--- a/packages/react/src/TransNoContext.tsx
+++ b/packages/react/src/TransNoContext.tsx
@@ -70,7 +70,7 @@ export function TransNoContext(
   }
 
   const FallbackComponent: React.ComponentType<TransRenderProps> =
-    defaultComponent || RenderFragment
+    defaultComponent || RenderChildren
 
   const i18nProps: TransRenderProps = {
     id,
@@ -111,9 +111,8 @@ export function TransNoContext(
   return <Component {...i18nProps}>{translation}</Component>
 }
 
-const RenderFragment = ({ children }: TransRenderProps) => {
-  // cannot use React.Fragment directly because we're passing in props that it doesn't support
-  return <>{children}</>
+const RenderChildren = ({ children }: TransRenderProps) => {
+  return children
 }
 
 const getInterpolationValuesAndComponents = (props: TransProps) => {


packages/react/package.json
```diff
@@ -81,7 +81,7 @@
   "peerDependencies": {
     "@lingui/babel-plugin-lingui-macro": "5.7.0",
     "babel-plugin-macros": "2 || 3",
-    "react": "^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0"
```
vonovak, line 84: technically speaking, this is a breaking change. Nothing wrong with that. But I say.. let's bump this to 18 and stop caring about the old versions

packages/react/src/TransNoContext.tsx
```diff
@@ -54,7 +54,7 @@ export function TransNoContext(
 
   const { values, components } = getInterpolationValuesAndComponents(props)
 
-  const _translation: string =
```
vonovak, line 57: having this explicitly typed makes it more readable imho. With the type, you and the compiler know what it should be.

Can't easily say that about

```
const _translation =
    i18n && typeof i18n._ === "function"
      ? i18n._(id, values, { message, formats })
      : id
```

packages/react/src/TransNoContext.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { ComponentType, ReactNode } from "react"
+import { ComponentType, ReactNode } from "react"
```
vonovak:
```suggestion
import type { ComponentType, ReactNode, ReactElement } from "react"
```

packages/react/src/TransNoContext.tsx
```diff
@@ -13,14 +13,14 @@ export type TransRenderProps = {
 
 export type TransRenderCallbackOrComponent =
   | {
-      component?: undefined
+      component?: never
       render?:
         | ((props: TransRenderProps) => React.ReactElement<any, any>)
```
vonovak, line 15:
```suggestion
        | ((props: TransRenderProps) => ReactElement<any, any>)
```

same in other places, if you like

packages/react/src/TransNoContext.tsx
```diff
@@ -107,12 +108,12 @@ export function TransNoContext(
...
     component || FallbackComponent
 
-  return React.createElement(Component, i18nProps, translation)
+  return <Component {...i18nProps}>{translation}</Component>
 }
 
 const RenderFragment = ({ children }: TransRenderProps) => {
   // cannot use React.Fragment directly because we're passing in props that it doesn't support
-  return <React.Fragment>{children}</React.Fragment>
+  return <>{children}</>
```
vonovak: `RenderFragment` could return just `children` and not add any Fragment around it, it seems unnecessary

vonovak (changes requested):

packages/react/package.json
yslpn: I agree with you. If we upgraded React, we could use `useSyncExternalStore` to sync react state with i18n state. It's a simpler and more performant option than using `useEffect`.

But I need other contributors' agreement that we're ready to upgrade React to version 18.

packages/react/src/TransNoContext.tsx
yslpn: I prefer to use React.{type} directly without importing, as I think it's the most common approach in the React ecosystem.

packages/react/src/TransNoContext.tsx
yslpn: I would add this rule to the tsconfig config to require explicitly specifying type when importing and exporting

```json
{
  "compilerOptions": {
    "verbatimModuleSyntax": true
  }
}
```

 but this requires removing "module": "commonjs" from the config. This will be done later after https://github.com/lingui/js-lingui/issues/2363

packages/react/src/TransNoContext.tsx
yslpn: Why RenderFragment cannot be removed

```tsx
const RenderFragment = ({ children }: TransRenderProps) => {
  return <>{children}</>
}
```

`RenderFragment` is used as a fallback component in `TransNoContext`:

```tsx
const FallbackComponent = defaultComponent || RenderFragment
// ...
<FallbackComponent {...i18nProps}>{translation}</FallbackComponent>
```

Where `i18nProps` contains `{ id, message, translation, children }`.

It's safer to leave it for now.

packages/react/src/TransNoContext.tsx
yslpn: Okay, so for me, the extra types make it harder to read. But that's not important.

packages/react/package.json
yslpn: @andrii-bodnar @timofei-iatsenko 

What do you think about raising the React version to 18?

packages/react/src/TransNoContext.tsx
vonovak: > Why RenderFragment cannot be removed

I suggested not to remove `RenderFragment` but to have it return `children`:

```
const RenderFragment = ({ children }: TransRenderProps) => {
  return children
}
```

yslpn: Done! Renamed to `RenderChildren` and removed the unnecessary Fragment wrapper

packages/react/package.json
timofei-iatsenko: I'm not sure about rising the version to 18. There is no real need for that. I'm ok with rising react version to 16.14. 

I'm also thinking that despite technically it's a breaking change we can release it without a new major version.  It is not requiring react to be bumped to next major, so users may easy bump their projects to `16.14.0`

timofei-iatsenko (approved):

packages/react/package.json
yslpn: > we can release it without a new major version. 

agree with you

vonovak (approved):

timofei-iatsenko:
@andrii-bodnar please add that consumers need to bump their react dependency to at least `16.14.0` right in the github realeases description.

andrii-bodnar:
Yeah, I am going to mention this in the release description

PR #2401, opened 2026-01-16
feat(message-utils): separate browser/node exports for generateMessageId

### LINKED ISSUE #2381: Dual nodejs/browser version for @lingui/message-utils
Related https://github.com/lingui/js-lingui/pull/2320#issuecomment-3270259661

The calculation of the lingui id is using js only cryptography, to be able to use this in browser. Performance could be improved for the nodejs execution using nodejs built-in methods. 

We can create node-only and browser-only version of these packages using exports condition https://nodejs.org/api/packages.html#conditional-exports

yslpn:
# Description

Split `generateMessageId` into separate browser (default) and Node.js implementations:

- Browser: use `sha256.array()` directly instead of hex string conversion
- Node.js: use native `crypto.createHash().digest('base64')`

Node.js `crypto` is faster and doesn't require `js-sha256` dependency. Browser version optimized by avoiding hex parsing. Other runtimes (Deno, Bun, edge workers) support Node.js exports and should work out of the box.

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes #2381

## Checklist

### DIFF
--- a/packages/message-utils/package.json
+++ b/packages/message-utils/package.json
@@ -62,7 +62,7 @@
   ],
   "dependencies": {
     "@messageformat/parser": "^5.0.0",
-    "js-sha256": "^0.11.1"
+    "js-sha256": "^0.10.1"
   },
   "devDependencies": {
     "@lingui/jest-mocks": "workspace:^",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3199,7 +3199,7 @@ __metadata:
     "@lingui/jest-mocks": "workspace:^"
     "@messageformat/date-skeleton": ^1.1.0
     "@messageformat/parser": ^5.0.0
-    js-sha256: ^0.11.1
+    js-sha256: ^0.10.1
     unbuild: 2.0.0
   languageName: unknown
   linkType: soft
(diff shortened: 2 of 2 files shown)


packages/message-utils/package.json
```diff
@@ -50,7 +62,7 @@
   ],
   "dependencies": {
     "@messageformat/parser": "^5.0.0",
-    "js-sha256": "^0.10.1"
+    "js-sha256": "^0.11.1"
```
timofei-iatsenko: Revert this please, we don't need to narrow the constraint unneccesarrily if our implementation works with both versions. 

The consumer can bump to the latest if they need since our constraint is not restricting that.

timofei-iatsenko (commented):
LGTM, except of the bumped dep. Thanks for the contribution!

timofei-iatsenko:
@Zxilly fyi

packages/message-utils/package.json
yslpn: Okay. Reverted

yslpn:
By the way, what is the use case for generateMessageId in a browser? I couldn't find any information.

Zxilly:
My initial idea was to create a faster `generateMessageId`, but after implementing multi thread, the performance was generally satisfactory to me. What might be more important to me is a customizable `generateMessageId`, but I haven't come up with a good approach if using the swc plugin.

timofei-iatsenko:
@Zxilly what in particulare you would like to customize in it?  Maybe that could be solved somehow else.

@yslpn it is not used directly in the lingui code, but used by the external integrations. 

One of the potential use-cases - when you download raw, non-compiled catalogs from the TMS (OTA delivery) and then you need to prepare them to be usable in app.

Another use-case which i am aware of, Vue integration which doesn't use compile time macro which is replacing source message with an id, but instead it generate message id at runtime (in a browser).

PR #2412, opened 2026-01-26
refactor!: strictNullChecks for cli lib packages

timofei-iatsenko:
# Description

This PR enables typescript's `strictNullChecks` for 

- `@lingui/cli`
- `@lingui/format-po`
- `@lingui/loader`
- `@lingui/message-utils`
- `@lingui/vite-plugin`
- `@lingui/extractor-vue`
- `@lingui/metro-transformer`

Packages still not using `strictNullChecks`:
- `format-po-gettext` this package doesn't even have `strict: true` at this point, so it requires a heavy modernizing (in a separate PR)
- `babel-plugin-lingui-macro` - many patterns in the code hard to rewrite, currently benefits are not worth the effort, maybe re-attempt next time

The most importantly that all "lib" packages such as `@lingui/cli` and `@lingui/conf` now fully strict, that means rewriting "leafs" from dependency graph packages is unblocked. 

## null vs undefined

Many of the changes in the PR caused by null vs undefined change. There is a lot of debates in typescript and javascript community about when and what should be used. This become even more problematic with `strictNullChecks` because

type `string | null | undefined` is not assignable to `string | undefined`  or vice versa `string | null`

So it's better to eliminate one or another and write consisently across codebase. 

Many teams opted for the `undefined` only, for example typescript in theirs codebase uses it. The main reason for that is because `undefined` could not be complietely eliminated - it would pops up when property or parameter is optional, but null could be fully eliminated from the code base if you are not assigning it manually. 

I also opted for `undefined` only, hence changes. 

## `ExtractedMessageType` vs `MessageType` and their corresponding `ExtractedCatalogType` and `CatalogType`

Now disticntion between them even more "distinct" and change the real code scenario `Extracted*` family used for messages produced by extractor, where versions without `Extracted` are corresponding for messages and catalogs from loaded catalogs. 

What the diffrence? For example just extracted messages would not have a `translation` property, `obsolete` and `extra`, they are parts of the locally stored catalog.  

And vice versa, `ExtractedMessageType` **always** have the `placeholders` and `comments` fields, even if they are empty, but catalog from disk could omit them. 

So to further harden the typesafety this distiction was improved. 

## Breaking changes description

It's not intentionally breaking change, but changing the types might break something for anyone who used lingui internal types (custom integrations) 

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/conf/src/__snapshots__/index.test.ts.snap
+++ b/packages/conf/src/__snapshots__/index.test.ts.snap
@@ -64,10 +64,6 @@ exports[`@lingui/conf > should return default config 1`] = `
       "useLingui",
     ],
   },
-  "service": {
-    "apiKey": "",
-    "name": "",
-  },
   "sourceLocale": "",
 }
 `;


packages/cli/src/api/catalog/getCatalogs.ts
```diff
@@ -91,8 +91,17 @@ export async function getCatalogs(
...
-    catalogs.push(...(await getExperimentalCatalogs(config)))
+  if (
+    config.experimental?.extractor &&
+    config.experimental.extractor.entries.length
+  ) {
+    catalogs.push(
+      ...(await getExperimentalCatalogs(
+        config,
+        format,
+        config.experimental.extractor
```
andrii-bodnar, line 102: If we already pass the full config, why would we need this new parameter?

timofei-iatsenko: `config.experimental.extractor` is optional so it's type is nullable, few lines above we checking that it is set with `if (config.experimental.extractor) {}`. Typescript could infer that and narrow type of this expression so it's non nullable anymore. 

If we would just pass the whole config to `getExperimentalCatalogs(` this type narrowing information would be lost and we would have to do either additional type casting, such as `config.experimental.extractor!` which is basically just shutting up the typescript, or properly check it again that is set with `if () {}` but in this case we should handle the branches of that if somehow. 

So the new signature of this function is basically saying, `config.experimental.extractor` should be already set to call this function and forcing a user to check it is set BEFORE this function is called.

timofei-iatsenko:
> I noticed a frequent use of the Non-Null Assertion Operator `!`. While this could be ok for Unit tests where you are intentionally passing specific data and want to keep the test code concise, in the library code we risk to have runtime `TypeError: Cannot read property...` by silencing TypeScript

this is a consequence for extra type safety enabled by the `"noUncheckedIndexedAccess": true,`.

In most of cases this operator used with `catalogs[locale]` or `catalog[message]` kind of expressions. In this cases we should check that it is not null, but we actually already know that because we just created it few lines above. So it's the case where we can assert that it is not null. 

Also I didn't want to introduce any functional changes, and additional `if`s and conditional checks is a functional change. 

Also covering the existing code with highest level of typesafety is not easy, because the code itself should be written in a way typescript could follow and infer the types.

packages/cli/src/api/catalog/getCatalogs.ts
andrii-bodnar: Thanks for the explanation!

PR #2414, opened 2026-01-26
feat(vue-extractor): add reactivity transform support for props destructuring

mister-what:
# Description

This PR adds support for Vue's Reactivity Transform in the `vue-extractor` to fix message ID mismatches when using destructured props from `defineProps()`.

## Problem

When using destructured props with Lingui macros like `plural()` or `select()`, the extractor and the runtime macro see different code representations:

```vue
<script setup>
const { selectedCount } = defineProps({ selectedCount: Number });

const text = plural(selectedCount, {
  one: "# item",
  other: "# items",
});
</script>
```

Extractor sees: `plural(selectedCount, ...)` and generates message ID for `{selectedCount, plural, ...}`
Runtime macro sees: `plural(__props.selectedCount, ...)` and generates message ID for `{0, plural, ...}`

This mismatch causes the translations to fail at runtime.

### Solution

The extractor now runs Vue’s `compileScript()` when the `reactivityTransform` option is enabled.
This applies the same reactivity transform that Vue uses internally, so both the extractor and the runtime macro operate on identical code (i.e. `__props.x` instead of destructured bindings). As a result, message IDs are generated consistently and no longer mismatch at runtime.

The feature is **opt-in** (`reactivityTransform: false` by default) to avoid breaking existing setups that rely on the previous behavior.

To support this configuration cleanly, the extractor API was changed to a **factory-based approach**. A preconfigured extractor instance is still exported for backwards compatibility, but it is now marked as deprecated in favor of `createVueExtractor()`.

### Usage

```ts
// lingui.config.js
import { createVueExtractor } from "@lingui/extractor-vue";

export default defineConfig({
  // ... other config
  extractors: [createVueExtractor({ reactivityTransform: true })],
});
```

## Types of changes

- [x] Documentation update
- [x] Examples update

## Checklist

- [x] I have read the https://github.com/lingui/js-lingui/blob/main/CONTRIBUTING.md and https://github.com/lingui/js-lingui/blob/main/CODE_OF_CONDUCT.md docs

### DIFF
--- a/packages/extractor-vue/src/extractor.test.ts
+++ b/packages/extractor-vue/src/extractor.test.ts
@@ -33,7 +33,7 @@ describe("vue extractor", () => {
       flow: false,
     },
   })
-  const vueExtractor = createVueExtractor();
+  const vueExtractor = createVueExtractor()
 
   it("should ignore non vue files in extractor", async () => {
     const match = vueExtractor.match("test.js")
@@ -90,7 +90,7 @@ describe("vue extractor", () => {
     const code = fs.readFileSync(filePath, "utf-8")
 
     let messages: ExtractedMessage[] = []
-    const vueExtractor = createVueExtractor({reactivityTransform: true})
+    const vueExtractor = createVueExtractor({ reactivityTransform: true })
     await vueExtractor.extract(
       "props-destructuring.vue",
       code,


timofei-iatsenko:
Hey, you also can mitigate that by using a named placeholder, which is going to be recommended way when we finish with V6 release

```ts
const { selectedCount } = defineProps({ selectedCount: Number });

// {selectedCount} shortcut to {selectedCount: selectedCount}, which is {label: value}
const text = plural({selectedCount}, {
  one: "# item",
  other: "# items",
});
```

timofei-iatsenko:
@mister-what we currently have a code-freeze for the `main` branch and development goes in the `next`, your contribution is valuable and we happily accept it, please change the target to the `next` and resolve conflicts if any. 

This fix would be included in upcoming v6 release.

mister-what:
@timofei-iatsenko alright 👍 I changed the target branch to `next` and resolved all conflicts.

mister-what:
I fixed the failing snapshot tests and annotated some type-only imports.

andrii-bodnar:
Thanks for the contribution! Is this something only relevant to the experimental (dependency tree) extractor and not the regular extractor?

It introduces the framework-specific configuration in the `conf` package. What about making it a configuration of the Vue Extractor? Something like:

```js
// lingui.config.js
import { createVueExtractor } from '@lingui/extractor-vue'

export default {
  extractors: [
    createVueExtractor({
      reactivityTransform: false,
      // ...
    })
  ]
}
```

Although it's more complicated, it's more future-proof as we can easily add more options if needed. @mister-what @timofei-iatsenko what do you think?

timofei-iatsenko:
This would be the best option - but a breaking change for existing users.  If that would be implemented for the vue extractor the same should be done for babel extractor for consistency. 

So instead of object both vue and babel would export a factory function from the package similarly to formatters.

mister-what:
I could make it non breaking by exposing the factory and exposing the configured extractor. The latter could then be deprecated.

packages/extractor-vue/src/vue-extractor.ts
```diff
@@ -67,4 +76,9 @@ export const vueExtractor: ExtractorType = {
         ),
     )
   },
-}
+})
+
+/**
+ * @deprecated use createVueExtractor instead
+ */
```
timofei-iatsenko, line 83:
```suggestion
/**
 * @deprecated use {@link createVueExtractor} instead
 */
```

andrii-bodnar:
@mister-what please update the PR description for consistency. Thanks!

mister-what:
@andrii-bodnar description is updated 👍

PR #2418, opened 2026-01-27
chore: update prettier v3 import to use named format function

yslpn:
# Description

Fix Prettier v3 import in macro tests. Named import `{ format }` works correctly with Prettier's async API in Vitest environment.

## Types of changes

## Checklist


timofei-iatsenko:
what the difference?

yslpn:
> what the difference?

I misunderstood why the tests were failing locally

timofei-iatsenko:
Yeah, you probably fallen into a trap with non build packages before running the tests. After ESM migration packages consumed from artifacts in the monorepo, not from the source. So you need to build dependencies every time you run the tests. Annoying, but not a priority for now

PR #2420, opened 2026-01-28
chore(cli): replace threads with tinypool

yslpn:
# Description

Replace `threads` library with `tinypool` for worker pool management in CLI package.

**Motivation:**
- `tinypool` is a more actively maintained and lightweight alternative
- Eliminates the need for a custom yarn patch for `threads`
- New Promise-based usage with `Promise.all` instead of `pool.queue`/`pool.completed`

**Key changes:**
- Created `typedPool.ts` - generic typed wrapper around Tinypool
- Created `workerPools.ts` - factory functions for all worker pools (extract, extract-experimental, compile)
- Updated all worker wrappers to export default function instead of using `expose()` from threads
- Simplified worker invocation: `pool.run([...args])` instead of `pool.queue(worker => worker(...))`
- Removed yarn patch for threads library

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/cli/src/api/catalog/extractFromFiles.ts
+++ b/packages/cli/src/api/catalog/extractFromFiles.ts
@@ -127,7 +127,7 @@ export async function extractFromFilesWithWorkerPool(
 
   await Promise.all(
     paths.map(async (filename) => {
-      const result = await workerPool.run([filename, resolvedConfigPath])
+      const result = await workerPool.run(filename, resolvedConfigPath)
 
       if (!result.success) {
         catalogSuccess = false
--- a/packages/cli/src/api/typedPool.ts
+++ b/packages/cli/src/api/typedPool.ts
@@ -1,7 +1,7 @@
 import Tinypool, { type Options } from "tinypool"
 
-export interface TypedPool<TArgs extends unknown[], TResult> {
-  run(args: TArgs): Promise<TResult>
+export type TypedPool<TArgs extends unknown[], TResult> = {
+  run(...args: TArgs): Promise<TResult>
   destroy(): Promise<void>
 }
 
@@ -11,7 +11,7 @@ const createTypedPool = <TArgs extends unknown[], TResult>(
   const pool = new Tinypool(options)
 
   return {
-    run: (args: TArgs) => pool.run(args) as Promise<TResult>,
+    run: (...args: TArgs) => pool.run(args) as Promise<TResult>,
     destroy: () => pool.destroy(),
   }
 }
--- a/packages/cli/src/api/workerPools.ts
+++ b/packages/cli/src/api/workerPools.ts
@@ -5,7 +5,7 @@ import type { CompileWorkerFunction } from "../workers/compileWorker.js"
 
 export type ExtractWorkerPool = WorkerPool<ExtractWorkerFunction>
 
-interface PoolOptions {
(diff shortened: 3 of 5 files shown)


timofei-iatsenko:
I pushed some slight changes in types, just wanted to have call to `pool.run(...args)` look more like a method execution with argument instead passing args as array/tuple. 

Also i'm wondering about:

```ts
minThreads: poolSize,
maxThreads: poolSize,
```

Probably we should set only a maxThreads, so if there would be only 2 tasks, and  poolSize = 5, it will not spawn 3 extra workers. 

It's better to consult actually what their docs saying about best practicies and maybe check vitest code how they works with those parameters.

yslpn:
> I pushed some slight changes in types, just wanted to have call to `pool.run(...args)` look more like a method execution with argument instead passing args as array/tuple.

Thanks

> Also i'm wondering about:
> 
> ```ts
> minThreads: poolSize,
> maxThreads: poolSize,
> ```
> 
> Probably we should set only a maxThreads, so if there would be only 2 tasks, and poolSize = 5, it will not spawn 3 extra workers.
> 
> It's better to consult actually what their docs saying about best practicies and maybe check vitest code how they works with those parameters.

I found that vitest was removed from vitest and they wrote their own implementation https://vitest.dev/guide/migration.html#pool-rework

I looked at how facebook/docusaurus uses it; they also use the same values ​​for maxThreads and minThreads. They calculate the ideal number for themselves based on pages:
https://github.com/facebook/docusaurus/blob/df259ddb269153e652f3731ab5cd52d12244b45f/packages/docusaurus/src/ssg/ssgExecutor.ts#L126

50 pages → 1 thread
150 pages → 2 threads
and so on.

In our case, we have two options:
ref: https://lingui.dev/ref/cli#extract-workers

1) We calculate the number of threads ourselves, before starting, using --workers 
To avoid breaking the cli API and creating new flags, we can leave the same number of threads specified with --workers, and we guarantee this number of threads.
```
minThreads: poolSize,
maxThreads: poolSize,
```

2) Or we calculate automatically.
Use our current logic for calculating the number of threads and pass this value to maxThreads.

```
minThreads: 0,
maxThreads: poolSize,
```

Plus, we need to adjust all the messages, as there will be two cases:

`Use worker pool of size ${poolSize}` and `Use worker pool with up to ${poolSize} workers`

Is that okay?

timofei-iatsenko:
It seems from the docs (from upstream piscina):

> minThreads: (number) Sets the minimum number of threads that are always running for this thread pool. The default is the number provided by [os.availableParallelism](https://nodejs.org/api/os.html#osavailableparallelism).
> maxThreads: (number) Sets the maximum number of threads that are running for this thread pool. The default is the number provided by [os.availableParallelism](https://nodejs.org/api/os.html#osavailableparallelism) * 1.5.

That `minThreads` is not based on the number of tasks.  So i'm confused now how that works.  Reading this docs seems your's 

```
minThreads: poolSize,
maxThreads: poolSize,
```

totally makes sense

timofei-iatsenko:
@andrii-bodnar regarding coverage - disregard this report, since workers are not instrumented by covergae because they run in kinda separate process. The functionality is covered in the original implementation - this change doesn't add anything to logic we already have.

PR #2424, opened 2026-01-29
docs: v6 migration guide

andrii-bodnar:
# Description

#2363 

## Types of changes

- [x] Documentation update
- [ ] Examples update

## Checklist

### DIFF
--- a/website/docs/releases/migration-6.md
+++ b/website/docs/releases/migration-6.md
@@ -4,9 +4,7 @@ This guide will help you migrate from Lingui 5.x to 6.x. It covers the most impo
 
 Need to upgrade an older project to v5 first? See our [older migration guide](/releases/migration-5).
 
-<!-- TODO: create dedicated deployment for 5.x documentation -->
-
-If you're looking for 5.x documentation, you can find it [here](https://lingui.dev).
+If you're looking for 5.x documentation, it will continue to be available at [lingui.dev](https://lingui.dev) until v6 is officially released.
 
 ## Node.js Version
 
@@ -32,7 +30,7 @@ All Lingui packages have been converted to ESM-only distribution, except `@lingu
 
 ### Migration
 
-For most users, no changes are required. Modern bundlers (Vite, Webpack, esbuild, Rollup) and Node.js versions that support `require()` for ESM modules (20.16+, 22.19+, or 24+) handle ESM imports transparently.
+For most users, no changes are required. Modern bundlers (Vite, Webpack, esbuild, Rollup) and Node.js versions that support `require()` for ESM modules (22.19+, or 24+) handle ESM imports transparently.
 
 ## Deprecated `format` String and `formatOptions` Removed
 


website/docs/releases/migration-6.md
```diff
@@ -0,0 +1,168 @@
...
+  format: formatter({ lineNumbers: true }),
+});
+```
+
+If you were using a different format, import the corresponding formatter:
+
+- **PO format**: `import { formatter } from "@lingui/format-po"`
+- **PO Gettext format**: `import { formatter } from "@lingui/format-po-gettext"`
+- **JSON format**: `import { formatter } from "@lingui/format-json"`
+- **CSV format**: `import { formatter } from "@lingui/format-csv"`
```
timofei-iatsenko, line 73: I would change this in the way: 

- if you have in your configuration `` format: `po` `` and no `formatOptions` - just drop this line entirely. Lingui will defaults to `po` if no format specified. 

If you use different formatter or use `formatOptions`, you need to import formatter from the package and pass options into the function.

website/docs/releases/migration-6.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Migration
+
+For most users, **no changes are required**. Modern bundlers (Vite, webpack 5, esbuild, Rollup) and Node.js 20+ handle ESM imports transparently.
+
+If you encounter issues with ESM imports in a non-standard setup:
+
+1. Ensure your bundler is configured to handle ESM modules
+2. If using Node.js directly, ensure you're on Node.js 20+ or update your `package.json` to include `"type": "module"`
+3. Update any `require()` calls to use `import` syntax where applicable
```
timofei-iatsenko: I would suggest to not go into much details here, coz  it's implementation details. i would formulate it like this: 

Lingui 6.0 is distributed as ESM only.  ESM-only requires a modern Node that supports `ESM require` (`20.16+`,` 22.19+`, or `24+`).


The part about bundlers require and import and type: module are totally incorrect, with  `ESM require`  feature this not needed, and bundlers never cared about this.

website/docs/releases/migration-6.md
```diff
@@ -0,0 +1,168 @@
...
+
+This guide will help you migrate from Lingui 5.x to 6.x. It covers the most important changes and breaking changes.
+
+Need to upgrade an older project to v5 first? See our [older migration guide](/releases/migration-5).
+
+If you're looking for 5.x documentation, you can find it [here](https://lingui.dev).
+
+## Node.js Version
+
+The minimum supported version of Node.js in Lingui v6 is v22.
```
timofei-iatsenko: Maybe i forgot, but did we agreed to `20.16+`,` 22.19+`, or `24+` ?

andrii-bodnar: We haven't agreed yet, but v20 LTS ends in April and v22 will be maintained until April 2027. Therefore, it feels reasonable to upgrade to the next LTS version with each major Lingui release.

timofei-iatsenko: Fair enough, but it should be not just 22, but 22.19+, because ESM require was ported retrospectively from v24 to older versions.

andrii-bodnar: Agree, it should be specified exactly as `22.19+`

andrii-bodnar:
Merging this PR as it is to unblock the v6 pre-release. If you have any comments, please leave them here and I will polish this guide prior to the official v6 release.

PR #2426, opened 2026-01-29
chore: use workspace protocol, reconsider peerDeps

timofei-iatsenko:
# Description

- use workspace protocol to link packages in the monorepo. It's more conviniet to work, since version is not changing after every release and not produce diffs and merge conflicts. Also that ensures versions are synchronized. I tested and modern lerna understand workspace protocol without problems. 
- I moved `@lingui/babel-plugin-lingui-macro` to the direct deps of the core and react packages. I considered all pros and cons, and decided that it has more pros to just install and forget. (that was added as optional peer because user might use swc plugin and don't need this babel, but it seems that drawbacks of bumping peerDependencies on every release is bigger than installation size of the package). 
- i removed `@babel/runtime` from direct dependencies of many packages - this was a leftover from previous build pipeline. 

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/babel-plugin-lingui-macro/package.json
+++ b/packages/babel-plugin-lingui-macro/package.json
@@ -57,7 +57,6 @@
     "@babel/core": "^7.20.12",
     "@babel/types": "^7.20.7",
     "@lingui/conf": "workspace:*",
-    "@lingui/core": "workspace:*",
     "@lingui/message-utils": "workspace:*"
   },
   "peerDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -3218,7 +3218,6 @@ __metadata:
     "@babel/traverse": ^7.20.12
     "@babel/types": ^7.20.7
     "@lingui/conf": "workspace:*"
-    "@lingui/core": "workspace:*"
     "@lingui/message-utils": "workspace:*"
     "@types/babel-plugin-macros": ^2.8.5
     babel-plugin-react-compiler: ^1.0.0


packages/babel-plugin-lingui-macro/package.json
```diff
@@ -55,11 +55,9 @@
   },
   "dependencies": {
     "@babel/core": "^7.20.12",
-    "@babel/runtime": "^7.20.13",
     "@babel/types": "^7.20.7",
-    "@lingui/conf": "5.9.0",
-    "@lingui/core": "5.9.0",
```
andrii-bodnar, line 61: This is not needed?

andrii-bodnar:
Looks good to me. I had to look into how it will be handled during the release process. I'm just leaving this information here.

Since Lerna 6+, when you run `lerna publish`, Lerna delegates to Yarn (because of `"npmClient": "yarn"` in the `lerna.json`), and Yarn handles the `workspace:` protocol transformation according to its documented behavior:

| Your `package.json` | Published `package.json` |
|---------------------|--------------------------|
| `"workspace:*"` | `"=5.9.0"` |
| `"workspace:^"` | `"^5.9.0"` |
| `"workspace:~"` | `"~5.9.0"` |

https://yarnpkg.com/features/workspaces#cross-references

Using `workspace:*` will give:

```json
"@lingui/conf": "=5.9.0"
```

Which is the same as the `5.9.0` value.

packages/babel-plugin-lingui-macro/package.json
timofei-iatsenko: Yes `"@lingui/core"` not used directly in the macro code, and caused a cycle dependency error during build.  I'm not sure why i added it previously but technically it's not needed.

Issue #2411, opened 2026-01-24
Proposal: Delete the create-react-app example from code and docs


yslpn:
On February 14, 2025, the React team officially announced the deprecation of Create React App (CRA). However, create-react-app has been abandoned for many years.

I propose removing the CRA example from the lingui repository and documentation. Maintaining a deprecated example can confuse users and create the impression that CRA is still a good valid way to build React applications.

Ref: https://react.dev/blog/2025/02/14/sunsetting-create-react-app

PR #2430, opened 2026-02-01
chore: explicit "type" field in package.json and alignment for "exports"

yslpn:
# Description

ESM migration: added `"type": "module"` to root and several packages. Removed unused Jest files, replaced `strip-ansi` with `node:util`.

Added explicit  `"type": "commonjs"` to `babel-plugin-lingui-macro`, `metro-transformer` and `remote-loader`.  Publint recommends explicitly specifying the type, for example https://publint.dev/@lingui/react@5.9.0

Note: `website/` without ESM due to Docusaurus limitation https://github.com/facebook/docusaurus/issues/6520

Removed legacy workarounds for older React Native versions that didn't support the `"exports"` field. Lingui supports React Native >= 0.73 (`@lingui/metro-transformer` peerDependencies). React Native 0.72+ supports `"exports"`

## Types of changes

- [ ] Documentation update
- [ ] Examples update

## Checklist


packages/conf/__typetests__/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "compilerOptions": {
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "moduleResolution": "nodenext"
```
timofei-iatsenko: i think this file is not needed at all, it was added as a workaround previously, now when everethyting is strict and not `@ts-ignores` in the code it should be not needed (will follow package's root tsconfig)

packages/conf/package.json
```diff
@@ -30,7 +28,10 @@
     "check-types": "tsc --noEmit"
   },
   "exports": {
-    ".": "./dist/index.mjs"
+    ".": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.mjs"
+    }
```
timofei-iatsenko: Basically you don't need to specify a `types` condition for typescript. It will infer typings perfectly fine automatically.

In all places where i changed recently i did not use that, just for simplicity.  So for consistency please simplify it to just: 

```json
".": "./dist/index.mjs"
```

here in all other places

packages/conf/__typetests__/tsconfig.json
yslpn: Yes, it works without it. Thank you.

packages/conf/package.json
yslpn: I've been reading the documentation, and it's true. Thanks

ref
https://www.typescriptlang.org/docs/handbook/modules/reference.html#packagejson-exports

yslpn: If you don't specify types in exports, Ubuild won't generate them. I'll look at it in more detail later.

timofei-iatsenko: Yes, i explicitly ask it to generate types with: 

```
  "unbuild": {
    "declaration": "node16"
  }
```

yslpn: I was thinking about creating the config as a separate file in each package. It has type checking, and you can go to the source code and read the comments in the ts types. Plus, this adds consistency, as currently in some places there are separate configs (React), while in others you need to go to package json.

timofei-iatsenko: don't spend too much time on this, it's likely possible that build pipeline would be replaced to nx in the near future.

yslpn: Okay, good

timofei-iatsenko:
@yslpn i fixed verdaccio script. but tests in the CRA is failing, better to know why that happend because the error looks weird to me.

yslpn:
> @yslpn i fixed verdaccio script. but tests in the CRA is failing, better to know why that happend because the error looks weird to me.

I removed the old files needed by Jest 27
https://github.com/lingui/js-lingui/pull/2430/changes#diff-7da44f92d02898353f3d20536ca010c9f91f38b4b5df6e8d1d23fb9fd910f674

but I didn't notice that CRA uses Jest.

Do we support Jest 27? If so, I'll revert the removed files. If not, I'll add moduleNameMapper to CRA's package.json.

I previously proposed removing CRA from example maps. https://github.com/lingui/js-lingui/issues/2411 We can do that later.

yslpn:
> chore: remove CRA, write tests for vite-babel example, correct macro exports #2432

Thank you

PR #2433, opened 2026-02-04
refactor: proper macro browser shim

timofei-iatsenko:
# Description

Previously, the Vite plugin contained hacks to prevent macro implementation code from being pulled into the bundle. These hacks were also used to detect and warn when a macro was used incorrectly (eq. when the macro plugin was not configured).

In this PR, I replaced that approach with a more robust, bundler-agnostic solution for detecting incorrect macro usage. Instead of performing this check inside the Vite plugin, the logic is now implemented using package exports conditions:

- A special condition that resolves to a browser shim
- A default condition that resolves to the code required by the babel macro plugin

This allows us to surface a clear warning when macros are used without proper setup, without relying on bundler-specific behavior.

As a result, the solution should work consistently across modern bundlers (Vite, Webpack, Rspack, esbuild, etc.), rather than being limited to Vite only.

## Sidechange

I also found few more places where absolute version left instead of `workspace:*` protocol. 

## ESM Migration

This change deleted the last place in the `babel-plugin-lingui-macro` used CJS syntax, and i was able to switch package to ESM only.

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/react/macro/index.d.cts
+++ b/packages/react/macro/index.d.cts
@@ -107,11 +107,6 @@ export const SelectOrdinal: (props: PluralChoiceProps) => ReactNode
  */
 export const Select: (props: SelectChoiceProps) => ReactNode
 
-type TFunction = {
-  (descriptor: MacroMessageDescriptor): string
-  (literals: TemplateStringsArray, ...placeholders: any[]): string
-}
-
 /**
  *
  * Macro version of useLingui replaces _ function with `t` macro function which is bound to i18n passed from React.Context
@@ -136,5 +131,8 @@ type TFunction = {
  * ```
  */
 export function useLingui(): Omit<I18nContext, "_"> & {
-  t: TFunction
+  t: {
+    (descriptor: MacroMessageDescriptor): string
+    (literals: TemplateStringsArray, ...placeholders: any[]): string
+  }
 }


packages/core/package.json
```diff
@@ -36,13 +36,18 @@
   },
   "exports": {
     ".": "./dist/index.mjs",
-    "./macro": "./macro/index.cjs",
+    "./macro": {
+      "browser": "./macro/browser.cjs",
+      "types": "./macro/index.d.cts",
+      "default": "./macro/index.cjs"
```
timofei-iatsenko, line 42: since there are two diffrent implementations under conditions we have to sepcify `type` explicitly, otherwise typescript is confused

packages/babel-plugin-lingui-macro/package.json
```diff
@@ -59,18 +59,9 @@
...
     "@lingui/message-utils": "workspace:*"
   },
-  "peerDependencies": {
-    "babel-plugin-macros": "2 || 3"
-  },
-  "peerDependenciesMeta": {
-    "babel-plugin-macros": {
-      "optional": true
-    }
-  },
```
timofei-iatsenko, line 69: i moved macro creation from this package to upstream core/react, they also defining the `babel-plugin-macros` as peer

packages/react/macro/index.d.cts
```diff
@@ -137,5 +131,8 @@ declare function _t(
  * ```
  */
 export function useLingui(): Omit<I18nContext, "_"> & {
-  t: typeof _t
+  t: {
+    (descriptor: MacroMessageDescriptor): string
+    (literals: TemplateStringsArray, ...placeholders: any[]): string
+  }
```
timofei-iatsenko, line 137: I found that `_t` function type declaration was public (available for import from the package) which is not correct as it's a helper type for useLingui, i changed the definition so it's no longer exported

Issue #2207, opened 2025-03-21
Generate Lingui ID with URL and filename safe alphabet


jimmymcpeter:
### Problem Description

My translation system's key names support pretty much anything.  I'm able to import format-json files without issue.  One challenge I've encountered however relates to the generated Lingui IDs using the standard base64 encoding alphabet-- https://github.com/lingui/js-lingui/blob/b350ea07ce53d26894e49da67bd046f75adba44b/packages/message-utils/src/generateMessageId.ts#L5-L7

The 62nd and 63rd characters in the base 64 alphabet are `+` and `/` and a lot of my generated IDs contain these -- https://datatracker.ietf.org/doc/html/rfc4648#section-4

If my translations need to subsequently get sent to a different vendor/translation system, it can fail if XLIFF v2.0 is used.  The `id` element type in the [XLIFF 2.0 schema](https://docs.oasis-open.org/xliff/xliff-core/v2.0/os/schemas/xliff_core_2.0.xsd) is `xs:NMTOKEN`.  This means the value may only consist of letters, digits, periods, hyphens, underscores, and colons.  No whitespace is allowed.

### Proposed Solution

It would be nice if an option was added to generate Lingui ID's using the URL and filename safe Base 64 alphabet. 
 The 62nd and 63rd characters would instead be `-` and `_` -- https://datatracker.ietf.org/doc/html/rfc4648#section-5

Below is an example overriding the current generateMessageId function -- 

```ts
import crypto from 'node:crypto';

const UNIT_SEPARATOR = "\u001F"

export function generateMessageId(msg: string, context = "") {
  const result = crypto.createHash('sha256');

  result.write(msg + UNIT_SEPARATOR + (context || ""));

  result.end();

  return result.digest('base64url').slice(0, 6);
}
```

### Alternatives Considered

A custom formatter is necessary to work around this.

### Additional Context

Kudos on such a great library. I've worked on quite a few i18n/l10n projects, and it's been really refreshing to work on one that uses Lingui.

andrii-bodnar:
Hello, and thank you for reaching out! I recall a similar request—#1789. Unfortunately, this change would be a breaking one and would require a major version update.

jimmymcpeter:
Hi, wouldn't adding it as a new [config](https://lingui.dev/ref/conf) for this only be a minor version?

andrii-bodnar:
[...]

PR #2438, opened 2026-02-08
feat!: use base64 urlsafe alphabet for id

timofei-iatsenko:
# Description

This PR adresses this issue: https://github.com/lingui/js-lingui/issues/2207
SWC PR: https://github.com/lingui/swc-plugin/pull/194

This PR brings url-safe base64 decoding for ids. 

Note for migration:

- If Po format and bundlers loaders is used - no manual migration is required. 
- If `lingui compile` command used - You need to recompile your catalogs after updating.
- If some of the formats that explicitly store Lingui IDs is used - A manual migration is required. You’ll need to update the IDs.

The URL-safe variant differs from standard Base64 only by the following transformations:

```js
    .replace(/\+/g, "-")
    .replace(/\//g, "_")
    .replace(/=+$/, "")
```

In practice, this means you can migrate existing IDs with a simple search-and-replace in your text editor.

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/babel-plugin-extract-messages/test/__snapshots__/extract-messages.test.ts.snap
+++ b/packages/babel-plugin-extract-messages/test/__snapshots__/extract-messages.test.ts.snap
@@ -191,7 +191,7 @@ exports[`@lingui/babel-plugin-extract-messages > should extract Plural messages 
   {
     "comment": undefined,
     "context": "Some context",
-    "id": "8qNz+K",
+    "id": "8qNz-K",
     "message": "{count, plural, one {# book} other {# books}}",
     "origin": [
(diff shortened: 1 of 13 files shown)


yslpn:
Thanks for taking this on.

> If Po format and bundlers loaders is used - no manual migration is required.

Could this cause migration issues in the `isGeneratedId` checks used by the `translationIO` services and `format-po`? I’m concerned that legacy IDs might be incorrectly treated as explicit IDs.

timofei-iatsenko:
Should not, because they doesn't store lingui id anywhere, they are produce them from the message every time.

PR #2445, opened 2026-02-10
fix(cli): avoid extract-experimental hang with workers when no outputs

Adonis0123:
## Summary

  path), so this PR only adds a regression test for the no-entry + worker-
  pool scenario.

  ## Changes
  worker pool`

  ## Scope
  - No public API changes
  - No type/interface changes
  - Test coverage enhancement only

  ## Target release
  - `next` branch (upcoming v6)

### DIFF
--- a/packages/cli/test/extractor-experimental-no-entries/lingui.config.ts
+++ b/packages/cli/test/extractor-experimental-no-entries/lingui.config.ts
@@ -1,0 +1,13 @@
+import { defineConfig } from "@lingui/conf"
+
+export default defineConfig({
+  locales: ["en", "pl"],
+  sourceLocale: "en",
+  catalogs: [],
+  experimental: {
+    extractor: {
+      entries: ["<rootDir>/fixtures/does-not-exist/**/*.page.{ts,tsx}"],
+      output: "<rootDir>/actual/{entryName}.{locale}",
+    },
+  },
+})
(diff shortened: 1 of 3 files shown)


timofei-iatsenko:
@Adonis0123 the code in the `main` branch is outdated, please port your changes to the `next` branch. Start from the test, possible current implementation would already solve this issue.

Adonis0123:
I re-checked this issue and confirmed it appears fixed on `next` (via the tinypool migration path, no empty-queue wait on
  `pool.completed()` anymore).

  Would you like me to:
  1) close this PR as already resolved on `next`, or
  2) keep this PR and only add a dedicated regression test for the no-entry + worker-pool case?

  Also, could you share which release/version is expected to include this fix?

timofei-iatsenko:
> keep this PR and only add a dedicated regression test for the no-entry + worker-pool case?

This would be best option. 

> Also, could you share which release/version is expected to include this fix?

Next branch is for upcoming v6 release.

Adonis0123:
> > keep this PR and only add a dedicated regression test for the no-entry + worker-pool case?
> 
> This would be best option.
> 
> > Also, could you share which release/version is expected to include this fix?
> 
> Next branch is for upcoming v6 release.

Thanks for the clarification. I kept this PR and converted it to test-only.

  - Implementation fix is already on `next` (tinypool migration path).
  - This PR now only adds a dedicated regression test for the no-entry +
  worker-pool case.
  - Target release remains the upcoming v6 line.

timofei-iatsenko:
@Adonis0123 thank you!

PR #2449, opened 2026-02-11
chore: fix eslint config for react

yslpn:
# Description

After migrating to the new JSX transform, ESLint required additional configuration:
https://github.com/lingui/js-lingui/pull/2387

This PR includes additional follow-up changes after fixing the rules so linting works correctly for the React package and other files.

All eslint dependencies were updated.

And the config was slightly updated to better align with the flat config mental model.

This is was deprecated:
```ts
import { config } from "typescript-eslint"
```
new recommendation:
```ts
import { defineConfig } from "eslint/config"
```

## Types of changes

- [ ] Documentation update
- [ ] Examples update

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "test:types": "tstyche",
     "test:all": "yarn test && yarn test:e2e && yarn test:types",
     "lint:types": "yarn workspaces foreach -v run check-types",
-    "lint:eslint": "eslint ./packages",
+    "lint:eslint": "eslint ./packages --flag v10_config_lookup_from_file",
     "lint:all": "yarn lint:eslint && yarn lint:types",
     "prettier": "prettier --write '**/*.{ts,tsx,js,jsx}'",
     "prettier:check": "prettier --check '**/*.{ts,tsx,js,jsx}'",
@@ -36,12 +36,12 @@
     "node": ">=22.19.0"
   },
   "devDependencies": {
-    "@eslint/js": "^10.0.1",
+    "@eslint/js": "^9.39.3",
     "@size-limit/preset-small-lib": "^11.1.6",
     "@types/mock-fs": "^4.13.1",
     "@types/node": "22",
     "@vitest/coverage-v8": "4.0.18",
-    "eslint": "^10.0.0",
+    "eslint": "^9.39.3",
     "eslint-import-resolver-typescript": "^4.4.4",
     "eslint-plugin-import": "^2.32.0",
     "husky": "^8.0.3",
@@ -55,7 +55,7 @@
     "tstyche": "^4.0.0",
     "tsx": "^4.21.0",
     "typescript": "^5.7.2",
-    "typescript-eslint": "^8.55.0",
+    "typescript-eslint": "^8.56.1",
     "vitest": "4.0.18"
   },
   "workspaces": [
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -78,7 +78,7 @@
     "@lingui/test-utils": "workspace:*",
     "@testing-library/react": "^14.0.0",
     "@types/react": "^18.2.13",
-    "eslint": "^10.0.0",
+    "eslint": "^9.39.3",
     "eslint-plugin-react": "^7.37.5",
     "eslint-plugin-react-hooks": "^7.0.1",
     "happy-dom": "^20.0.10",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1381,68 +1381,80 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@eslint-community/regexpp@npm:^4.12.2":
+"@eslint-community/regexpp@npm:^4.12.1, @eslint-community/regexpp@npm:^4.12.2":
   version: 4.12.2
   resolution: "@eslint-community/regexpp@npm:4.12.2"
   checksum: 1770bc81f676a72f65c7200b5675ff7a349786521f30e66125faaf767fde1ba1c19c3790e16ba8508a62a3933afcfc806a893858b3b5906faf693d862b9e4120
   languageName: node
   linkType: hard
 
-"@eslint/config-array@npm:^0.23.0":
-  version: 0.23.1
-  resolution: "@eslint/config-array@npm:0.23.1"
+"@eslint/config-array@npm:^0.21.1":
+  version: 0.21.1
+  resolution: "@eslint/config-array@npm:0.21.1"
   dependencies:
-    "@eslint/object-schema": ^3.0.1
+    "@eslint/object-schema": ^2.1.7
     debug: ^4.3.1
-    minimatch: ^10.1.1
-  checksum: 5954433753f46fe40ee5d9267e795305fe0222d409cd306b30ab9f55d639a3942b1193175c8792ed55a125fe434514a844e95aadaa9aee3021e85556f8e6e0e8
+    minimatch: ^3.1.2
+  checksum: fc5b57803b059f7c1f62950ef83baf045a01887fc00551f9e87ac119246fcc6d71c854a7f678accc79cbf829ed010e8135c755a154b0f54b129c538950cd7e6a
   languageName: node
   linkType: hard
 
-"@eslint/config-helpers@npm:^0.5.2":
-  version: 0.5.2
-  resolution: "@eslint/config-helpers@npm:0.5.2"
+"@eslint/config-helpers@npm:^0.4.2":
+  version: 0.4.2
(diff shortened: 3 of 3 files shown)


packages/babel-plugin-lingui-macro/src/macroJs.ts
```diff
@@ -1,12 +1,18 @@
-import * as babelTypes from "@babel/types"
-import * as t from "@babel/types"
```
timofei-iatsenko, line 2: Were these changes caused by the new eslint rules? having `t.*` variable is just a convinient way used in babel plugins. IT provide better developer experience because you can invoke autocomplete after hitting `t.` and not go back and forth between code and import declaration to understand what methods exists. 

I would insist to keep the `t` import as it was.

packages/react/src/I18nProvider.tsx
```diff
@@ -43,7 +43,7 @@ export const I18nProvider = ({
   defaultComponent,
   children,
 }: I18nProviderProps) => {
-  const latestKnownLocale = useRef<string | undefined>(i18n.locale)
+  const latestKnownLocale = useRef<string | null>(i18n.locale || null)
```
timofei-iatsenko, line 46: What is the driver of this change?

eslint.config.js
```diff
@@ -48,17 +54,43 @@ export default config(
...
+      sourceType: "module",
+    },
+  },
+  {
+    files: ["packages/react/**/*.{ts,tsx,js,jsx}"],
+    extends: [
+      pluginReact.configs.flat.recommended,
+      pluginReact.configs.flat["jsx-runtime"],
+      reactHooks.configs.flat["recommended-latest"],
+    ],
```
timofei-iatsenko: i'm thinking would be tidier to leverage eslint's 10 new config resolver mechanism and create a separate eslint config in the react folder, so all dependncies would also be installed only for react package, wdyt?

eslint.config.js
```diff
@@ -48,17 +54,43 @@ export default config(
...
   {
-    files: ["./packages/react/*.{ts,tsx,js,jsx}"],
-    ...pluginReact.configs.flat.recommended,
+    // TODO: something is wrong with this file,
+    // it should be commonjs, but it has esm syntax, need to investigate
+    files: ["packages/**/macro/browser.cjs"],
+    languageOptions: {
+      sourceType: "module",
+    },
+  },
```
timofei-iatsenko: it's my fault, simply rename these files to mjs (and in package.json exports as well)

timofei-iatsenko (changes requested):

packages/react/src/I18nProvider.tsx
yslpn: This pattern is considered valid by the eslint rule. It doesn't affect the logic of code.

The error looks like this:

> 
> Error: Cannot access refs during render
> 
> React refs are values that are not needed for rendering. Refs should only be accessed outside of render, such as in event handlers or effects. Accessing a ref value (the `current` property) during render can cause your component not to update as expected (https://react.dev/reference/react/useRef).
> 
> To initialize a ref only once, check that the ref is null with the pattern `if (ref.current == null) { ref.current = ... }`eslint(react-hooks/refs)
>

packages/babel-plugin-lingui-macro/src/macroJs.ts
yslpn: Before this, `"eslint-plugin-import"` wasn't working at all. Now it's highlighting an error about duplicate imports. I'm thinking about how to make it more consistent.

I'll take another look at this today.

> I would insist to keep the t import as it was.

Got it

eslint.config.js
yslpn: I thought about it, but at first I didn't like the idea. Although now I think it's ok.

Perhaps this is a more scalable approach.

eslint.config.js
yslpn: Fixed in core and in react packages

packages/babel-plugin-lingui-macro/src/macroJs.ts
yslpn: Okay, I just added 

`eslint-disable-next-line import/no-duplicates`

packages/react/eslint.config.js
```diff
@@ -0,0 +1,18 @@
...
+export default defineConfig(...baseConfig, {
+  files: ["**/*.{ts,tsx,js,jsx}"],
+  extends: [
+    pluginReact.configs.flat.recommended,
+    pluginReact.configs.flat["jsx-runtime"],
+    reactHooks.configs.flat["recommended-latest"],
+  ],
+  settings: {
+    react: {
+      version: "18.2",
```
yslpn, line 18: Detect doesn't work anymore because it's looking for a react dependency in the project root. I assume so.

`version: "detect",`

timofei-iatsenko:
@yslpn you have conflicts here

packages/react/macro/index.d.mts
```diff
@@ -136,3 +136,9 @@ export function useLingui(): Omit<I18nContext, "_"> & {
     (literals: TemplateStringsArray, ...placeholders: any[]): string
   }
 }
+
+declare const linguiMacro: {
+  isBabelMacro: boolean
+}
+
+export default linguiMacro
```
timofei-iatsenko: These are internal implementation details of babel-macro-plugin, they should not be be exported from the public typings surface, since user's will not have them in runtime and they generally don't care.  What was the reason to add them here?

packages/react/macro/index.mjs
```diff
@@ -0,0 +1,8 @@
+import babelPluginMacros from "babel-plugin-macros"
+import { macro } from "@lingui/babel-plugin-lingui-macro/macro"
+
+const { createMacro } = babelPluginMacros
+
+export default createMacro(macro, {
+  configName: "lingui",
+})
```
timofei-iatsenko, line 8: The macro entrypoints were intentionally left as cjs, because `babel-plugin-macros` will consume them using `require` and there is no benefits of ESM here, but added a bit of friction.

`babel-plugin-macros` is old and generated in incompatible way so ESM discovery of the nodejs failed to analyze existing exports. 

But i see you adressed this, if it works - then fine, we can live it as esm for consistency

packages/react/macro/index.d.mts
yslpn: It took me a while to figure out how it worked before and how I could get a workaround to avoid TS error.

timofei-iatsenko:
Hey @yslpn is there anything you plan to do before merge?

yslpn:
> Hey @yslpn is there anything you plan to do before merge?

Initially, I didn't notice the numerous warnings about incompatibility with eslint version 10. I downgraded to version 9 but added the --flag v10_config_lookup_from_file flag. This means that in a future update, we'll need to remove the flag to ensure consistent behavior.

Now I've done everything I wanted. I think it can be merged, if there are no objections.

timofei-iatsenko:
> > Hey @yslpn is there anything you plan to do before merge?
> 
> Initially, I didn't notice the numerous warnings about incompatibility with eslint version 10. I downgraded to version 9 but added the --flag v10_config_lookup_from_file flag. This means that in a future update, we'll need to remove the flag to ensure consistent behavior.
> 
> Now I've done everything I wanted. I think it can be merged, if there are no objections.

What exactly was not compatible with eslint 10?

yslpn:
> > > Hey @yslpn is there anything you plan to do before merge?
> > 
> > 
> > Initially, I didn't notice the numerous warnings about incompatibility with eslint version 10. I downgraded to version 9 but added the --flag v10_config_lookup_from_file flag. This means that in a future update, we'll need to remove the flag to ensure consistent behavior.
> > Now I've done everything I wanted. I think it can be merged, if there are no objections.
> 
> What exactly was not compatible with eslint 10?

Version 10 was released quite recently, and not all packages have had time to update peerDeps.

```bash
❯ yarn install
➤ YN0000: ┌ Resolution step
➤ YN0060: │ @lingui/react@workspace:packages/react provides eslint (p1cd96) with version 10.0.3, which doesn't satisfy what eslint-plugin-react-hooks requests
➤ YN0060: │ @lingui/react@workspace:packages/react provides eslint (p8f34b) with version 10.0.3, which doesn't satisfy what eslint-plugin-react requests
➤ YN0060: │ @lingui/react@workspace:packages/react [14f17] provides eslint (p573d9) with version 10.0.3, which doesn't satisfy what eslint-plugin-react-hooks requests
➤ YN0060: │ @lingui/react@workspace:packages/react [14f17] provides eslint (p545e0) with version 10.0.3, which doesn't satisfy what eslint-plugin-react requests
```

PR #2456, opened 2026-02-14
fix(react): refresh i18n context identity for React Compiler memoization

### LINKED ISSUE #2343: Compatible with React Compiler?
React Compiler has now launched version 1.0. It seems that usages like directly using i18n.date or i18n.number inline within components, as mentioned in the documentation (https://lingui.dev/tutorials/react), will be memoized. Should Lingui become more "React Compiler" friendly?

yslpn:
# Description

Recreate `i18n` reference in `I18nProvider` via `Proxy` so memoized React Compiler paths update on locale change.
Add a regression test for memoized `useLingui().i18n` locale updates.

Avoiding recreating the initial state of the context by passing a function to useState
https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes #2343

## Checklist

### DIFF
--- a/packages/react/src/I18nProvider.tsx
+++ b/packages/react/src/I18nProvider.tsx
@@ -56,7 +56,7 @@ export const I18nProvider = ({
    * We can't use useMemo hook either, because we want to recalculate value manually.
    *
    * We wrap `i18n` in a Proxy to create a new reference on each context update.
-   * This ensures React Compiler correctly invalidates memoized values that depend on `i18n`.
+   * This ensures React correctly invalidates memoized values that depend on `i18n`.
    */
   const makeContext = useCallback(
     () => ({


timofei-iatsenko:
I'm not sure this is acceptable way of doing this. If upper component would re-render for some reason i18n would be re-rendered as well unnecesarily.  This might be a possible usecase when components used in monorepo and Provider specified not on the app root level, but on a distributable component level. 

i18n instance should change only in case locale switched / or new messages loaded. If i'm not mistaken there is a dedicated event for both of this actions.

yslpn:
Thanks for the feedback, @timofei-iatsenko. But I think everything is safe here. I'd like to discuss it.

> If upper component would re-render for some reason i18n would be re-rendered as well unnecesarily.

This would be true for code like `<LinguiContext.Provider value={makeContext()}>` on every render. But we're taking the i18n value from the state (`context`), so parent re-renders alone do not recreate context value.

> i18n instance should change only in case locale switched / or new messages loaded. If i'm not mistaken there is a dedicated event for both of this actions.

Agreed. That's right. We have two ways to pass a new i18n reference to the context.

1) Initial mount of `I18nProvider`, we pass a function to `useState(makeContext)` as initializer. Perhaps this could be written more explicitly, for example:
```ts
const [context, setContext] = useState<I18nContext>(() => makeContext())
```
2) `i18n.on("change")` callback, where we call `setContext(makeContext())`.

timofei-iatsenko:
Yes, you are right. I check code for the whole component and it's more clear now for me. 

BTW, i'm thinking of what the usecases left for using i18n object from context directly? 

`date` and other methods are depricated, `_` is already re-exported and updated when needed. The only one thing left is `i18n.locale`, may be it makes sense to just add `locale` directly to context, and make usage of i18n from context not needed? 

WDYT?

yslpn:
> Yes, you are right. I check code for the whole component and it's more clear now for me.
> 
> BTW, i'm thinking of what the usecases left for using i18n object from context directly?
> 
> `date` and other methods are depricated, `_` is already re-exported and updated when needed. The only one thing left is `i18n.locale`, may be it makes sense to just add `locale` directly to context, and make usage of i18n from context not needed?
> 
> WDYT?

I think the current approach is simpler and more predictable.

Cons:
- There will be two valid methods: `i18n.locale` and `locale`, and one will work poorly. It's difficult to enforce usage based on context alone. It can't be made a private field or deprecated.
- Additional breaking change.
- There are also locales; someone could tie conditional logic to that.
```ts
const locales = useMemo(
  () => (Array.isArray(i18n.locales) ? i18n.locales.join(",") : ""),
  [i18n],
)

```

PR #2475, opened 2026-03-13
fix(loader): make webpack peer dependency optional and update Rspack example

yslpn:
# Description

Extend `@lingui/loader` peer dependency metadata for Rspack and Rsbuild environments, and update the loader docs to describe Webpack-compatible usage more clearly.

UPD:

Made webpack optional peerDependencies and updated the rspack example

## Types of changes

- [x] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist


timofei-iatsenko:
@yslpn does it work without adding them into peer dependencies? I'm questioning what the reason of adding it explicitly? Usually we add something if we use (require or import) something from there, so package manager would build a correct graph.

But in this case it used only as a metadata to specify what version range of rsbuild/rspack is supported?  

Maybe it would be just enough to update the docs?

yslpn:
> @yslpn does it work without adding them into peer dependencies? I'm questioning what the reason of adding it explicitly? Usually we add something if we use (require or import) something from there, so package manager would build a correct graph.
> 
> But in this case it used only as a metadata to specify what version range of rsbuild/rspack is supported?
> 
> Maybe it would be just enough to update the docs?

It works but I'd like to remove warnings from package managers for projects that don't require webpack.

```bash
❯ yarn install
➤ YN0000: ┌ Resolution step
➤ YN0002: │ rspack-react-ts-starter@workspace:. doesn't provide webpack (p82fb2), requested by @lingui/loader
➤ YN0000: │ Some peer dependencies are incorrectly met; run yarn explain peer-requirements <hash> for details, where <hash> is the six-letter p-prefixed code
➤ YN0000: └ Completed
➤ YN0000: ┌ Fetch step
➤ YN0000: └ Completed
➤ YN0000: ┌ Link step
➤ YN0008: │ esbuild@npm:0.25.12 must be rebuilt because its dependency tree changed
```

yslpn:
Should I add "@lingui/loader" to the rspack example?

timofei-iatsenko:
> > @yslpn does it work without adding them into peer dependencies? I'm questioning what the reason of adding it explicitly? Usually we add something if we use (require or import) something from there, so package manager would build a correct graph.
> > But in this case it used only as a metadata to specify what version range of rsbuild/rspack is supported?
> > Maybe it would be just enough to update the docs?
> 
> It works but I'd like to remove warnings from package managers for projects that don't require webpack.
> 
> ```shell
> ❯ yarn install
> ➤ YN0000: ┌ Resolution step
> ➤ YN0002: │ rspack-react-ts-starter@workspace:. doesn't provide webpack (p82fb2), requested by @lingui/loader
> ➤ YN0000: │ Some peer dependencies are incorrectly met; run yarn explain peer-requirements <hash> for details, where <hash> is the six-letter p-prefixed code
> ➤ YN0000: └ Completed
> ➤ YN0000: ┌ Fetch step
> ➤ YN0000: └ Completed
> ➤ YN0000: ┌ Link step
> ➤ YN0008: │ esbuild@npm:0.25.12 must be rebuilt because its dependency tree changed
> ```

Got it, so basically it's more about seting a webpack as optional. Maybe then let's just do that? Set webpack as optional without adding rspack and friends as peer deps.  

> Should I add "@lingui/loader" to the rspack example?

Yes please.

yslpn:
> Got it, so basically it's more about seting a webpack as optional. Maybe then let's just do that? Set webpack as optional without adding rspack and friends as peer deps.

I don't mind, but I want to understand. Are you concerned that specifying rspack/rsbuild will lead to new code additions and make maintenance more difficult?

On the other hand, I see benefits in defining acceptable versions.

> > Should I add "@lingui/loader" to the rspack example?
> 
> Yes please.

Done

timofei-iatsenko:
> I don't mind, but I want to understand. Are you concerned that specifying rspack/rsbuild will lead to new code additions and make maintenance more difficult?

This list is not completed, we will also need to specify turbopack version as webpack compatible (anything else what i'm not aware of?) and maintain the ranges unnecessarily. I believe all these webpack compatible tools is compatible with latest webpack api, so it's on their side to maintain compatibility - not on ours.

yslpn:
> > I don't mind, but I want to understand. Are you concerned that specifying rspack/rsbuild will lead to new code additions and make maintenance more difficult?
> 
> This list is not completed, we will also need to specify turbopack version as webpack compatible (anything else what i'm not aware of?) and maintain the ranges unnecessarily. I believe all these webpack compatible tools is compatible with latest webpack api, so it's on their side to maintain compatibility - not on ours.

Done. Pull request description updated.

timofei-iatsenko (approved):

timofei-iatsenko:
@andrii-bodnar all good let's merge this one

PR #2479, opened 2026-03-17
fix: really keep catalog extra from previous catalog

### LINKED ISSUE #2398: PO file flags are lost when running extract
### Description

PO file flags (like #, fuzzy or custom flags like #, myTag) are lost when running lingui extract again. The extra.flags field from the existing catalog is not preserved during the merge process.

The bug is in packages/cli/src/api/catalog/mergeCatalog.ts. When merging catalogs, the code takes rest from nextCatalog (extracted messages) which doesn't contain flags, and only preserves translation from prevCatalog. The extra.flags field is silently discarded.

### Verifications

### Reproduction Steps

# 1. Create a PO file with a flag
cat > messages.po << 'EOF'
msgid "Hello"
msgstr "Hallo"
#, myTag
EOF

# 2. Run extract
lingui extract

# 3. Check the PO file - the flag "#, myTag" is gone

Or programmatically:

import { mergeCatalog } from "@lingui/cli/api"

const prevCatalog = {
  "Hello": {
    translation: "Hallo",
    extra: { flags: ["myTag"] }
  }
}

const nextCatalog = {
  "Hello": {
    message: "Hello",
    origin: [["src/app.ts", 1]]
  }
}

const result = mergeCatalog(prevCatalog, nextCatalog, false, {})
console.log(result["Hello"].extra?.flags) // undefined

### Expected Behavior

The flags should be preserved from the existing catalog when running extract. Flags are translator/tooling metadata that should persist across extractions, similar to how translation is preserved.

### Macro Support

Not using macro

### Lingui Version

5.8.0

### Babel Version

@babel/core@7.28.3

### Framework

Affects CLI command

ljani:
25cfdde3b0c3d95697752a0e698bff9ede6d9a4e tries to keep `flags`, but it is incorrectly spreading `extra`, when it should not, causing `flags` to be lost.

Additionally, move `extra` after `rest` to ensure it is not overwritten. Not sure if this is needed, but the original commit hints it should be always preferred.

See also #2399.

Running the full test suite seems to fail on my machine, but the relevant tests pass.

# Description

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes #2398.

## Checklist

### DIFF
--- a/packages/cli/src/api/catalog/mergeCatalog.test.ts
+++ b/packages/cli/src/api/catalog/mergeCatalog.test.ts
@@ -170,9 +170,11 @@ describe("mergeCatalog", () => {
     const result = mergeCatalog(prevCatalog, nextCatalog, false, {})
     expect(result["Hello"]).toMatchInlineSnapshot(`
       {
-        flags: [
-          myTag,
-        ],
+        extra: {
+          flags: [
+            myTag,
+          ],
+        },
         message: Hello,
         origin: [
           [
--- a/packages/cli/src/api/catalog/mergeCatalog.ts
+++ b/packages/cli/src/api/catalog/mergeCatalog.ts
@@ -40,7 +40,7 @@ export function mergeCatalog(
       const { obsolete, ...rest } = nextCatalog[key]
       const { extra } = prevCatalog[key]
 
-      return [key, { ...extra, ...rest, translation }]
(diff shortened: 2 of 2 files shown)


timofei-iatsenko:
😳 omg, may bad. Thanks for the fix

PR #2505, opened 2026-04-09
feat: add optional configurable JSX placeholder naming

### LINKED ISSUE #1188: Allow named components in Trans in @lingui/macro
The [documentation of `Trans`](https://lingui.js.org/ref/macro.html#trans) has the following example:

```jsx
<Trans id="Read the <0>docs</0>." components={{0: <a href="/docs" />}} />
```

Since the number tags don't provide much context to the translator I was hoping I could use named tags instead as such:

```jsx
<Trans id="Read the <link>docs</link>." components={{link: <a href="/docs" />}} />
```

Unfortunately, it seems only numbered tags are supported.

mogelbrod:
# Description

Add a new feature that allows developers to customize the identifiers of JSX placeholders generated by the `<Trans>` macro. This results in placeholder identifiers in extracted messages that are more stable, human-readable, and contextual, instead of the default arbitrary numeric indices (e.g., replacing `<0>` with `<em>` or `<link>`).

It is accomplished via two new optional configuration options (via `lingui.config.js` in the `macro` options):
1. `jsxPlaceholderAttribute`: Allows defining a custom prop name (e.g., `_t`) that can be used on components to set an explicit placeholder name locally.
2. `jsxPlaceholderDefaults`: Allows providing a record of tag names to placeholder names, enabling global default renaming for specific JSX components.

SWC implementation at https://github.com/lingui/swc-plugin/pull/207

## Types of changes

- [x] Documentation update
- [ ] Examples update

Fixes #1188

## Checklist

- [x] I have added the necessary documentation (if appropriate) - _will do before merging_

#### lingui.config.js
```ts

export default defineConfig({
  sourceLocale: 'en',
  locales: ['en', 'sv'],
  macro: {
    jsxPlaceholderAttribute: '_t',
    jsxPlaceholderDefaults: {
      a: 'a',
      b: 'b',
      br: 'br',
    },
  },
})
```

#### globals.d.ts or similarly imported definitions file
```ts
import 'react'

declare module 'react' {
  interface Attributes {
    _t?: string
  }
}
```

#### source-file.tsx
```tsx
import { Trans } from '@lingui/react/macro'

function App() {
  return <Trans>Translated <a href="/">link</a> with some <div _t="div">additional</div> markup.</Trans>
}
```

## Disclaimer

Much of this PR was generated by Copilot using the Gemini 3.1 Pro (Preview) model with training/sharing disabled.

<details>
<summary>Initial prompt used</summary>

## Plan: Implement JSX named placeholders

Add opt-in support for named JSX placeholders through Lingui configuration (`jsxPlaceholderAttribute` and `jsxPlaceholderDefaults`), replacing numeric indices in extracted messages for better maintainability and readability. Include smart element deduplication that reuses element names if their props are identical, otherwise with direct suffix appended for duplicated names.

**Steps**
1. **Extend LinguiConfig**: Update type definition in types.ts to include `jsxPlaceholderAttribute?: string` and `jsxPlaceholderDefaults?: Record<string, string>` in the `macro` object.
2. **Update Validation Configuration**: Add these new properties to `exampleConfig` in makeConfig.ts so `jest-validate` will accept them. Let `defaultConfig` omit them (or set them to `undefined` / empty object).
3. **Pass Config to Macro**: In index.ts, extract the new options from the `linguiConfig` macro passing them to the `MacroJSX` instantiation inside `MacroJsxOpts`. 
4. **Update MacroJsxOpts Interface**: Expand `MacroJsxOpts` in macroJsx.ts to accept `jsxPlaceholderAttribute` and `jsxPlaceholderDefaults`.
5. **Add State for Placeholder Names**: Introduce a tracked collection in `macroJsx.ts` (scoped to `MacroJsxContext` per message evaluation) to track used placeholder names and their original JSX node props.
6. **Modify `tokenizeElement`**: In macroJsx.ts, parse out the requested name:
   - Check the JSX node's attributes for an exact match to `jsxPlaceholderAttribute`.
   - If found, use its string literal value as the requested placeholder name, and **remove the attribute** from the node's AST.
   - If not found, check `jsxPlaceholderDefaults` using the literal tag name.
   - If neither are present, fallback to the current behavior `this.ctx.elementIndex()`.
   - Apply deduplication: Check if the base name is already used.
     - Look up previously stored element by that name.
     - **Reuse rule:** Evaluate whether the previously stored node and the current node have **identical props** (either both have no additional props, or their existing props are structurally equal). If equal, reuse the exact same key without incrementing.
     - **Collision rule:** If the props differ (e.g. they both have different `href` props), generate a new key by directly appending an incremental numeric suffix to the original name (starting at 2, e.g. `a` -> `a2`, `a3`), avoiding an underscore separator. Store it and assign the newly suffixed name to the current element.
7. **Create Tests**: Add a test fixture testing scenarios with `_t` attribute, tag-name defaults, deep nesting, prop deduplication (including equal props and differing props), and direct suffix logic.

**Relevant files**
- packages/conf/src/types.ts — Add properties to `LinguiConfig.macro` type.
- packages/conf/src/makeConfig.ts — Update `exampleConfig` for validator compatibility.
- packages/babel-plugin-lingui-macro/src/index.ts — Route configuration from `macroContext` to `MacroJSX`.
- packages/babel-plugin-lingui-macro/src/macroJsx.ts — Implement name lookup, AST attribute pruning, and the prop equality / direct-suffix (`x2`) element duplication rules.

**Verification**
1. Validate `pnpm test` in babel-plugin-lingui-macro.
2. Compare emitted ICU format visually to directly match the test cases:
   - Identical elements (same props or exactly no props) reuse the same name.
   - Distinct elements with same tag/name emit `<a>...</a>` and `<a2>...</a2>`.

**Decisions**
- The configuration attribute is explicitly stripped from the extracted component output to avoid unused React props and potential runtime warnings.
- Deduplication relies on checking equality of props (or both having zero props) to map them to the same component index.
- Duplicate names follow direct appending mapping e.g., `x`, `x2`, `x3`.

</details>

### DIFF
--- a/packages/babel-plugin-lingui-macro/src/macroJsx.ts
+++ b/packages/babel-plugin-lingui-macro/src/macroJsx.ts
@@ -49,6 +49,9 @@ function maybeNodeValue(node: Node): { text: string; loc: SourceLocation } {
 export type MacroJsxContext = MacroJsContext & {
   elementIndex: () => number
   transImportName: string
+  elementsTracking: Map<string, JSXElement>
+  jsxPlaceholderAttribute?: string
+  jsxPlaceholderDefaults?: Record<string, string>
 }
 
 export type MacroJsxOpts = {
@@ -56,6 +59,8 @@ export type MacroJsxOpts = {
   stripMessageProp: boolean
   transImportName: string
   isLinguiIdentifier: (node: Identifier, macro: JsMacroName) => boolean
+  jsxPlaceholderAttribute?: string
+  jsxPlaceholderDefaults?: Record<string, string>
 }
 
 const choiceComponentAttributesWhitelist = [
@@ -86,6 +91,9 @@ export class MacroJSX {
       ),
       transImportName: opts.transImportName,
       elementIndex: makeCounter(),
+      elementsTracking: new Map(),
+      jsxPlaceholderAttribute: opts.jsxPlaceholderAttribute,
+      jsxPlaceholderDefaults: opts.jsxPlaceholderDefaults,
     }
   }
 
@@ -351,18 +359,85 @@ export class MacroJSX {
   }
 
   tokenizeElement = (path: NodePath<JSXElement>): ElementToken => {
-    // !!! Important: Calculate element index before traversing children.
-    // That way outside elements are numbered before inner elements. (...and it looks pretty).
-    const name = this.ctx.elementIndex()
+    const {
+      jsxPlaceholderAttribute,
+      jsxPlaceholderDefaults,
+      elementsTracking,
+    } = this.ctx
+
+    let node = path.node
+    let name: string | undefined = undefined
+
+    if (jsxPlaceholderAttribute) {
+      const { attributes } = node.openingElement
+      const attrIndex = attributes.findIndex(
+        (attr) =>
+          attr.type === "JSXAttribute" &&
+          attr.name.name === jsxPlaceholderAttribute,
+      )
+
+      if (attrIndex !== -1) {
+        const attr = attributes[attrIndex] as JSXAttribute
+        if (attr.value && attr.value.type === "StringLiteral") {
+          name = attr.value.value
+        }
+
+        const newAttributes = [...attributes]
+        newAttributes.splice(attrIndex, 1)
+
+        node = {
+          ...node,
+          openingElement: {
+            ...node.openingElement,
+            attributes: newAttributes,
+          },
+        }
+      }
+    }
+
+    if (!name && jsxPlaceholderDefaults) {
+      const tagName = node.openingElement.name
+      if (tagName.type === "JSXIdentifier") {
+        name = jsxPlaceholderDefaults[tagName.name]
+      }
+    }
+
+    if (!name) {
+      name = String(this.ctx.elementIndex())
+      elementsTracking.set(name, node)
+    } else {
+      const existingElement = elementsTracking.get(name)
+
+      if (existingElement) {
+        const existingAttrs = existingElement.openingElement.attributes
+        const openingAttrs = node.openingElement.attributes
+        if (
+          existingAttrs.length !== openingAttrs.length ||
+          !existingAttrs.every((a) =>
+            openingAttrs.some((b) => this.types.isNodesEquivalent(a, b)),
+          )
+        ) {
+          const eg = `(e.g. \`<element ${jsxPlaceholderAttribute || '_t'}="newName" />\`)`
+          const msg = `Multiple distinct JSX elements with the same placeholder name (\`${name}\`). `
+            + (jsxPlaceholderAttribute
+              ? `Differentiate them by adding/modifying the \`${jsxPlaceholderAttribute}\` attribute ${eg}.`
+              : `Differentiate them by setting \`macro.jsxPlaceholderAttribute\` in the lingui config and then adding the attribute to your JSX elements ${eg}.`
+            )
+          throw path.buildCodeFrameError(msg)
+        }
+      } else {
+        elementsTracking.set(name, node)
+      }
+    }
 
     return {
       type: "element",
       name,
       value: {
-        ...path.node,
+        ...node,
         children: [],
         openingElement: {
-          ...path.node.openingElement,
+          ...node.openingElement,
           selfClosing: true,
         },
       },
--- a/packages/babel-plugin-lingui-macro/src/macroJsx.ts
+++ b/packages/babel-plugin-lingui-macro/src/macroJsx.ts
@@ -378,9 +378,16 @@ export class MacroJSX {
 
       if (attrIndex !== -1) {
         const attr = attributes[attrIndex] as JSXAttribute
-        if (attr.value && attr.value.type === "StringLiteral") {
-          name = attr.value.value
+        if (
+          !attr.value ||
+          attr.value.type !== "StringLiteral" ||
+          !attr.value.value
+        ) {
+          throw path.buildCodeFrameError(
+            `The \`${jsxPlaceholderAttribute}\` attribute must be a non-empty string literal.`,
+          )
         }
+        name = attr.value.value
 
         const newAttributes = [...attributes]
         newAttributes.splice(attrIndex, 1)
@@ -406,12 +413,26 @@ export class MacroJSX {
       name = String(this.ctx.elementIndex())
       elementsTracking.set(name, node)
     } else {
+      if (/^\d+$/.test(name)) {
+        throw path.buildCodeFrameError(
+          `Placeholder name \`${name}\` is not allowed because it conflicts with auto-generated numeric placeholders. Use a non-numeric name instead.`,
+        )
+      }
+      if (!/^[a-zA-Z_$][a-zA-Z0-9_$]*$/.test(name)) {
+        throw path.buildCodeFrameError(
+          `Placeholder name \`${name}\` is not a valid identifier. Only valid JS identifers are allowed; /^[a-zA-Z_$][a-zA-Z0-9_$]*$/`,
+        )
+      }
+
       const existingElement = elementsTracking.get(name)
 
       if (existingElement) {
+        const existingTag = existingElement.openingElement.name
+        const currentTag = node.openingElement.name
         const existingAttrs = existingElement.openingElement.attributes
         const openingAttrs = node.openingElement.attributes
         if (
+          !this.types.isNodesEquivalent(existingTag, currentTag) ||
           existingAttrs.length !== openingAttrs.length ||
           !existingAttrs.every((a) =>
             openingAttrs.some((b) => this.types.isNodesEquivalent(a, b)),
--- a/packages/babel-plugin-lingui-macro/src/macroJsx.ts
+++ b/packages/babel-plugin-lingui-macro/src/macroJsx.ts
@@ -431,12 +431,26 @@ export class MacroJSX {
         const currentTag = node.openingElement.name
         const existingAttrs = existingElement.openingElement.attributes
         const openingAttrs = node.openingElement.attributes
+
+        const hasSpreads = existingAttrs.some(
+          (a) => a.type === "JSXSpreadAttribute",
+        )
+
+        // When spreads are present, attribute order matters for React
+        // semantics so we compare positionally. Otherwise, order-insensitive.
+        const attrsEqual =
+          existingAttrs.length === openingAttrs.length &&
+          (hasSpreads
+            ? existingAttrs.every((a, i) =>
+                this.types.isNodesEquivalent(a, openingAttrs[i]),
+              )
+            : existingAttrs.every((a) =>
+                openingAttrs.some((b) => this.types.isNodesEquivalent(a, b)),
+              ))
+
         if (
           !this.types.isNodesEquivalent(existingTag, currentTag) ||
-          existingAttrs.length !== openingAttrs.length ||
-          !existingAttrs.every((a) =>
-            openingAttrs.some((b) => this.types.isNodesEquivalent(a, b)),
-          )
+          !attrsEqual
         ) {
           const eg = `(e.g. \`<element ${jsxPlaceholderAttribute || '_t'}="newName" />\`)`
           const msg = `Multiple distinct JSX elements with the same placeholder name (\`${name}\`). `
--- a/packages/babel-plugin-lingui-macro/src/macroJsx.ts
+++ b/packages/babel-plugin-lingui-macro/src/macroJsx.ts
@@ -452,12 +452,12 @@ export class MacroJSX {
           !this.types.isNodesEquivalent(existingTag, currentTag) ||
           !attrsEqual
         ) {
-          const eg = `(e.g. \`<element ${jsxPlaceholderAttribute || '_t'}="newName" />\`)`
-          const msg = `Multiple distinct JSX elements with the same placeholder name (\`${name}\`). `
-            + (jsxPlaceholderAttribute
+          const eg = `(e.g. \`<element ${jsxPlaceholderAttribute || "_t"}="newName" />\`)`
+          const msg =
+            `Multiple distinct JSX elements with the same placeholder name (\`${name}\`). ` +
+            (jsxPlaceholderAttribute
               ? `Differentiate them by adding/modifying the \`${jsxPlaceholderAttribute}\` attribute ${eg}.`
-              : `Differentiate them by setting \`macro.jsxPlaceholderAttribute\` in the lingui config and then adding the attribute to your JSX elements ${eg}.`
-            )
+              : `Differentiate them by setting \`macro.jsxPlaceholderAttribute\` in the lingui config and then adding the attribute to your JSX elements ${eg}.`)
           throw path.buildCodeFrameError(msg)
         }
       } else {
--- a/packages/babel-plugin-lingui-macro/src/macroJsx.ts
+++ b/packages/babel-plugin-lingui-macro/src/macroJsx.ts
@@ -418,9 +418,9 @@ export class MacroJSX {
           `Placeholder name \`${name}\` is not allowed because it conflicts with auto-generated numeric placeholders. Use a non-numeric name instead.`,
         )
       }
-      if (!/^[a-zA-Z_$][a-zA-Z0-9_$]*$/.test(name)) {
+      if (!/^[a-zA-Z_]([\w.-]*\w)?$/.test(name)) {
         throw path.buildCodeFrameError(
-          `Placeholder name \`${name}\` is not a valid identifier. Only valid JS identifers are allowed; /^[a-zA-Z_$][a-zA-Z0-9_$]*$/`,
+          `Placeholder name \`${name}\` is not valid. Names must start and end with a letter/digit/underscore, but may contain \`.-\` in between.`,
         )
       }
 
--- a/packages/babel-plugin-lingui-macro/test/__snapshots__/jsx-placeholder.test.ts.snap
+++ b/packages/babel-plugin-lingui-macro/test/__snapshots__/jsx-placeholder.test.ts.snap
@@ -122,6 +122,23 @@ import { Trans as _Trans } from "@lingui/react";
 
 `;
 
+exports[`Deduplication: Same name on different element types throws an error 1`] = `
+[SyntaxError: <cwd>/<filename>jsx: Unsupported macro usage. Please check the examples at https://lingui.dev/ref/macro#examples-of-js-macros.
+ If you think this is a bug, fill in an issue at https://github.com/lingui/js-lingui/issues
+
+ Error: Multiple distinct JSX elements with the same placeholder name (\`same\`). Differentiate them by adding/modifying the \`_t\` attribute (e.g. \`<element _t="newName" />\`).
+  1 |
+  2 |         import { Trans } from '@lingui/react/macro';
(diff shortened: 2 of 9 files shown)


packages/babel-plugin-lingui-macro/test/__snapshots__/jsx-placeholder.test.ts.snap
```diff
@@ -0,0 +1,196 @@
...
+    ...{
+      id: "Qb0oHL",
+      message: "Hello <a>link 1</a>, normal, <a2>link 2</a2>.",
+      components: {
+        a: <a href="/a" />,
+        a2: <a href="/b" />,
+      },
+    }
+  }
+/>;
```
timofei-iatsenko: I think it's not very obvious behaviour, i would be more explicit here and throw an error if 2 elements with the same name but different attributes are found, forcing a user to provide a custom devloper crafted name, for example 

`Accept our <tnc-link>Terms and conditions</> and <privace-policy-link>Privacy policy</privace-policy-link>`

What do you think?

packages/babel-plugin-lingui-macro/test/__snapshots__/jsx-placeholder.test.ts.snap
```diff
@@ -0,0 +1,196 @@
...
+      message: "Hello <link1>link 1</link1>, normal, <link2>link 2</link2>.",
+      components: {
+        link1: <a href="/a" />,
+        link2: <a href="/b" />,
+      },
+    }
+  }
+/>;
+
+`;
```
timofei-iatsenko: I don't understand this case, it seems the code is not aligned with the case name. In both jsx elements inside the `_t` is used

packages/babel-plugin-lingui-macro/test/__snapshots__/jsx-placeholder.test.ts.snap
mogelbrod: I think that makes sense! Will update this PR, re-request a review from you and once it looks good also mirror the changes in the SWC PR.

packages/babel-plugin-lingui-macro/test/__snapshots__/jsx-placeholder.test.ts.snap
mogelbrod: Good catch - will update!

timofei-iatsenko:
@yslpn @Bertg @andrii-bodnar hey, could you share your opinion on the feature design?

I think the feature is cool, but i want to make sure that we don't forget anything and the initial design is good to go.

timofei-iatsenko:
@mogelbrod these new options should also be added to the types/validation in the lingui config.  Please also add these new options with a descriptive jsdoc inline comments, explaining how they works and what affects.

packages/babel-plugin-lingui-macro/test/__snapshots__/jsx-placeholder.test.ts.snap
```diff
@@ -0,0 +1,203 @@
...
+
+exports[`Deduplication: Implicit names with distinct attributes throw an error 1`] = `
+[SyntaxError: <cwd>/<filename>jsx: Unsupported macro usage. Please check the examples at https://lingui.dev/ref/macro#examples-of-js-macros.
+ If you think this is a bug, fill in an issue at https://github.com/lingui/js-lingui/issues
+
+ Error: Multiple distinct JSX elements with the same placeholder name (\`a\`). Differentiate them by adding/modifying the JSX attribute (e.g. \`<element undefined="newName" />\`).
```
andrii-bodnar: element undefined?

mogelbrod:
> @mogelbrod these new options should also be added to the types/validation in the lingui config. Please also add these new options with a descriptive jsdoc inline comments, explaining how they works and what affects.

Thanks for the reminder! This should be done in the most recent push 👍

andrii-bodnar (commented):
@mogelbrod thanks for the contribution!

It looks good to me. I can't think of any risks or cases that were overlooked. I have left one minor comment.

packages/babel-plugin-lingui-macro/test/__snapshots__/jsx-placeholder.test.ts.snap
mogelbrod: Good catch! I've updated the error message to provide slightly different instructions depending on if `jsxPlaceholderAttribute` is on or not. Does this sound reasonable?

`jsxPlaceholderAttribute: '_t'`:
>  Error: Multiple distinct JSX elements with the same placeholder name (`link`). Differentiate them by adding/modifying the `_t` attribute (e.g. `<element _t="newName" />`).


`jsxPlaceholderAttribute: undefined`:
>   Error: Multiple distinct JSX elements with the same placeholder name (`a`). Differentiate them by setting `macro.jsxPlaceholderAttribute` in the lingui config and then adding the attribute to your JSX elements (e.g. `<element _t="newName" />`).

mogelbrod: It would also be possible to only allow `jsxPlaceholderDefaults` if `jsxPlaceholderAttribute` is also enabled, but I figured it's better to let the developer decide.

timofei-iatsenko:
@mogelbrod please don't rebase and force push, i'm going to jump on your branch and help with failing snapshots.

yslpn:
> @yslpn @Bertg @andrii-bodnar hey, could you share your opinion on the feature design?

Regarding the feature design, in my personal opinion, it would be better to start with a simple implementation, where the names are taken from the JSX tag name: `<a href="/">link</a>` => `<a>link<a>`, `<button type="button">Click Me!</button>` => `<button>Click Me!</button>`, and so on. Only then, in the future, add customization via `lingui.config.js` and a custom `_t` attribute.

This will be easier to maintain and more safe.

UPD: I've found several very critical issues in the current implementation that are very difficult to detect. Therefore, I would recommend starting with a simpler version. I'm afraid there are still some critical issues I simply haven't found.

packages/babel-plugin-lingui-macro/src/macroJsx.ts
```diff
@@ -351,18 +359,85 @@ export class MacroJSX {
...
+      const { attributes } = node.openingElement
+      const attrIndex = attributes.findIndex(
+        (attr) =>
+          attr.type === "JSXAttribute" &&
+          attr.name.name === jsxPlaceholderAttribute,
+      )
+
+      if (attrIndex !== -1) {
+        const attr = attributes[attrIndex] as JSXAttribute
+        if (attr.value && attr.value.type === "StringLiteral") {
```
yslpn: If a custom placeholder attribute is present but its value is not a non-empty string literal, such as 

`_t={linkName} `or `_t=""`, 

this code still removes the attribute from the JSX and then silently falls back to the auto-generated numeric placeholder.

Throw an error?

packages/babel-plugin-lingui-macro/src/macroJsx.ts
```diff
@@ -351,18 +359,85 @@ export class MacroJSX {
...
+          name = attr.value.value
```
yslpn: If `_t` contains a string like `"foo-bar"` or `"0"`, this value is further used as the tag name in the message and as a key in components. `"foo-bar"` makes the generated object syntactically invalid `(foo-bar: <a />)`, and `"0"` conflicts with the auto-generated numeric placeholders for unnamed elements in the same <Trans>, causing one of the component mappings to be silently overwritten. The new branch should validate or reserve such names before generating code.

packages/babel-plugin-lingui-macro/src/macroJsx.ts
```diff
@@ -351,18 +359,85 @@ export class MacroJSX {
...
+    }
+
+    if (!name) {
+      name = String(this.ctx.elementIndex())
+      elementsTracking.set(name, node)
+    } else {
+      const existingElement = elementsTracking.get(name)
+
+      if (existingElement) {
+        const existingAttrs = existingElement.openingElement.attributes
```
yslpn, line 432: 1) Both placeholders will be rendered as either `<em />` or `<strong />` depending on the traversal order.

```tsx
<Trans>
  <em _t="same">A</em> and <strong _t="same">B</strong>
</Trans>
```

2) The order of attributes isn't compared. For regular props, this often doesn't matter, but for spread, it's important:

```tsx
const props = { href: "/b" }

<Trans>
  <a _t="same" {...props} href="/a">A</a>
  <a _t="same" href="/a" {...props}>B</a>
</Trans>
```

That is, the React semantics are different. But the current check sees "there is a spread and there is an href" in both cases and considers the elements identical. After this, one of them quietly wins, and both placeholders begin to render identically, even though the original JSX was different.

mogelbrod:
> Regarding the feature design, in my personal opinion, it would be better to start with a simple implementation, where the names are taken from the JSX tag name: `<a href="/">link</a>` => `<a>link<a>`, `<button type="button">Click Me!</button>` => `<button>Click Me!</button>`, and so on. Only then, in the future, add customization via lingui.config.js and a custom _t attribute.
> 
> This will be easier to maintain and more safe.

Thanks for the feedback! Limiting the functionality to only use the component name as placeholder name with no way of overriding it introduces a number of issues. Most notably, it would no longer be possible to use two or more of the same JSX element in a single `<Trans>` element due to naming collisions. Enabling the setting would thus not be possible in code bases where this happens. If a developer needs to add an already occurring JSX element to an existing `<Trans>` block they'd have to turn the option off.

Replacing one JSX element with another (but otherwise not touching the copy) would also require a re-translation due to the name changing. Eg. `<a href="/">copy</a>` -> `<Link href="/">copy</Link>` results in different placeholders, even though the placeholder itself hasn't fundamentally changed.

Offering both options seems necessary to ensure maintainability and giving developers a tool to work around issues that could otherwise arise.

Appreciate the comments @yslpn! I'll get right on addressing the issues you raised.

yslpn:
> Most notably, it would no longer be possible to use two or more of the same JSX element in a single `<Trans>` element due to naming collisions.

Why? We can add indexes just like we did before for `<0><0>`, `<1><1>`, and so on. Just add indexes to the tag names.

```
<button1><button1>
<button2><button2>
```

packages/babel-plugin-lingui-macro/src/macroJsx.ts
mogelbrod: Thanks! Should be addressed in a18eeea83454581a8f3c11995057a2796adc8ed6

packages/babel-plugin-lingui-macro/src/macroJsx.ts
mogelbrod: Thanks! Should be addressed in 24d15331d66ba0bfc75eb7c14f6b392efcdb262a

packages/babel-plugin-lingui-macro/src/macroJsx.ts
mogelbrod: Thanks! Should be addressed in a18eeea83454581a8f3c11995057a2796adc8ed6
Please resolve this thread if you think the solution covers the issue @yslpn 😃

mogelbrod:
> Why? We can add indexes just like we did before for `<0><0>`, `<1><1>`, and so on. Just add indexes to the tag names.
> 
> ```
> <button1><button1>
> <button2><button2>
> ```

Ah gotcha - that more closely mirrored the original behaviour of this PR. But it also suffers from the same problem as the current placeholders (`<0>`, `<1>`, …) where re-arranging elements in the source string causes their names to change, which may break translations.

packages/babel-plugin-lingui-macro/src/macroJsx.ts
timofei-iatsenko: I believe this `"foo-bar"` should works fine. When you print AST back to the code it will automatically turn into 

```js
components: {
      "foo-bar": <a />,
    },
```

But i would add an explicit test with this value to check. 

Regarding `"0"`, you are right, i either should be explicitly validated or went the same path of "duplication" and throw an error.

packages/babel-plugin-lingui-macro/src/macroJsx.ts
timofei-iatsenko: Good catch. However i think this is quite extreme edge case

timofei-iatsenko:
> Regarding the feature design, in my personal opinion, it would be better to start with a simple implementation, where the names are taken from the JSX tag name: `<a href="/">link</a>` => `<a>link<a>`, `<button type="button">Click Me!</button>` => `<button>Click Me!</button>`, and so on. Only then, in the future, add customization via `lingui.config.js` and a custom `_t` attribute.
> 
> This will be easier to maintain and more safe.
> 
> UPD: I've found several very critical issues in the current implementation that are very difficult to detect. Therefore, I would recommend starting with a simpler version. I'm afraid there are still some critical issues I simply haven't found.

Taking the name from a just a JSX element tag name has a little value in my opinion. From my experience usually only simple elements inside of the translation strings are used, mostly `strong` or `link`, `br` etc.  Taking the name of JSX element into a translation we pushing an internal implementation details to a translation.

In most of the projects i worked on, links were implemented in a few different way such as `<Link>` (as custom element), as `<a>` html element, as `<button>` and so on. So inferring the name from the jsx tag will open a pandora's box of DX issues which we will throw to our users. 

The proposed design allows to set a semantic, stable name for the placeholder which will not depend on the internal implementation. You can swap `<a>` to `<Link>` without any issue. 

So basically users would be able to define a default set such as `strong` `br` `a`. And then provide an explicit name in every usage place. 

>  UPD: I've found several very critical issues in the current implementation that are very difficult to detect. Therefore, I would recommend starting with a simpler version. I'm afraid there are still some critical issues I simply haven't found.

Appreciate for the digging deeper in this PR code. I think the initial design is OK for me. We can mark this feature as experimental, users don't have to use right away, but those who will start could provide feedback and we can improve all the cases in the future.

packages/babel-plugin-lingui-macro/src/macroJsx.ts
```diff
@@ -351,18 +359,120 @@ export class MacroJSX {
...
+      if (/^\d+$/.test(name)) {
+        throw path.buildCodeFrameError(
+          `Placeholder name \`${name}\` is not allowed because it conflicts with auto-generated numeric placeholders. Use a non-numeric name instead.`,
+        )
+      }
+      if (!/^[a-zA-Z_$][a-zA-Z0-9_$]*$/.test(name)) {
+        throw path.buildCodeFrameError(
+          `Placeholder name \`${name}\` is not a valid identifier. Only valid JS identifers are allowed; /^[a-zA-Z_$][a-zA-Z0-9_$]*$/`,
+        )
+      }
```
timofei-iatsenko: as said in the comment, this validation is probably too much restrictive. The `foo-bar` is a perfect js indentifier and i don't see any issue using it.  Actually from the JS point of view we don't have any real restrictions here. The restrictions may come from the `formatElement` function and would it be able to parse it. 

I thik we need to keep the validation here, but it should allow:

```
foo-bar
foo.bar -> need to test will `formatElement` understand it
foo1234
foo_bar
FooBar
```

And this should be blocked: 

```
foo< (will break  `formatElement`  parser)
(spaces and tabulation)
=
/
and so on
```

mogelbrod: I'll relax the requirements a bit to allow `-` and `.` in the identifier then 👍

mogelbrod: Implemented in c52e5c77e4190ed9a9f5f70ebff3aeebab171621

Bertg:
Man, I love this change. Having translations with sometimes more than 5 tags are a PITA. 

Some notes:

* I like the `_t` id, but I forse problems with that. Eg when devs/ai start copy pasting that outside of lingui macros. I would prefer something like `data-lingui-name` which would not break React rules. 
* when we have tags without attributes (eg `strong` or `br`) then repeating the name in the translation string should be fine. When there are attributes I would use a numeric suffix always. 
* I prefer `<a 1>...</a>` over `<a1>...</a1>`. If that is invalid I would even go for `<a id=1>...</a>` or `<a name=1>...</a>`
* with setting a custom name (see above) I would go with `<a id=theCustomNameHere>...</a>`. (Or name instead of id). 
* if we really want to do custom tags, we could still go with `data-lingui-tag`

(Sorry this is a bit messy, but I'm writing this from my phone and don't feel like reformatting)

mogelbrod:
> Man, I love this change. Having translations with sometimes more than 5 tags are a PITA.

Glad to hear it!

> * I like the `_t` id, but I forse problems with that. Eg when devs/ai start copy pasting that outside of lingui macros. I would prefer something like `data-lingui-name` which would not break React rules.
> * ...
> * with setting a custom name (see above) I would go with `<a id=theCustomNameHere>...</a>`. (Or name instead of id).
> * if we really want to do custom tags, we could still go with `data-lingui-tag`

The attribute name is completely configurable (and disabled by default), with `_t` just being a suggestion in the docs. I guess this could be covered in a repo's AGENTS.md or the like, which would be on-par with how other frameworks handle it.

> * when we have tags without attributes (eg `strong` or `br`) then repeating the name in the translation string should be fine. When there are attributes I would use a numeric suffix always.

Same tags with none/identical attributes reuse the same placeholder name and JSX element in the current implementation. I previously appended a incremented suffix for non-unique elements, but as @timofei-iatsenko pointed out [here](https://github.com/lingui/js-lingui/pull/2505#discussion_r3072159338) it's probably better to throw since those cases are likely unintentional and would be better off explicitly named.

> * I prefer `<a 1>...</a>` over `<a1>...</a1>`. If that is invalid I would even go for `<a id=1>...</a>` or `<a name=1>...</a>`

Are you referring to the placeholders in the translation string? My guess is most TMSes don't accept spaces in element names as it's typically a delimeter between tag/attributes, but I haven't double checked. IMO it's better to stick with syntax that's as similar to HTML/JSX as possible for maximum compatibility and understandability.

timofei-iatsenko:
For me the current design looks good.  I propose to merge it as-is and start gathering feedback. I think at the current stage without actually trying to work with it in real projects it's hard to anticipate what works good and what needs polishing.

timofei-iatsenko (approved):
The code itself could be structured a little bit better, but it good for now as-is. It LGTM for me. 

Let's focus on merging the base of this feature and blocking other PRs - the rest of polishing could be provided as separate PRs

mogelbrod:
Great to hear! Happy to clean up and/or refactor in this PR if I get some pointers, as well as following up on any issues that are reported post merge 👍

mogelbrod:
@timofei-iatsenko do you think it's time to update https://github.com/lingui/swc-plugin/pull/207 to mirror this implementation and rebase https://github.com/lingui/js-lingui/pull/2514 against this branch now?

timofei-iatsenko:
@mogelbrod no, wait while this one would be merged to next, and then update https://github.com/lingui/js-lingui/pull/2514 over the main. You don't need to use rebase and force push (and i even encourage you to not doing this) since we will squash and merge the PR anyway. 

> do you think it's time to update https://github.com/lingui/swc-plugin/pull/207

I think so

andrii-bodnar:
@mogelbrod @timofei-iatsenko let's keep this feature in sync with the SWC implementation. I would like to merge them both and then release the current scope as `v6.0.0-next.4` and `v6.0.0-next.3` (SWC plugin)

I hope this will be the last one before the stable v6 release.

mogelbrod:
Great! Working on aligning the SWC plugin as we speak 👍

andrii-bodnar (approved):

PR #2502, opened 2026-04-06
fix: trans values typing and runtime

yslpn:
# Description

This PR narrows `Trans` interpolation values to supported scalar types and fixes the rendering logic to match that contract across `@lingui/core`, `@lingui/react`, and macro typings.

Before this change, some values were accepted by TypeScript but did not behave correctly at runtime in `Trans`. 

This PR makes the typing stricter

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/lingui/js-lingui/blob/main/CONTRIBUTING.md) and [CODE_OF_CONDUCT](https://github.com/lingui/js-lingui/blob/main/CODE_OF_CONDUCT.md) docs
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added the necessary documentation (if appropriate)

PR #2507, opened 2026-04-10
fix(core): add bigint to MessagePlaceholder type

pkasarda:
## Problem

Libraries like Zod type validation bounds as `number | bigint` (e.g. `ZodTooSmallIssue.minimum: number | bigint`). Passing these values directly in `t` template literals causes a TypeScript error:

```
error TS2345: Argument of type 'number | bigint' is not assignable 
to parameter of type 'MessagePlaceholder'.
  Type 'bigint' is not assignable to type 'MessagePlaceholder'.
```

## Solution

Add `bigint` to the `MessagePlaceholder` type in `@lingui/core/macro`:

```diff
 export type MessagePlaceholder =
   | string
   | number
+  | bigint
-  | LabeledExpression<string | number>
+  | LabeledExpression<string | number | bigint>
```

This is consistent with `i18n._()` which already accepts `Values = Record<string, unknown>`.

## Workaround users currently need

```ts
// Zod validation bounds
t`Min. ${Number(min)} characters`  // wrapping in Number()
```

With this fix:
```ts
t`Min. ${min} characters`  // works directly
```

### DIFF
--- a/packages/core/macro/index.d.ts
+++ b/packages/core/macro/index.d.ts
@@ -59,7 +59,8 @@ export type LabeledExpression<T> = Record<string, T>
 export type MessagePlaceholder =
   | string
   | number
-  | LabeledExpression<string | number>
+  | bigint
+  | LabeledExpression<string | number | bigint>
 
 /**
  * Translates a template string using the global I18n instance


timofei-iatsenko:
Related https://github.com/lingui/js-lingui/pull/2502

PR #2510, opened 2026-04-13
feat: solidjs integration

### LINKED ISSUE #2425: Take a closer look for SolidJS integration
Investigate how we can help with Solidjs integration https://github.com/Azq2/js-lingui-solid

I did some changes in the lingui in the past to help reduce duplication but it seems it was not applied https://github.com/lingui/js-lingui/pull/2101#issuecomment-2748271457

the integration looks a bit unmaintained and i think we can take it on ours ownership (in the original repo, or move it to the lingui monorepo)

Huliiiiii:
# Description

- Implemented `@lingui/solid` based on the `@lingui/react` design
- Add the `jsxRuntime` option in `@lingui/conf`
- Added solid-specific logic to `babel-plugin-lingui-macro`

## Differences from React

1. Components are compiled into:
    ```typescript
    {
        0: (props) => <elm {...props}>
    }
    ```
    instead of:
    ```typescript
    {
        0: <elm {...props}>
    }
    ```

2. Reusing the same node multiple times in a translation string (for example, "<0/> <0/>") is not supported.
  
    For example, this source code:

    ```tsx
    <Trans>
     Read the <a href="/docs">docs</a>.
    </Trans>
    ```
    produces a message like:
    `Read the <0>docs</0>.`
 
    A translation such as this is not supported:
  
    `Read the <0>docs</0> or the <0>guide</0>.`
    
    It will render as `Read the  or the <a href="/docs">guide</a>.`

    This is because HTML elements cannot be cloned trivially. In this implementation, the same element is reused by reference instead of copies.
   
    The same limitation also applies when a JSX element stored in a variable is referenced multiple times.
    
    For example:
    
    ```tsx
    const link = <a href="/docs">docs</a>
    
    <Trans>
      {link} or {link}
    </Trans>
    ````
    
    The output will be similar to:
    
    `or <a href="/docs">docs</a>`
    
    Using a component instead avoids this problem:
    
    ```tsx
    const link = () => <a href="/docs">docs</a>
    ```

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Closes #2425

## Checklist


timofei-iatsenko:
@Huliiiiii please add description for the PR, with details: 

- What was implentend
- What decision was taken
- What decision considered but wasn't implemented
- What the diffrences you had to implement to support solid. 

That will make reviewing and merging the PR easier and faster. Thanks!

packages/babel-plugin-extract-messages/src/index.ts
```diff
@@ -202,10 +202,12 @@ export default function ({ types: t }: { types: BabelTypes }): PluginObj {
...
-      path
-        .get("openingElement")
-        .get("name")
-        .referencesImport("@lingui/react", "Trans")
+      ["@lingui/react", "@lingui/solid"].some((moduleName) =>
+        path
+          .get("openingElement")
+          .get("name")
+          .referencesImport(moduleName, "Trans"),
+      )
```
timofei-iatsenko: Todo for myself: check if we can drop it entirely or make configurable.  Don't want to have list of packages hardcoded

packages/babel-plugin-lingui-macro/test/jsx-trans.test.ts
```diff
@@ -364,6 +364,41 @@ macroTester({
...
+        import { Trans as _Trans } from '@lingui/solid';
+        <_Trans
+          {
+            /*i18n*/
+            ...{
+              id: 'lqG909',
+              message: 'Hello <0>docs</0>.',
+              components: {
+                0: (props) => <a href="/docs" {...props} />,
+              },
```
timofei-iatsenko: could you elaborate a bit onwhy this needed for Solid? 

I wonder could we avoid branching and generate for both react and solid the same way? From the first glance it seems react would also work with no issues with this syntax

packages/conf/src/__snapshots__/getConfig.test.ts.snap
```diff
@@ -33,6 +33,7 @@ exports[`getConfig > default config and snapshots > should return default config
     ],
     "jsxPackage": [
       "@lingui/react/macro",
+      "@lingui/solid/macro",
```
timofei-iatsenko: this would be added from the `defineConfig` from the solid package, isn't it? The default code should not be touched

packages/solid/package.json
```diff
@@ -0,0 +1,91 @@
+{
+  "name": "@lingui/solid",
+  "version": "6.0.0-next.3",
+  "sideEffects": false,
+  "description": "SolidJS bindings for Lingui with Trans components, providers, and compile-time macros",
+  "type": "module",
+  "author": {
+    "name": "Tomáš Ehrlich",
+    "email": "tomas.ehrlich@gmail.com"
+  },
```
timofei-iatsenko, line 10: You can put your name here

packages/babel-plugin-lingui-macro/test/jsx-trans.test.ts
Huliiiiii: React uses VDOM, but Solid eagerly creates real DOM nodes. So preserve laziness manually is needed. Otherwise, for a component like this, there’s no correct way to insert children at the right position (using the approach in #2101).

```
const Comp = (props) => <div><div>before</div>{props.children}<div>after</div></div>
```

Maybe there’s a better solution, but I can’t think of one at the moment.

Huliiiiii:
I’ve updated the description.

packages/babel-plugin-lingui-macro/test/jsx-trans.test.ts
timofei-iatsenko: Could you drop this `expected` section in favour of autogenerated snapshot.

timofei-iatsenko (changes requested):

timofei-iatsenko:
> Reusing the same node multiple times in a translation string (e.g. "<0/> <0/>") is not supported.
This is because html elements cannot be trivially cloned.

Could you elaborate on this? What this mean for a user? Could you write few examples what would work - what is not with possible workaround

Huliiiiii:
> > Reusing the same node multiple times in a translation string (e.g. "<0/> <0/>") is not supported.
> > This is because html elements cannot be trivially cloned.
> 
> Could you elaborate on this? What this mean for a user? Could you write few examples what would work - what is not with possible workaround

I've added some examples to the description. Would that be enough?

timofei-iatsenko:
@Huliiiiii

> 
> I've added some examples to the description. Would that be enough?


What happened with the following message? 

```jsx
<Trans>Hello <strong>user</strong> from planet <strong>Earth</strong>
```

Current macro will dedupe `<strong>` into one element like so:


```jsx
{
    0: <strong>
}
```

`Hello <0>user</0> from planet <0>Earth</0>`

UPD: 

I reread your description one more time, and it seems i didn't get it correctly. Correct me if'm wrong - this issue you described 

> Reusing the same node multiple times in a translation string (for example, "<0/> <0/>") is not supported.

You already solved with `() => <jsx>` syntax, so it's supported correctly, right?

Huliiiiii:
> @Huliiiiii
> 
> > I've added some examples to the description. Would that be enough?
> 
> What happened with the following message?
> 
> ```js
> <Trans>Hello <strong>user</strong> from planet <strong>Earth</strong>
> ```
> 
> Current macro will dedupe `\<strong\>` into one element like so:
> 
> ```js
> {
>     0: <strong>
> }
> ```
> 
> `Hello <0>user</0> from planet <0>Earth</0>`
> 
> UPD:
> 
> I reread your description one more time, and it seems i didn't get it correctly. Correct me if'm wrong - this issue you described
> 
> > Reusing the same node multiple times in a translation string (for example, "<0/> <0/>") is not supported.
> 
> You already solved with `() => <jsx>` syntax, so it's supported correctly, right?

`() => <jsx>` will create a new instance.

Deduplication should works correctly if the current tests already cover that.

The core difference is that solid can only use references, but react uses clones/copies.

Huliiiiii:
Tests failed because `@babel/plugin-syntax-jsx` is missing. Should I fix this in this PR as well?

andrii-bodnar:
@Huliiiiii thank you for the contribution!

Improving Lingui's compatibility with SolidJS and the broader ecosystem (Astro, Vue, Svelte) is a major focus for the community at the moment.

With v6 nearly finalized, our plan is to ship the current milestone first and then introduce dedicated SolidJS support in a subsequent release.

In addition to the feature itself, it would be great to have a dedicated example project featuring the integration, a documentation page (similarly to the existing [React Apps Internationalization](https://lingui.dev/tutorials/react)) and a blog post (optionally). Please let us know your thoughts!

timofei-iatsenko:
@Huliiiiii we reached v6 milestone. We still interersted in this integration. Could you update the branch to the latest main? (please note, the main branch is renamed from what was `next` previously, you may need some additional fixes in git). You don't need to keep a git history clean - we will squash all commits into one while merging anyway. 

Thanks!

timofei-iatsenko:
@Huliiiiii i'm going to jump in and help with this PR, so please don't rebase/force push. Thanks!

timofei-iatsenko:
FYI: i'm working on issue: 

> Error: Error: Client-only API called on the server side. Run client-only code in onMount, or conditionally run client-only component with <Show>.
 ❯ notSup ../../node_modules/solid-js/web/dist/server.js:760:9

The issue related to a bug in the Vitest, i will update vitest in a separate PR and then update this branch.

Huliiiiii:
@timofei-iatsenko I'm unable to reproduce the CI error locally.

timofei-iatsenko:
aghr... beautiful windows. Did you try to run it on windows machine?

Huliiiiii:
@timofei-iatsenko The CI has been fixed.

timofei-iatsenko (approved):

packages/babel-plugin-extract-messages/src/index.ts
timofei-iatsenko: @Huliiiiii i think we can use `macro.runtimeConfigModule.Trans` option here and avoid hardcode, could you make it change and at a simple test testing that this property is respected? Thanks!

timofei-iatsenko (changes requested):

timofei-iatsenko:
@Huliiiiii could you please not force push in this branch. We don't care about the history in the feature bunches since they would be squahed and merged any way. 

I'm going to push to your branch to help with CI issues.

packages/babel-plugin-extract-messages/src/index.ts
Huliiiiii: Done.

Huliiiiii:
Sorry, I accidentally force-pushed again.

Huliiiiii:
I’m unable to reproduce the timeout issue locally. Maybe it’s an issue with the action runner?

packages/solid/package.json
```diff
@@ -0,0 +1,92 @@
+{
+  "name": "@lingui/solid",
+  "version": "6.0.0-next.3",
```
andrii-bodnar: Please update to the latest version

andrii-bodnar (approved):
@Huliiiiii thank you for the contribution!

PR #2514, opened 2026-04-14
feat: add lingui-set/lingui-reset comment directives

### LINKED ISSUE #2357: [macros] Configurable default `context` value generation for macros
### Problem Description

My organization is currently evaluating switching from i18next to another i18n solution, and lingui is currently the most promising candidate.

After converting most of our existing i18next usage to lingui macros using a custom codemod we ended up with many reused translation strings across multiple files. We'd like to avoid string reuse, which can be accomplished by specifying `context`. It's unfortunately not immediately obvious to a developer when they'd be expected to provide `context`, as it's to my knowledge only possible to see if a translation is reused by manually checking the po file after running the extract command. This implies that the only safe approach is to manually specify `context` everywhere, which would essentially nullify the DX improvement which lingui provides by using key-less translations.

Also, specifying `context` for `<Trans>` macros is simple enough but I'm not aware of any way to do so for the `t` macro. **Is it possible to specify `context` for `t` macros without having to switch to `i18n._()`?**

### Proposed Solution

If it was possible to automatically generate default `context` values for the `t`/`<Trans>` macros the above limitations could be avoided. Two possible solution I can think of:

1. Pass a `(filePath: string) => string | undefined` function in the lingui config that returns the default context for a given file path. If a JS function isn't possible I'd suggest a regex approach like `^.*/src/(?<context>.+?)(\.[jt]sx?)?$` so that leading/trailing noise can be omitted.
2. Allow custom pragmas/directives to control lingui macro conversion on a per-file basis, example:
   ```
   /* @lingui context=myContext */
   ```
   This could apply until the next `@lingui` comment in the same file.
   This would be more repetitive and easier to forger than (1) since devs would need to remember to add it, although it can be forced via eslint rule or similar.

### Alternatives Considered

- Manually specifying `context` everywhere, which would essentially nullify the DX improvement which lingui provides by using key-less translations.

### Additional Context

Looking forward to hearing your thoughts!

mogelbrod:
# Description

Adds support for `lingui-set` and `lingui-reset` comment directives that provide default `context`, `comment` values to subsequent lingui macro calls in a file. This avoids repeating the same `context` or `comment` on every macro call when many messages in a file share the same value. They also expose the ability to set a common `idPrefix` on subsequent macro calls, optionally configured to be opt-in rather than opt-out via a new `macro.idPrefixLeader` setting.

Both block and line comments are supported:

```ts
/* lingui-set context="checkout" comment="Checkout page strings" */
// lingui-set context="checkout" idPrefix="checkout."
```

#### `lingui-set` and `lingui-reset`

- `lingui-set` accumulates: each `lingui-set` merges its values with all previously accumulated directive values.
- `lingui-reset` clears all accumulated state, then optionally sets new values.
- Setting a key to an empty string (`context=""`) unsets that specific key while keeping others intact.
- If a macro call already has an explicit `context` or `comment` prop, the directive value for that prop is ignored.
- Directive `context` still affects hash-based ID generation in production, even when the context property is stripped from the output.

#### Examples

```ts
import { t } from "@lingui/core/macro"

// lingui-set context="dashboard"
const title = t`Welcome`       // context: "dashboard"
const subtitle = t`Overview`   // context: "dashboard"

// lingui-set comment="user prefs"
const heading = t`Preferences` // context: "dashboard", comment: "user prefs"

// lingui-reset
const other = t`Standalone`    // no context, no comment (cleared)
```

#### `idPrefix` and `idPrefixLeader`

`idPrefix` only applies to explicitly provided message IDs. Auto-generated hash IDs are not affected.

When `macro.idPrefixLeader` is set in the Lingui config (e.g. `"."`), `idPrefix` is only prepended to explicit IDs that start with the leader string. IDs without the leader are left unchanged.

```ts
/* lingui-set idPrefix="module" */
const msg = t({ id: ".greeting", message: "Hello" }) // id: "module.greeting"
const msg2 = t({ id: "fixed", message: "Hello" })    // id: "fixed"
```

## Types of changes

- [x] Documentation update
- [x] Examples update

Fixes #2357

## Checklist

## Disclaimer

Much of this PR was generated by Copilot using the Claude Opus 4.6 model with training/sharing disabled.


Bertg:
Just adding my 2 cents here. 

From the example:

```js
import { t } from "@lingui/core/macro"

/* @lingui context="dashboard" */
const title = t`Welcome` // context: "dashboard"
const subtitle = t({ comment: "hello", message: "Overview" }) // context: "dashboard", comment: "hello"

/* @lingui comment="prefix example" idPrefix="myPrefix." */
const heading = t({ id: "heading", message: "Preferences" }) // id="myPrefix.heading" comment="prefix example"
```

However, I would expect the last heading details to be different.

```js
/* @lingui context="dashboard" */

// ...

/* @lingui comment="prefix example" idPrefix="myPrefix." */
const heading = t({ id: "heading", message: "Preferences" }) // id="myPrefix.heading" comment="prefix example" context="dashboard"
```

My expectation would be that each `@lingui` directive merges with the previous one, not reset the previous one. I'm not saying that my expectation is the right one, but that expectations on how this works my be different.

I think form a DX experience it is important to get this one right, and that it should be well documented.

mogelbrod:
Thanks for the feedback @Bertg! I was also considering that behaviour, and am somewhat on the fence myself as there are pros and cons to both:

### Not merging directives
✅ Easier to follow as a reader/dev as only the nearest directive is considered
✅ Modifying an existing directive does not risk affecting anything outside of its "scope"
⚠️ Not possible to inherit subset of options from earlier directives, instead being required to repeat these in each directive of the file

### Merging directives in the same file
✅ Familiar pattern, common with for example `/* eslint-disable ... */`
✅ Possible to reduce repetition if multiple directives are used in the same file
⚠️ Modifying a single directive can easily mess up others - requires looking through each subsequent directive and possibly also the compiled output

I'd love to hear more thoughts on this!

timofei-iatsenko:
Is there any way to reset the directive? Like that: 

```js
/* @lingui context="dashboard" */
const title = t`Welcome` // context: "dashboard"

/* @lingui context="" */
const title = t`Welcome` // context: undefined

```

mogelbrod:
> Is there any way to reset the directive?

Good idea. I just pushed 1a070da196793f6a16a3f1edf5279dfb27f71baf which changes the behavior of the directive to merge with previous directive(s) by default, with a new `reset` param added to disable this. What do you think @timofei-iatsenko @Bertg?

```ts
// @lingui context="first context" comment="comment"
const m1 = msg`1` // context: "first context" comment="comment"
const m2 = msg({ context: "context override", message: "2" }) // context: "context override" comment="comment"

// @lingui context="second context"
const m3 = msg`3` // context: "second context" comment="comment"

// @lingui reset context="third context"
const m4 = msg`4` // context="third context"

// @lingui reset
const m5 = msg`5` // empty
```

timofei-iatsenko:
@mogelbrod i'm not sure this specific `reset` keyword. Feels unnatural, just simple `comment=""` (to empty string) would not work? Or you did want to reset all to default in one shot without listing what was specified

mogelbrod:
Side note: This new `@lingui` comment directive also solves another DX limitation I encountered with lookup tables. I was originally planning on opening another issue + PR to add a way to easily create lookup tables via something like:
```ts
const lookupTable = defineMessageTable({
  idPrefix: 'my.lookup.table.',
  context: 'some context',
}, {
  a: 'translation for "a"',
  b: 'translation for "b"',
})
```

This can now be achieved using the directive:
```ts
// @lingui idPrefix="my.lookup.table." context="some context"
const lookupTable = {
  a: msg`translation for "a"`,
  b: msg`translation for "b"`,
} as const
// @lingui reset
```

mogelbrod:
> @mogelbrod i'm not sure this specific `reset` keyword. Feels unnatural, just simple `comment=""` (to empty string) would not work? Or you did want to reset all to default in one shot without listing what was specified

Yes - I wanted a way to reset all values without having to write each one out. It also reduces the risk for bugs that could otherwise happen if a dev adds a param to an earlier line without also modifying subsequent line(s).

packages/babel-plugin-lingui-macro/test/lingui-directive.test.ts
```diff
@@ -0,0 +1,493 @@
...
+  })
+
+  it("should return null for non-directive comments", () => {
+    expect(parseLinguiDirective(" some comment ")).toBeNull()
+    expect(parseLinguiDirective(" i18n ")).toBeNull()
+  })
+
+  it("should return null for @lingui with no valid params", () => {
+    expect(parseLinguiDirective(" @lingui ")).toBeNull()
+    expect(parseLinguiDirective(' @lingui unknown="value" ')).toBeNull()
```
timofei-iatsenko: Would it be beeter to throw or warn if the incorrect syntax is used? Currently the error would be just swallowed

packages/babel-plugin-lingui-macro/test/lingui-directive.test.ts
```diff
@@ -0,0 +1,493 @@
...
+
+  it("should omit empty string values", () => {
+    expect(parseLinguiDirective(' @lingui context="" ')).toEqual({
+      reset: false,
+      values: {},
+    })
+    expect(
+      parseLinguiDirective(' @lingui context="" comment="note" '),
+    ).toEqual({ reset: false, values: { comment: { text: "note" } } })
+  })
```
timofei-iatsenko: May be it should not be omit, but instead should reset?

timofei-iatsenko:
> This can now be achieved using the directive:
> 
> ```ts
> // @lingui idPrefix="my.lookup.table." context="some context"
> const lookupTable = {
>   a: msg`translation for "a"`,
>   b: msg`translation for "b"`,
> } as const
> // @lingui reset
> ```

Not sure i understood the example, didn't you say that `idPrefix` would affect only explicitly set ids?

mogelbrod:
> Not sure i understood the example, didn't you say that `idPrefix` would affect only explicitly set ids?

You are correct, I forgot to update the copied source. It should look like this:
```ts
// @lingui idPrefix="my.lookup.table." context="some context"
const lookupTable = {
  a: msg({ id: 'a', message: `translation for "a"` }),
  a: msg({ id: 'b', message: `translation for "b"` }),
} as const
// @lingui reset
```

packages/babel-plugin-lingui-macro/test/lingui-directive.test.ts
mogelbrod: Good point, will add error handling!

timofei-iatsenko:
let's bring this feature to a bigger forum. @yslpn could you also take a look here and provide a feedback? 

regarding failing workflows - let's focus and finish this one first https://github.com/lingui/js-lingui/pull/2505 and then come back to polish this feature.

Bertg:
I like where this is going. But there is a bit of asymmetry with the `reset` option. What if we do this:

```
// @lingui set context="hello"
// @lingui reset 
```

Using `set` is explicit. And allows to expand with more commands in the future.

mogelbrod:
> I like where this is going. But there is a bit of asymmetry with the `reset` option. What if we do this:
> 
> ```
> // @lingui set context="hello"
> // @lingui reset 
> ```
> 
> Using `set` is explicit. And allows to expand with more commands in the future.

I think that's a good idea fot future proofing. @timofei-iatsenko @yslpn if I get another 👍  I'll update this implementation 😃

packages/babel-plugin-lingui-macro/src/macroJsx.ts
```diff
@@ -351,18 +364,82 @@ export class MacroJSX {
...
+    }
+
+    if (!name) {
+      name = String(this.ctx.elementIndex())
+      elementsTracking.set(name, node)
+    } else {
+      const existingElement = elementsTracking.get(name)
+
+      if (existingElement) {
+        const existingAttrs = existingElement.openingElement.attributes
```
yslpn, line 430:
```tsx
<Trans>
  <a _t="link" className="cta">Docs</a>
  <button _t="link" className="cta">Submit</button>
</Trans>;
```

```tsx
<Trans
  {...{
    message: "<link>Docs</link><link>Submit</link>",
    components: {
      link: <button className="cta" />
    }
  }}
/>;
```

packages/babel-plugin-lingui-macro/src/macroJsx.ts
```diff
@@ -351,18 +364,82 @@ export class MacroJSX {
...
+        const openingAttrs = node.openingElement.attributes
+        if (
+          existingAttrs.length !== openingAttrs.length ||
+          !existingAttrs.every((a) =>
+            openingAttrs.some((b) => this.types.isNodesEquivalent(a, b)),
```
yslpn: We've been through this before)

```ts
const props = {
   href: "/from-props"
   // href": "/a"
}
```

```tsx
<Trans>
  <a _t="link" {...props} href="/a">First</a>
  <a _t="link" href="/a" {...props}>Second</a>
</Trans>
```

```tsx
message: "<link>First</link><link>Second</link>",
components: {
  link: <a href="/a" {...props} />
}
```

yslpn:
> This PR is currently stacked on top of #2505. Let me know if they need to be separated.

@mogelbrod we need to change the target branch. Right? 

And important: here I see unfixed errors from MR https://github.com/lingui/js-lingui/pull/2505

timofei-iatsenko:
@yspln 

> > This PR is currently stacked on top of #2505. Let me know if they need to be separated.
> 
> @mogelbrod we need to change the target branch. Right?
> 
> And important: here I see unfixed errors from MR #2505

Please focus on the feature design, the issues related to #2505 would be fixed once #2505 would be merged.

timofei-iatsenko:
> I like where this is going. But there is a bit of asymmetry with the `reset` option. What if we do this:
> 
> ```
> // @lingui set context="hello"
> // @lingui reset 
> ```
> 
> Using `set` is explicit. And allows to expand with more commands in the future.

I'm not sure about this. The explicitly is fine, but if this would be used a lot day to day, i prefer shortiness over explicitness.  And `@lingui context="hello"` looks totally understandable for me. 

But if we want to go this direction - may be let's follow the eslint approach, something like 

```
/// {directive} {parameters}
```

So explicit set and reset directives as: 

```
 // @lingui-set context="hello"
 // @lingui-reset 
```

But as i said, i'm ok with the current design, it seems quite ergonomic and straight forward to me.

packages/babel-plugin-lingui-macro/src/macroJsx.ts
mogelbrod: These are addressed in the parent PR - I'll update the base branch for this PR and rebase against it once it's code has stabilized 😃

packages/babel-plugin-lingui-macro/src/macroJsx.ts
mogelbrod: These are addressed in the parent PR - I'll update the base branch for this PR and rebase against it once it's code has stabilized 😃

mogelbrod:
Had to rebase and force-push to keep each commit atomic.
I liked the idea of separating `lingui-set` and `lingui-reset`, as it pairs well with the existing `lingui-extract-ignore` directive. Working on implementing this now in a follow up commit. No more force pushes after this :)

mogelbrod:
02d88b649c358546ce3b4901e842f0c2ad238556 splits `@lingui` into `lingui-set` and `lingui-reset` directives - let me know what you think!

timofei-iatsenko (approved):

mogelbrod:
I realized that we could use a way to opt out of the `idPrefix` when configured, so that files that almost exclusively rely on the `idPrefix` save for a few IDs don't have to reset it. This could be achieved by adding a configurable `idPrefixLeader?: string` (`idPrefixAppliesToPrefix`?) setting that limits what IDs to prefix. Example with `idPrefixLeader: '.'` set:

```tsx
// lingui-set idPrefix="checkout"

<Trans id=".usesPrefix">Hello</Trans>
<Trans id=".usesPrefix.with.subpath">Hello</Trans>
<Trans id="unprefixed.key">Hello</Trans>
<Trans id="unprefixed">Hello</Trans>
```

I'll take a stab at implementing this right away unless somebody objects to it 🙏

mogelbrod:
I've pushed 1abe659080e8e60f3b790bfffc9fe4f68e863d3f which adds the optional `macro.idPrefixLeader` setting, as well as updated the PR body to reflect recent changes. Let me know what you think!

mogelbrod:
@timofei-iatsenko @yslpn: Anything I can do to help land this? This is currently somewhat of a blocker for my org adopting lingui as we'd like to try using generated keys for the added DX, but for us that depends on being able to use `// lingui-set context="..."` to provide stable context values without having to resort to the `t({ context: '...', message: '...' })` syntax. With that said the recently added `idPrefixLeader` should still be super helpful to anybody using explicit keys (including us if we decide to later switch).

andrii-bodnar (approved):

mogelbrod:
@timofei-iatsenko @andrii-bodnar: Found a bug where `idPrefix` wasn't applied if `idPrefixLeader` hadn't been configured. When adding tests for this Copilot also found an issue with the global config instance in packages/babel-plugin-lingui-macro/src/index.ts:

> This breaks correctness when the plugin is used for multiple transforms in the same process with different configs. A transform that sets `macro.idPrefixLeader` can leak that setting into a later transform that should not have it, which in turn caused problems in the `idPrefix`/`idPrefixLeader` tests. The fix makes config per-transform instead of global: resolve it once in `Program.enter`, store it for that run, and pass it explicitly into helpers like `getMacroImports` and `isLinguiIdentifier`.

I've pushed some changes to packages/babel-plugin-lingui-macro/src/index.ts that addresses these.

mogelbrod:
Submitted SWC implementation to https://github.com/lingui/swc-plugin/pull/211

timofei-iatsenko:
LGTM! Could be merged. We can do an announcement and mark this feature as experemental and ask for feedback in the specific discussion in the github.

Issue #2506, opened 2026-04-09
`lingui compile --strict` semantics do not match `extract` "Missing"


yslpn:
## Description

There is an inconsistency between:

- the `Missing` counts reported by `lingui extract`
- the documented behavior of [`lingui compile --strict`](https://lingui.dev/ref/cli#compile-strict)
- the actual runtime behavior when [`fallbackLocales`](https://lingui.dev/ref/conf#fallbacklocales) are configured

The CLI docs currently say that [`--strict`](https://lingui.dev/ref/cli#compile-strict) will:

> Fail if a catalog has missing translations.

At the same time, the config docs say that [`fallbackLocales`](https://lingui.dev/ref/conf#fallbacklocales) are used when a translation for a locale is missing.

Today these two concepts do not line up clearly:

- `lingui extract` reports `Missing` based on raw catalog entries with empty translations
- `lingui compile --strict` appears to validate the resolved result after fallbacks are applied

Because of that, a locale can still show `Missing > 0` in extract stats, while `lingui compile --strict` exits successfully.

This is especially confusing in CI for teams that:

- use `fallbackLocales` at runtime
- still want to fail builds until each locale catalog is fully translated

Related but distinct from [#2195](https://github.com/lingui/js-lingui/issues/2195): that issue asks for a dedicated check command without writing files. This issue is about the current semantic mismatch between documented strict-mode behavior and what the CLI actually validates.

## Verifications

- [x] I've checked [the docs](https://lingui.dev) and this isn't covered there.
- [x] I've searched existing issues on [GitHub](https://github.com/lingui/js-lingui/issues).

## Reproduction Steps

Documentation references:

- CLI strict mode: https://lingui.dev/ref/cli#compile-strict
- `fallbackLocales`: https://lingui.dev/ref/conf#fallbacklocales
- `sourceLocale`: https://lingui.dev/ref/conf#sourcelocale

`lingui.config.js`

```js
export default {
  locales: ["en-US", "en-GB"],
  sourceLocale: "en-US",
  fallbackLocales: {
    default: "en-US",
  },
  catalogs: [
    {
      path: "<rootDir>/{locale}",
      include: ["<rootDir>"],
      exclude: [],
    },
  ],
}
```

`en-US.po`

```po
msgid "Hello"
msgstr "Hello"
```

`en-GB.po`

```po
msgid "Hello"
msgstr ""
```

Run:

```sh
lingui extract
```

Result:

```text
Catalog statistics for {locale}:
[...]

Issue #23, opened 2017-06-18
Allow built-in components in render prop


tricoder42:
Just a small nitpick.

It's possible to use React.Component to render translation: `<Trans render={Component}>...</Trans>`, but built-ins can be passed as elements only: `<Trans render={<h1 />}>...</Trans>`.

Proposal: `<Trans render="h1">...</Trans>`

It's like 4 chars difference, but nice to have 😸

PR #2515, opened 2026-04-16
feat(cli): add lingui check sync | missing commands

### LINKED ISSUE #2195: Add a command to check for missing translations
### Problem Description

We would like to have a command that checks for missing translations, so we do not ship those builds in production.

`lingui compile --strict` does this, but when you use the Vite plugin, it will also compile the files, which is not needed.

### Proposed Solution

A dedicated `lingui check` (or similar) command that does what `lingui compile --strict` does, but without writing the files, would be very useful.

### Alternatives Considered

`Run `lingui compile --strict` then remove the resulting output files, but this is not very handy

### Additional Context

_No response_

yslpn:
# Description

This PR adds a dedicated `lingui check` command for validating catalogs without writing output files.

It is based on #2506 and should be merged after that PR.

It also closes #2195 by adding a CI-friendly way to check catalogs without running compilation.

Included in this PR:

- `lingui check missing` to fail when translations are still missing after `fallbackLocales` are applied, matching `lingui compile --strict`
- `lingui check missing --mode catalog` to fail on catalog-level missing translations before `fallbackLocales` are applied
- `lingui check sync` to fail when catalogs are out of sync with `lingui extract`
- Shared missing-translation handling that ignores obsolete catalog entries in both `resolved` and `catalog` modes

In `packages/format-po/src/po.ts`, PO serialization was updated to preserve meaningful existing header state, merge custom headers deterministically, and drop empty default gettext headers. This makes serialization more stable and avoids noisy diffs, which is important for `check sync`.

## Types of changes

- [x] Documentation update
- [ ] Examples update

Fixes #2195
Related to #2506

## Checklist


packages/cli/src/api/formats/formatterWrapper.ts
```diff
@@ -16,19 +16,31 @@ export class FormatterWrapper {
...
-  ): Promise<void> {
-    const content = await this.f.serialize(catalog, {
+    existing?: string,
+  ): Promise<string> {
+    return await this.f.serialize(catalog, {
       locale,
       sourceLocale: this.sourceLocale,
-      existing: await readFile(filename),
+      existing: existing ?? (await readFile(filename)),
       filename,
```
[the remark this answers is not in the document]
yslpn: Fixed

packages/cli/src/lingui-check.ts
```diff
@@ -0,0 +1,202 @@
...
+  return command
+    .option("--config <path>", "Path to the config file")
+    .option(
+      "--locale <locale, [...]>",
+      "Only check the specified locales",
+      parseLocales,
+    )
+    .option(
+      "--workers <n>",
+      "Number of worker threads to use (default: CPU count - 1, capped at 8). Pass `--workers 1` to disable worker threads and run everything in a single process",
```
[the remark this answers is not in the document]
yslpn: Fixed

packages/cli/src/api/check/sync.ts
```diff
@@ -0,0 +1,132 @@
...
+
+      if (existing === expected) {
+        return undefined
+      }
+
+      return {
+        code: "catalog_out_of_sync",
+        message: existing
+          ? `Catalog ${catalogPath} is out of sync with extract output`
+          : `Catalog ${catalogPath} is missing and would be created by extract`,
```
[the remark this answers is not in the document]
yslpn: Fixed

andrii-bodnar:
Hi @yslpn - apologies this one sat without much attention for a while, and thanks for your patience! I'd like to pick it back up and make this PR the main home for the #2506 / #2195 work.

`lingui check missing` (resolved) and `lingui check missing --mode catalog` (raw catalog) already cover exactly the mismatch #2506 described, in a CI-friendly command that doesn't write output - so once this lands, I'm not sure a separate `compile --fail-on-missing` flag (over in #2512) really earns its place; it'd be a second public way to run the same validation. Curious what you think.

We're happy with the shape overall - I'll do a closer pass on the details once it's rebased. Thanks again!

andrii-bodnar:
One more thing, looking ahead: it'd be great for `lingui check` to grow into the home for all catalog validation, so I'd like to keep the design open to more checks beyond `sync` / `missing`. A few I have in mind for follow-ups:

- ICU syntax validity
- placeholder consistency between source and translation
- plural-form completeness
- etc.

And directionally I'm picturing:

- a bare `lingui check` that runs all configured checks
- per-check configuration in `lingui.config` (enabling checks, and severity like error vs. warning)

None of this needs to land in this PR - I'd just like to make sure the check registry stays a clean extension point, so adding a new check later is essentially a single registration, and that the finding shape has room to carry things like severity. WDYT?

yslpn:
> Hi @yslpn - apologies this one sat without much attention for a while, and thanks for your patience!

Don't worry, this is asynchronous work. I understand that it can be difficult to review all PRs, especially if they are large and require further development to ensure the decisions made are scalable in the future.

> I'd like to pick it back up and make this PR the main home for the #2506 / #2195 work.

Okay)

> `lingui check missing` (resolved) and `lingui check missing --mode catalog` (raw catalog) already cover exactly the mismatch #2506 described, in a CI-friendly command that doesn't write output - so once this lands, I'm not sure a separate `compile --fail-on-missing` flag (over in #2512) really earns its place; it'd be a second public way to run the same validation. Curious what you think.

Agree. I think I initially tried to leave this in place to remove the API asymmetry between compile and lingui.config options. But I think it's acceptable to have different settings. This is justified by the fact that bundler performs compilation within its own pipeline.

> a bare `lingui check` that runs all configured checks

I don't like this idea; I'd like to separate these checks at the cli and code level. This will be easier to extend and maintain. Any addition of a new check to a single check could be considered a breaking change. It would be difficult to develop.

But I see a possible option for a unified `lingui check`.

```ts
// lingui.config.ts
checks: {
  sync: {
    severity: "error",
  },
  missing: {
    mode: "catalog",
    severity: "error",
  },
  icu: {
    severity: "warning",
  },
}
```

lingui check only runs checks enabled in the configuration; But we also have `failOnMissing`, which doesn't really fit into this direction now.

> We're happy with the shape overall - I'll do a closer pass on the details once it's rebased. 

I'll come back to this this week and fix the conflicts.

andrii-bodnar:
@yslpn Thanks! Agreed on keeping checks separated and a bare `lingui check` would just be a thin runner for the enabled ones.

For the config, I looked at how Biome does it and I'd suggest starting even simpler: plain severity strings (`"off" | "on" | "warn" | "error"`, where `"on"` enables a check at its recommended severity):

```ts
check: {
  rules: {
    missing: "error",
    sync: "off",
    "icu-translation": "on",
  },
}
```

Options like `--mode` can stay CLI flags for now - `missing` is the only check that has one, so config options feel premature. If we ever need them, Biome's `{ level, options }` object form can be added later without breaking the string values.

On check additions not being breaking - 100% agree, I'd just solve it with defaults instead of config-only run-all: `sync` and `missing` stay enabled by default at `error` (as in this PR), while future checks ship disabled and only run once you turn them `"on"`. Same guarantee, but a bare `lingui check` still works with zero config.

During a major releases, we can reconsider some checks and enable/disable them.

What do you think?

yslpn:
@andrii-bodnar Sounds good — agreed on the `"off"/"on"/"warn"/"error"` shape, keeping `--mode` CLI-only for now, and the defaults strategy. The check registry is already a clean single-registration extension point, and since `CheckFinding`/`CheckResult` aren't part of the public API, adding a `severity` field later (whenever `check.rules` actually lands) is a trivial, non-breaking addition — no need to pre-add it now.

packages/cli/src/api/check/sync.ts
```diff
@@ -0,0 +1,129 @@
...
+export const syncCheck: CheckDefinition = {
+  name: "sync",
+  description:
+    "Verify that locale catalogs are already synchronized with what lingui extract would write.",
+  cli: {
+    options: [
+      {
+        name: "clean",
+        runOption: "clean",
+        description: "Mirror extract --clean behavior when running sync check",
```
andrii-bodnar, line 85: Is it possible to describe how it works without referring to the extract command? This would require checking the "--clean" option to understand it

packages/cli/src/api/findings.ts
andrii-bodnar, line 1: Do we really need this as a separate file? It seems like it's only used in types.ts

packages/cli/src/lingui-check.ts
```diff
@@ -0,0 +1,226 @@
...
+  return command
+    .option("--config <path>", "Path to the config file")
+    .option(
+      "--locale <locale, [...]>",
+      "Only check the specified locales",
+      parseLocales,
+    )
+    .option(
+      "--workers <n>",
+      "Number of worker threads to use (default: based on CPU count, capped at 8, with special handling for small machines). Pass `--workers 1` to disable worker threads and run everything in a single process",
```
andrii-bodnar, line 113: One small wording suggestion for the `--workers` help text: "with special handling for small machines" is accurate but doesn't tell the user what actually happens. The real rule is short enough to state directly:

```
Number of worker threads to use (default: CPU count - 1, capped at 8; on 1-2 core machines, all cores). Pass `--workers 1` to disable worker threads and run everything in a single process
```

Would be nice to align docs with the same wording. What do you think?

packages/format-po/src/po.ts
```diff
@@ -146,19 +146,138 @@ function isGeneratedId(id: string, message: MessageType): boolean {
...
+    formatPotCreationDate(new Date())
+  nextHeaders["MIME-Version"] = "1.0"
+  nextHeaders["Content-Type"] = "text/plain; charset=utf-8"
+  nextHeaders["Content-Transfer-Encoding"] = "8bit"
+  nextHeaders["X-Generator"] = "@lingui/cli"
+
+  if (language) {
+    nextHeaders.Language = language
+  }
+
```
andrii-bodnar, line 213: Do we need to normalize headers on existing files at all? For `check sync` it should be enough that serialization is round-trip stable - preserving existing headers verbatim (as before) plus the `headerOrder` fix already gives us that. Force-updating `X-Generator` / MIME headers / `Language` means every existing catalog gets rewritten once after upgrade, and `check sync` fails until users re-run extract. I'd keep the improved headers for newly created files only, and leave existing files untouched.

packages/format-po/src/po.ts
```diff
@@ -146,19 +146,138 @@ function isGeneratedId(id: string, message: MessageType): boolean {
...
+  const po = parsePo(content)
+
+  // pofile-ts 4.0.3 can lose the obsolete marker when an obsolete item
+  // immediately follows an active item without an intervening comment.
+  const obsoleteItems = content
+    .split(/\r?\n\r?\n/)
+    .filter((section) => /^#~\s+(?:msgctxt|msgid)\b/m.test(section))
+    .flatMap((section) => parsePo(section).items)
+    .filter((item) => item.obsolete)
+
```
andrii-bodnar, line 265: Could you please open an issue for this in pofile-ts and link it here? Would be good to have this workaround tracked so we can drop it once it's fixed upstream.

packages/format-po/src/po.ts
andrii-bodnar, line 1: The `format-po` changes concern me the most in this PR - they are user-visible for everyone, not just `check` users. Can we keep the upgrade diff on existing catalogs as close to zero as possible?

andrii-bodnar:
Hi @yslpn, no rush on this one. The main thing left is the `format-po` header handling: it's user-visible for everyone, not just `check` users, so the upgrade diff on existing catalogs should stay as close to zero as possible. The rest are small, plus a rebase.

Happy to take it from here if you're short on time - whatever works for you.

yslpn:
> Hi @yslpn, no rush on this one. The main thing left is the `format-po` header handling: it's user-visible for everyone, not just `check` users, so the upgrade diff on existing catalogs should stay as close to zero as possible. The rest are small, plus a rebase.
> 
> Happy to take it from here if you're short on time - whatever works for you.

Hi @andrii-bodnar Sorry for leaving this unaddressed. I wanted to get back to it sooner, but I was focused on another project.

I’ll get back to this within the coming week. I hope to bring it to a production-ready state. The MR turned out to be quite large.

packages/cli/src/api/check/sync.ts
yslpn: Agreed. I updated the CLI help and documentation to describe `--clean` directly: it removes obsolete messages from the expected catalog before comparing it with the existing catalog, while obsolete messages that remain in the existing catalog are reported as out of sync.

packages/cli/src/api/findings.ts
yslpn: Agreed. I moved `CheckFindingBase` into `api/check/types.ts`, changed the import in `catalog/translations.ts` to a type-only import, and removed `api/findings.ts`.

packages/cli/src/lingui-check.ts
yslpn: I used the exact wording you suggested in all five CLI commands (`check`, `compile`, `extract`, `extract-template`, and `extract-experimental`) and synchronized the documentation and blog post with the same text.

packages/format-po/src/po.ts
yslpn: Agreed. Existing PO files now preserve the header values and order recorded in `headerOrder`; headers injected by `pofile-ts` but absent from the source file are not serialized. Managed headers are normalized only for new files, while explicit `customHeaderAttributes` still override existing values or add configured headers. I also added regression coverage for legacy headers, custom headers, their order, and new-file normalization.

packages/format-po/src/po.ts
yslpn: I kept the workaround and linked the existing [pofile-ts issue #22](https://github.com/sebastian-software/pofile-ts/issues/22) and [PR #23](https://github.com/sebastian-software/pofile-ts/pull/23) directly in the code. I did not create a duplicate issue, and added a regression test for an active and obsolete entry with the same `msgid`.

packages/format-po/src/po.ts
yslpn: Addressed by the header-handling change above. Existing catalogs keep their actual headers and their order, so serialization is round-trip stable after the upgrade; only newly created catalogs receive the normalized managed headers. The full test suite passes, including the `check sync` legacy-header regression test.

yslpn:
@andrii-bodnar  Thanks for the follow-up. I’ve addressed the remaining review comments. Please let me know if I missed anything. Next time I won't delay, I'll answer quickly.

andrii-bodnar (approved):
@yslpn thank you for the contribution!

Issue #2436, opened 2026-02-07
Rust extractor


timofei-iatsenko:
I’m pretty close to building a Rust version of the extractor using SWC.

The [recent changes](https://github.com/lingui/swc-plugin/pull/193) finally closed the gap between the Babel macro and SWC, so they’re now at feature parity. That was the hardest part, since most of the heavy lifting actually happens inside the macro itself.

When I say “extractor”, I don’t mean a full CLI rewrite. I’m talking only about the part that analyzes source files and extracts messages — the bit that’s currently done by Babel.

The idea is to ship the Rust extractor as a separate package at first, alongside the existing Babel one. After some real-world testing, it could become the default, while keeping the Babel extractor around for Vue and other JS-based setups.

To get there, we’d need to port `@lingui/babel-plugin-extract-messages` to Rust. The plugin is pretty small (~400 LOC), and I think an LLM could handle most of that work without much trouble.

After that, we can build a small Rust program on top of SWC, bake in all the required transforms, compile it into a single binary, and expose it to Node.js using [napi-rs](https://napi.rs/). In the end, this would just be a prebuilt binary with one job: extract messages.

timofei-iatsenko:
@Zxilly @karmeleon @jessicalc @AndrewIngram
I know some of you have run into Lingui performance issues in the past, and I’d really appreciate your thoughts on this proposal. It would also be great if you could share your use cases and any concrete stats you have. 

Looking forward to any feedback — thanks!

timofei-iatsenko:
I have a quick update: I’ve successfully rewrote `@lingui/babel-plugin-extract-messages` using Rust and SWC, including full source map management. While it’s still in the early stages and needs some polishing, it has already passed the entire original test suite.

I expect this version to be significantly faster than Babel for two main reasons:

**Optimized AST Processing:**

Unlike a full SWC compilation cycle, I’m only using the `swc_parser`. We don’t need the extra overhead of removing TypeScript types, downleveling syntax, hygiene passes or printing AST back to code. Our workflow is much leaner:

1. Parse string to AST.
2. Run a resolver pass for scopes and identifiers.
3. Run the LinguiMacro and Extractor visitor passes.

**Native Execution (No WASM and serialization overhead):**

[...]

PR #2540, opened 2026-04-30
perf(cli): speed up catalog merge key partitioning

almsh:
# Description

In our project, we have around 20k messages in Lingui catalogs, and this change cuts our extraction time nearly in half.

`mergeCatalog` currently partitions catalog keys with repeated `Array.includes` checks:

```ts
const newKeys = nextKeys.filter((key) => !prevKeys.includes(key))
const mergeKeys = nextKeys.filter((key) => prevKeys.includes(key))
const obsoleteKeys = prevKeys.filter((key) => !nextKeys.includes(key))
```

For large catalogs, this becomes expensive because each `includes` call may scan the other key array. This PR uses `Set.has` when both catalogs contain keys, reducing key partitioning from `O(N * P)` repeated scans to `O(N + P)` Set construction and lookup work. It also keeps a fast path for empty catalogs, avoiding unnecessary Set creation during initial extraction or fully-obsolete cases.

### Operation Count Estimate

Let:

```txt
N = nextKeys.length
P = prevKeys.length
```

Current implementation:

```txt
newKeys:      N items * O(P) lookup in prevKeys
mergeKeys:   N items * O(P) lookup in prevKeys
obsoleteKeys: P items * O(N) lookup in nextKeys

total:       O(N * P)
```

Optimized version:

```txt
build Sets:  O(P + N)
newKeys:     N items * O(1) lookup in prevKeySet
mergeKeys:   N items * O(1) lookup in prevKeySet
obsoleteKeys: P items * O(1) lookup in nextKeySet

total:       O(N + P)
```

In our case, both the previous and next catalogs contain around 20k messages. With the current implementation, each extraction can require hundreds of millions of array comparison steps while partitioning catalog keys. With this change, the same step is reduced to roughly 100k Set build/lookups.

This keeps the behavior unchanged while making large catalog merges significantly cheaper.

One question about the checklist: since this change keeps the existing behavior covered by the current `mergeCatalog` tests and only changes the key partitioning strategy, would you like me to add a small performance check or benchmark test for this path?

## Types of changes

- [ ] Documentation update
- [ ] Examples update

## Checklist

### DIFF
--- a/packages/cli/src/api/catalog/mergeCatalog-performance.test.ts
+++ b/packages/cli/src/api/catalog/mergeCatalog-performance.test.ts
@@ -1,0 +1,79 @@
+import { performance } from "node:perf_hooks"
+import { describe, expect, test } from "vitest"
+
+import {
+  defaultMergeOptions,
+  makeNextMessage,
+  makePrevMessage,
+} from "../../tests.js"
+import { mergeCatalog } from "./mergeCatalog.js"
+import { CatalogType, ExtractedCatalogType } from "../types.js"
+
+function median(values: number[]) {
+  const sorted = [...values].sort((a, b) => a - b)
+
+  return sorted[Math.floor(sorted.length / 2)]!
+}
+
+function measure(fn: () => void, runs = 15) {
+  const times: number[] = []
+
+  // Warm-up for JIT
+  for (let i = 0; i < 20; i++) {
+    fn()
+  }
+
+  for (let i = 0; i < runs; i++) {
+    const start = performance.now()
+    fn()
+    const end = performance.now()
+
+    times.push(end - start)
+  }
+
+  return median(times)
+}
+
+function makePrevCatalog(size: number) {
+  return Object.fromEntries(
+    Array.from({ length: size }, (_, index) => [
+      `prev.${index}`,
+      makePrevMessage({ translation: `Translation ${index}` }),
+    ]),
+  ) as CatalogType
+}
+
+function makeNextCatalog(size: number) {
+  return Object.fromEntries(
+    Array.from({ length: size }, (_, index) => [
+      `next.${index}`,
+      makeNextMessage({ message: `Message ${index}` }),
+    ]),
+  ) as ExtractedCatalogType
+}
+
+describe("mergeCatalog performance", () => {
+  test("merge scales approximately linearly", () => {
+    const sizes = [1_000, 2_000, 4_000, 8_000]
+
+    const measurements = sizes.map((size) => {
+      const prevCatalog = makePrevCatalog(size)
+      const nextCatalog = makeNextCatalog(size)
+
+      return {
+        size,
+        timeMs: measure(() => {
+          mergeCatalog(prevCatalog, nextCatalog, false, defaultMergeOptions)
+        }),
+      }
+    })
+
+    const small = measurements[0]!
+    const large = measurements[measurements.length - 1]!
+    const inputGrowth = large.size / small.size
+    const timeGrowth = large.timeMs / small.timeMs
+
+    // Allow CI noise while still keeping the budget well below quadratic growth.
+    expect(timeGrowth).toBeLessThan(inputGrowth * 3)
+  })
+})
--- a/packages/cli/src/api/catalog/mergeCatalog.ts
+++ b/packages/cli/src/api/catalog/mergeCatalog.ts
@@ -11,7 +11,7 @@ export function mergeCatalog(
   const nextKeys = Object.keys(nextCatalog)
   const prevKeys = Object.keys(prevCatalog)
 
-  const hasKeysInBothCatalogs = prevKeys.length && nextKeys.length
+  const hasKeysInBothCatalogs = prevKeys.length > 0 && nextKeys.length > 0
   let newKeys: string[]
   let mergeKeys: string[]
   let obsoleteKeys: string[]
--- a/packages/cli/src/api/catalog/mergeCatalog.ts
+++ b/packages/cli/src/api/catalog/mergeCatalog.ts
@@ -11,9 +11,23 @@ export function mergeCatalog(
   const nextKeys = Object.keys(nextCatalog)
   const prevKeys = Object.keys(prevCatalog)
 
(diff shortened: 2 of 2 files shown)


timofei-iatsenko:
Thanks! That's interesting. Recently i [tested perf](https://github.com/lingui/js-lingui/issues/2436) of extractor quite a lot, i also found this suspicious code but it in my test it haven't contributed to the total extraction time at all. Most of the time was spent on the extraction (with babel) and po files parsing/serializing. 

So that quite a surprising me.

> would you like me to add a small performance check or benchmark test for this path?

How would you do this? Actually adding a small performance check would be nice, but currently we don't have any established pattern in the lingui repo. Would apreciate if introduce one. 

Thanks!

almsh:
Thanks for the context!

That makes sense. One important difference in our setup is that we use a custom PO formatter based on [`gettext-parser`](https://www.npmjs.com/package/gettext-parser) rather than the default [`pofile`](https://www.npmjs.com/package/pofile)-based one, so parsing/serializing is much faster for us. That probably made this merge step more visible.

Regarding the performance check: sorry, I don’t have a proper benchmark infrastructure proposal that would be broadly reusable across the repo. For this specific path, I see two possible lightweight approaches, both of which are hacky in their own way:

1. A [complexity regression test](https://github.com/almsh/js-lingui/blob/merge-catalog-keys-perf-tests/packages/cli/src/api/catalog/mergeCatalog-complexity.test.ts) that counts key-array access operations during `mergeCatalog` and asserts that the implementation does not do work proportional to `prevKeys.length * nextKeys.length`. This avoids timing-based assertions, but relies on a scoped `Object.keys` mock to return tracked key arrays.

2. A [small timing-based scale test](https://github.com/almsh/js-lingui/blob/merge-catalog-keys-perf-tests/packages/cli/src/api/catalog/mergeCatalog-scale.test.ts) that measures median runtime for increasing catalog sizes and checks that runtime growth stays roughly linear. This is closer to a performance check, but it is inherently noisier because it depends on runtime/CI conditions.

I added both variants to another branch in my fork for now. Which approach would you prefer for this PR, if any?

timofei-iatsenko:
I think i like the second approach. Let's add this to the test suite. Thanks!

timofei-iatsenko:
Regarding  https://www.npmjs.com/package/gettext-parser could you share how much faster it is? We can consider switching out implementation to that. 

Do you use parsing as stream or any other performance optimizations?

almsh:
> I think i like the second approach. Let's add this to the test suite. Thanks!

Thanks! I added the second test in a separate file, `mergeCatalog-performance.test.ts`, since it is timing-based and a bit different from the regular behavioral tests in `mergeCatalog.test.ts`.

Would you prefer me to move it into `mergeCatalog.test.ts`, or is it okay to keep it separate?



> Regarding https://www.npmjs.com/package/gettext-parser could you share how much faster it is? We can consider switching out implementation to that.
> 
> Do you use parsing as stream or any other performance optimizations?

I do not have an isolated `gettext-parser` vs `pofile` benchmark, so the numbers I have are for our formatter as a whole rather than for the library swap alone. We made this change more than a year ago, and I found these numbers in an old Slack thread, so they were measured with older versions of the tools and should be treated as historical rather than precise current benchmarks🙂 

On our workload at the time, our formatter made `lingui extract` about 2.5× faster and `lingui compile` about 50× faster.

In concrete pipelines, we saw:

- Build pipeline with single-language extraction/compilation: 6m33s → 5m29s
- Upload to CrowdIn pipeline with extraction for 9 languages: 4m46s → 53s
- Download from CrowdIn pipeline with compilation for 9 languages: 5m06s → 36s
- Vite build with @lingui/vite-plugin and 9 imported PO files: 5m40s → 40s

We are not using the streaming API. Our formatter calls `gettextParser.po.parse(content)` / `gettextParser.po.compile(...)` on the full content. The other important optimization is memoization of parse/compile work inside our formatter.

My assumption is that `gettext-parser` is also faster than `pofile` on its own because its PO parser is lexer/state-machine based, while `pofile` is much more regexp-heavy, but I have not measured that part separately.

One trade-off worth mentioning: unlike the `pofile`-based formatter, our formatter is not stable with respect to obsolete/commented messages. Removed entries are moved to the bottom of the file instead of staying commented in place, so diffs are less convenient if PO files are tracked in VCS.

timofei-iatsenko:
> On our workload at the time, our formatter made lingui extract about 2.5× faster and lingui compile about 50× faster.
> 
> In concrete pipelines, we saw:
> 
> Build pipeline with single-language extraction/compilation: 6m33s → 5m29s
> Upload to CrowdIn pipeline with extraction for 9 languages: 4m46s → 53s
> Download from CrowdIn pipeline with compilation for 9 languages: 5m06s → 36s
> Vite build with @lingui/vite-plugin and 9 imported PO files: 5m40s → 40s

Interesting numbers. The Compile function has a performance bottleneck in the previous versions. It's fixed now on the level of the cli. So probably your formatter should not show that big difference anymore. Regarding the perfromance comparison i found that benchmark https://github.com/sebastian-software/pofile-ts#po-file-parsing  that compares `pofile` with `gettext-parser` and it's own implementation. 

We are going to give a chance for this `pofile-ts` and use in the lingui.

almsh:
Happy to hear you are looking at `pofile-ts` - we would be glad to switch to it once it lands in lingui.

In the meantime, I ran some local benchmarks on our project (20k messages, 9 locales, lingui 6.0.1) to see how the two changes interact:

| Formatter | Optimization | Extract | Compile |
|---|---|---|---|
| gettext-parser + memoization | without this PR | 9.20s | 2.71s |
| pofile-ts | without this PR | 8.49s | 1.13s |
| gettext-parser + memoization | with this PR | 4.29s | 1.63s |
| pofile-ts | with this PR | 3.95s | 1.15s |

The PR optimization still roughly halves extract time regardless of the formatter in our setup, so the two improvements look complementary. Is there anything missing from my side in this PR?

timofei-iatsenko (approved):

andrii-bodnar (approved):

PR #2553, opened 2026-05-24
fix: throw on dynamic macro message ids

### LINKED ISSUE #2523: [v6] Inconsistent behaviour of dynamic IDs in JS/JSX macros
### Description

While working on #2514 I became aware of a potential bug/discrepancy in how Lingui processes dynamic IDs between JS Macros and JSX Macros:

1. **JS Macros (`t({ id: dynId, message: "Welcome" })`)**: 
   Preserves the dynamic AST node `dynId` in the output because Babel processes object properties directly. The output retains `"id": dynId`.
2. **JSX Macros (`<Trans id={dynId}>Hello</Trans>`)**:
   The dynamic attribute is lost. `maybeNodeValue()` inside `macroJsx.ts` evaluates non-static JSX attributes to `null`. It strips the attribute, completely drops `dynId`, and instead relies on the hash fallback identifier from the text. 

I'm not sure what the expected behaviour is but figured I'd surface this in case it's a bug.

### Example test snapshot

```tsx
import { Trans } from "@lingui/react/macro";
const dynId = "dynamic";
<Trans id={dynId}>Hello</Trans>;

↓ ↓ ↓ ↓ ↓ ↓

import { Trans as _Trans } from "@lingui/react";
const dynId = "dynamic";
<_Trans
  {
    /*i18n*/
    ...{
      id: "uzTaYi",
      message: "Hello",
    }
  }
/>;
```

### Verifications

### Macro Support

Babel with babel-macro-plugin

### Lingui Version

6.0.0-next.3

yslpn:
# Description

Throw a compile-time error when Lingui macros receive a dynamic message `id` in JS or JSX.

Note: the SWC plugin has the same gap and should be fixed separately.

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes #2523

## Checklist

### DIFF
--- a/packages/babel-plugin-lingui-macro/test/jsx-trans.test.ts
+++ b/packages/babel-plugin-lingui-macro/test/jsx-trans.test.ts
@@ -41,14 +41,6 @@ macroTester({
       `,
     },
     {
-      name: "Preserve custom ID (dynamic expression)",
-      code: `
-        import { Trans } from '@lingui/react/macro';
-        const dynId = "dynamic";
-        <Trans id={dynId}>Hello World</Trans>;
-      `,
-    },
-    {
       name: "Should preserve reserved props: `comment`, `context`, `render`, `id`",
       code: `
         import { Trans } from '@lingui/react/macro';
--- a/packages/babel-plugin-lingui-macro/test/jsx-trans.test.ts
+++ b/packages/babel-plugin-lingui-macro/test/jsx-trans.test.ts
@@ -34,10 +34,35 @@ macroTester({
       `,
     },
     {
+      name: "Preserve custom ID (TypeScript literal assertion)",
+      useTypescriptPreset: true,
+      code: `
+        import { Trans } from '@lingui/react/macro';
+        <Trans id={"msg.hello" as const}>Hello World</Trans>;
+      `,
+    },
+    {
       name: "Preserve custom ID (template expression)",
       code: `
         import { Trans } from '@lingui/react/macro';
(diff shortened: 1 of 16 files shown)


packages/babel-plugin-lingui-macro/test/jsx-trans.test.ts
```diff
@@ -40,6 +40,14 @@ macroTester({
         <Trans id={\`msg.hello\`}>Hello World</Trans>;
       `,
     },
+    {
+      name: "Preserve custom ID (dynamic expression)",
+      code: `
+        import { Trans } from '@lingui/react/macro';
+        const dynId = "dynamic";
+        <Trans id={dynId}>Hello World</Trans>;
+      `,
```
timofei-iatsenko: This is a misuse and it should throw. Or maybe i'm missing some valid use case?

timofei-iatsenko (changes requested):

timofei-iatsenko:
As wrote above, passing a dynamic id is acceptable only on the runtime Trans. This is a valid pattern to resolve "lazy" translations. However passing a dynamic id to the Trans macro - break the logic. Extractor would not be able to extract this correctly, and it most likely a misuse. if you have a dynamic id - means you defined a message somewhere else. the message and parameters of Trans macro would be discarded.

packages/babel-plugin-lingui-macro/test/jsx-trans.test.ts
yslpn: I was careless and misunderstood the intent. The best solution would be to throw an error.

PR #2554, opened 2026-05-24
chore: harden GitHub actions supply chain

yslpn:
# Description

Hardens the GitHub Actions supply chain by pinning workflow actions to full commit SHAs while keeping the resolved versions as inline comments. Also limits default `GITHUB_TOKEN` permissions in read-only workflows, enables Yarn Hardened Mode for manual release install steps, and adds Dependabot coverage for monthly GitHub Actions updates.

## Types of changes

- [ ] Documentation update
- [ ] Examples update

## Checklist

### DIFF
--- a/.github/workflows/release-docs.yml
+++ b/.github/workflows/release-docs.yml
@@ -3,6 +3,9 @@ name: release-docs
 on:
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   release:
     runs-on: ubuntu-latest
@@ -19,6 +22,8 @@ jobs:
 
       - name: Install dependencies
         working-directory: website
+        env:
+          YARN_ENABLE_HARDENED_MODE: "1"
         run: yarn install
 
       - name: Build
--- a/.github/workflows/release-docs.yml
+++ b/.github/workflows/release-docs.yml
@@ -11,12 +11,12 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 0
 
       - name: Setup node
-        uses: actions/setup-node@v6
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
           node-version: 'lts/*'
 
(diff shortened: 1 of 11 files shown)


andrii-bodnar (commented):
@yslpn thanks for the contribution!

Looks good overall. My only concern is maintainability of the SHA-pinned actions: without Dependabot coverage for GitHub Actions, these pins can become stale and updates will be easy to miss.

I haven’t looked into this deeply, but could we automate updates for the pinned action SHAs?

yslpn:
> @yslpn thanks for the contribution!
> 
> Looks good overall. My only concern is maintainability of the SHA-pinned actions: without Dependabot coverage for GitHub Actions, these pins can become stale and updates will be easy to miss.
> 
> I haven’t looked into this deeply, but could we automate updates for the pinned action SHAs?

Good point. Dependabot can handle this for SHA-pinned GitHub Actions as long as the version comment stays on the same line, which is the format used here.

I can add this config to current PR:

```yaml
version: 2
updates:
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
    commit-message:
      prefix: "chore"
```

The `chore` prefix should keep Dependabot PR titles compatible with our semantic title check.

https://docs.github.com/en/code-security/reference/supply-chain-security/supported-ecosystems-and-repositories#github-actions

andrii-bodnar:
@yslpn sounds great, let's automate it this way. I would only change the schedule interval to monthly, there's no need to do it every week as this can generate too much noise. Also, what about using the `ci` prefix?

.github/workflows/release-docs.yml
```diff
@@ -3,22 +3,27 @@ name: release-docs
...
       - name: Setup node
-        uses: actions/setup-node@v6
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
           node-version: 'lts/*'
 
       - name: Install dependencies
         working-directory: website
+        env:
+          YARN_ENABLE_HARDENED_MODE: "1"
```
timofei-iatsenko, line 26: can this be enabled in the yarn configuration file? So when install on local machine the same rules are applied?

yslpn: There is no convenient way to enable this by default, as it is not recommended.

image


https://yarnpkg.com/blog/release/4.0#hardened-mode
https://yarnpkg.com/configuration/yarnrc#enableHardenedMode

timofei-iatsenko: ok, i'm not familiar with latest changes in Yarn. We use pnpm at my job, and i supposed that this hardened mode is a similar things to what pnpm did recently - disable postinstall scripts by default and set minimumReleaseAge to 1 day.

timofei-iatsenko (approved):

andrii-bodnar (approved):

PR #2562, opened 2026-06-02
feat(cli): add support for `BatchExtractor`

timofei-iatsenko:
Summary

  - Add `Experimental__BatchExtractorType` — a new extractor shape that receives all matched file paths at once and handles file I/O and parallelism internally
  - Extractors are processed in config order: each extractor's `match()` consumes files from the pool, remaining files pass to the next extractor

  Motivation

  Currently extractors are executed per file. The CLI reads each file, then calls the extractor's extract() method one at a time (or via Node.js worker threads). For a native Rust/SWC-based extractor, this design has significant overhead:

  - Constant back-and-forth between Node.js and Rust for each file
  - File reads and native extension calls all compete for libuv's thread pool (hardcoded to 4 workers), saturating it and fighting for priority

  A native Rust extractor can handle parallelism without involving libuv or any internal Node.js scheduling — reading and parsing files across all available CPU cores with minimal overhead.

  Design

  The new type is a union member alongside the existing `PerFileExtractorType`:
```ts
  export type ExtractorType = PerFileExtractorType | Experimental__BatchExtractorType
```
  Both types have a match() method. Extractors are processed in config order — each one claims the files it matches, removing them from the pool. This allows mixing batch and per-file extractors naturally:

```ts
  // Example: native Rust handles JS/TS, Vue extractor handles .vue files
  extractors: [nativeBatchExtractor, vuePerFileExtractor]
```

  For batch extractors, matched files are passed directly to `extractFromFiles()` — no `fs.readFile`, no worker pool. For per-file extractors, the existing behavior is preserved (file reading + optional worker pool).

  The `Experimental__ prefix` allows us to merge and publish this while continuing to iterate on the API until it's stable.

## Performance

```
══════════════════════════════════════════════════════════════
  Lingui Benchmark — Preset: medium
  1000 files · 10.0k messages · 5 locales
  Node v24.13.1 · darwin arm64
  Apple M3 Max (16 cores)
══════════════════════════════════════════════════════════════

Running: Extract (full pipeline)...

  Extract (scan → extract → merge → write):
    SWC · 1 worker         █████████████████████████  1.35s  739 files/s
    SWC · 2 workers        ████████████████████░░░░░  1.10s  905 files/s ±11.7%
    SWC-batch · 1 worker   █████████████░░░░░░░░░░░░  697ms  1435 files/s ⚡
    SWC-batch · 2 workers  █████████████░░░░░░░░░░░░  712ms  1405 files/s ±17.5%

  Summary:
    SWC-batch · 1 worker is 1.9x faster than SWC · 1 worker
    SWC-batch · 1 worker is 1.6x faster than SWC · 2 workers
    SWC-batch · 1 worker is 1.0x faster than SWC-batch · 2 workers
```

When batch extractor is used it bypass and disable all parallelism on the CLI level, even if it's reqeusted explicitly by user. (because there is nothing to parallel, all files send as-is to native code)

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

### DIFF
--- a/packages/cli/src/api/catalog.ts
+++ b/packages/cli/src/api/catalog.ts
@@ -4,7 +4,6 @@ import { globSync } from "node:fs"
 import normalize from "normalize-path"
 
 import {
-  Experimental__BatchExtractorType,
   ExtractedMessage,
   ExtractorType,
   LinguiConfigNormalized,
@@ -34,13 +33,10 @@ import {
 } from "./utils.js"
 import { AllCatalogsType, CatalogType, ExtractedCatalogType } from "./types.js"
 import { ExtractWorkerPool } from "./workerPools.js"
-import { getConfiguredExtractors } from "./extractors/index.js"
-
-function isBatchExtractor(
-  ext: ExtractorType,
-): ext is Experimental__BatchExtractorType {
-  return "extractFromFiles" in ext
-}
+import {
+  getConfiguredExtractors,
+  isBatchExtractor,
+} from "./extractors/index.js"
 
 const LOCALE = "{locale}"
 const LOCALE_SUFFIX_RE = /\{locale\}.*$/
--- a/packages/cli/src/api/catalog.ts
+++ b/packages/cli/src/api/catalog.ts
@@ -174,7 +174,7 @@ export class Catalog {
 
     const extractors = getConfiguredExtractors(this.config)
 
-    // optimized hot path, is there is not btach extractors defined, skip more complex logic
+    // Optimized hot path: if there are no batch extractors defined, skip the more complex logic.
     if (!extractors.some(isBatchExtractor)) {
       const success = options.workerPool
         ? await extractFromFilesWithWorkerPool(
@@ -210,10 +210,20 @@ export class Catalog {
     for (const extractor of extractors) {
       if (remaining.length === 0) break
 
-      const matched = remaining.filter((f) => extractor.match(f))
+      const matched: string[] = []
+      const unmatched: string[] = []
+
+      for (const f of remaining) {
+        if (extractor.match(f)) {
+          matched.push(f)
+        } else {
+          unmatched.push(f)
+        }
+      }
+
       if (matched.length === 0) continue
 
-      remaining = remaining.filter((f) => !extractor.match(f))
+      remaining = unmatched
 
       if (isBatchExtractor(extractor)) {
         try {
--- a/packages/cli/src/api/catalog.ts
+++ b/packages/cli/src/api/catalog.ts
@@ -167,29 +167,43 @@ export class Catalog {
       paths = paths.filter((path: string) => regex.test(normalize(path)))
     }
 
-    const extractors = getConfiguredExtractors(this.config)
-
-    if (!extractors.some(isBatchExtractor)) {
-      if (options.workerPool) {
-        return await extractFromFilesWithWorkerPool(
-          options.workerPool,
-          paths,
-          this.config,
-        )
-      }
-
-      return await extractFromFiles(paths, this.config)
+    const messages: ExtractedCatalogType = {}
+    const onMessageExtracted = (next: ExtractedMessage) => {
+      mergeExtractedMessage(next, messages, this.config)
     }
 
-    return await this.collectWithExtractors(extractors, paths, options)
+    const extractors = getConfiguredExtractors(this.config)
+
+    // optimized hot path, is there is not btach extractors defined, skip more complex logic
+    if (!extractors.some(isBatchExtractor)) {
+      const success = options.workerPool
+        ? await extractFromFilesWithWorkerPool(
+            options.workerPool,
+            paths,
+            onMessageExtracted,
+            this.config,
+          )
+        : await extractFromFiles(paths, onMessageExtracted, this.config)
+
+      return success ? messages : undefined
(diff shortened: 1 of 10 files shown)


packages/conf/src/types.ts
```diff
@@ -49,6 +59,39 @@ export type ExtractorType = {
...
+  extractFromFiles(
+    filenames: string[],
+    onMessageExtracted: (msg: ExtractedMessage) => void,
+    ctx: ExtractorCtx,
+  ): Promise<void>
+}
+
+export type ExtractorType =
+  | PerFileExtractorType
+  | Experimental__BatchExtractorType
```
andrii-bodnar, line 93: `ExtractorType` is now a union - breaking type change. `.extract` now lives only on the `PerFileExtractorType` arm, so any v6 consumer who typed a custom extractor as `ExtractorType` and touched `.extract`/`.match` gets a compile error

packages/conf/src/types.ts
```diff
@@ -49,6 +59,39 @@ export type ExtractorType = {
...
+/**
+ * Batch extractor that receives all matched file paths at once and handles
+ * file I/O and parallelism internally.
+ *
+ * Files are matched using `match` and only those are passed to `extractFromFiles`.
+ * The CLI does not read file contents or use worker pools for batch extractors.
+ *
+ * @experimental This type is experimental and may change in future versions.
+ */
+export type Experimental__BatchExtractorType = {
```
andrii-bodnar, line 71: We usually don't use double-underscore for this. Maybe use `Experimental_BatchExtractorType`?

packages/cli/src/api/catalog.ts
```diff
@@ -156,15 +163,92 @@ export class Catalog {
...
+      if (isBatchExtractor(extractor)) {
+        try {
+          await extractor.extractFromFiles(matched, onMessageExtracted, {
+            linguiConfig: this.config,
+          })
+        } catch (e) {
+          console.error(`Batch extractor failed: ${(e as Error).message}`)
+          console.error((e as Error).stack)
+          catalogSuccess = false
+        }
```
andrii-bodnar, line 233: One thing about this try/catch. Since `onMessageExtracted` runs inside `extractFromFiles`, it also catches the "different default translations" error that `mergeExtractedMessage` throws on purpose for duplicate-ID conflicts. So user error gets reported as "Batch extractor failed: ..." and quietly turned into `catalogSuccess = false`, instead of crashing the way it should

andrii-bodnar (commented):
Direction looks good on its own. A few things before merge:

- Even as experimental, do we need some docs about this? 
- What would be the signal of dropping the experimental flag?

What are the next steps for this initiative?

timofei-iatsenko:
> * Even as experimental, do we need some docs about this?

I prefer to document it once the experimental flag would be dropped. Currently here is a comprehensive jsdoc inlined. 

> What would be the signal of dropping the experimental flag?

Since the extractor which suppose to use this new api in the separate repository and i could not check everethying at once, this experimental flag could be dropped once i test both parties and validate this works and worth it. 

> What are the next steps for this initiative?

We merge swc-extractor fork to the original repo, publish it under `@lingui/` organization, update docs, and make a publish anouncment. The general plan is to have this extractor opt-in in the current version, and opt-out in the next major version.

packages/conf/src/types.ts
timofei-iatsenko: This is not particularly true, we have a type test testing assignability of the custom extractor to a type

https://github.com/lingui/js-lingui/blob/85278f4c0ff93b73bfbda9bed9889b7757594ac8/packages/conf/__typetests__/index.tst.ts#L47-L61

And i also test it with a direct assign:

Image

So we are good here.

packages/conf/src/types.ts
timofei-iatsenko: I don't see any difference between two or one, i propose to live as is, just to not make any code changes. This is going to be a short-living flag anyway.

packages/cli/src/api/catalog.ts
timofei-iatsenko: I followed both code pathes w/wo batch extractor and found that behaviour is the same. If the duplication error is thrown from the code it gets cought, then console.error'ed and full extraction process marked as failed. 

The idea is that it doesn't stop on first discovered issue but instead it go over all files and shows all issues. 

I added a specific test case covering this particular behavior for the batch extractor. 

For reference this is implementation in non batch extractor: 

https://github.com/lingui/js-lingui/blob/45fc3e6d8920b5c9656db11ad336643c9ba6ed1d/packages/cli/src/api/extractors/index.ts#L50-L61

and test covering this:

https://github.com/lingui/js-lingui/blob/72bbaa87a3a24149262f5ac00ba79aaf411ed12c/packages/cli/src/api/catalog.test.ts#L447-L470

Note that despite test says "should throw an error" actually it just call a `console.error` and continue

andrii-bodnar (approved):

timofei-iatsenko:
@andrii-bodnar this expermental functionality could be reverted back, because after all optimisations done to the Rust extractor turned out that file parallelism is no longer a bottleneck.

andrii-bodnar:
It's not possible to automatically revert here anymore, so I'll do it manually

Issue #2526, opened 2026-04-22
Migrate examples to Lingui v6


andrii-bodnar:
https://github.com/lingui/js-lingui/tree/main/examples

It would be better to create a separate pull request for each example project.

PR #2566, opened 2026-06-05
chore: update vanilla js example to lingui v6

suzunn:
## Problem

The vanilla JavaScript example still referenced Lingui v5 prerelease packages, while issue #2526 tracks moving examples to Lingui v6.

## Solution

Updated the vanilla JavaScript example's Lingui runtime, CLI, and Babel macro plugin dependencies to the v6 line used by the already-migrated examples.

## Testing

Inspected the example package.json and compared the dependency pattern with the v6 Next.js/Babel example through the GitHub web UI. I could not run package installation or tests because this run is limited to Chrome-only browser edits.

### DIFF
--- a/examples/js/package.json
+++ b/examples/js/package.json
@@ -12,13 +12,13 @@
     "extract": "lingui extract --clean"
   },
   "dependencies": {
-    "@lingui/core": "^5.0.0-next.3"
+    "@lingui/core": "^6.0.0"
   },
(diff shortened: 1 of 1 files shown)


andrii-bodnar:
@suzunn thanks for the contribution! However, bumping version numbers alone may not be enough - Lingui v6 has several breaking changes beyond a version bump.

PR #2579, opened 2026-06-18
fix: respect corePackage/jsxPackage setting in experimental extractor

mogelbrod:
# Description

This PR updates the experimental extractor (`lingui extract-experimental`) to respect the `macro.corePackage` / `macro.jsxPackage` settings.

Up until now it only considers files that import directly from `@lingui/*/macro` for extraction using a hardcoded regex (`/from ["']@lingui(\/.+)?\/macro["']/`).

## Fix

The pre-filter is now built from the configured macro packages (`macro.corePackage` + `macro.jsxPackage`) in addition to the built-in `@lingui/*/macro` entrypoints (kept for backwards compatibility, so the default config and the legacy `@lingui/macro` package keep working even when a project overrides `corePackage`/`jsxPackage`). The regex is also built once per plugin instance instead of per file, which removes a latent `RegExp`-with-`/g` + `.test()` `lastIndex` statefulness bug.

## Tests

Added an integration test `should extract messages imported from a configured custom macro package` with a new fixture (`packages/cli/test/extractor-experimental-custom-macro`) that imports `t`, `msg` and `<Trans>` through a `#macros` subpath barrel declared in `macro.corePackage` / `macro.jsxPackage`, and asserts all three messages are extracted. The test fails on `main` (no messages extracted) and passes with this change. Existing `extractor-experimental` tests continue to pass, covering the backwards-compatible default behavior.

## Types of changes

- [ ] Documentation update
- [ ] Examples update

## Checklist

### DIFF
--- a/packages/cli/src/extract-experimental/linguiEsbuildPlugin.ts
+++ b/packages/cli/src/extract-experimental/linguiEsbuildPlugin.ts
@@ -10,44 +10,57 @@ import { LinguiConfigNormalized } from "@lingui/conf"
 
 export const pluginLinguiMacro = (options: {
   linguiConfig: LinguiConfigNormalized
-}): Plugin => ({
-  name: "linguiMacro",
-  setup(build) {
-    build.onLoad({ filter: babelRe, namespace: "" }, async (args) => {
-      const filename = path.relative(process.cwd(), args.path)
+}): Plugin => {
+  const configuredPackages = [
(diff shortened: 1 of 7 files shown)


timofei-iatsenko:
This is already covered in this complete rework of experimental extractor: https://github.com/lingui/js-lingui/pull/2572

PR #2583, opened 2026-06-26
feat: add pseudolocalize options

### LINKED ISSUE #2582: Pseudolocale options (Simulating DE (+~30%), FI, etc.)
### Problem Description

Pseudolocalization is useful for testing, however the Lingui doesn't add length expansion for UI testing.
Ideally this would be an option that can be passed to the Pseudolocalization settings.

### Proposed Solution

The posibility to add options to the pseudolocalization in Lingui

### Alternatives Considered

Writing my own patch or use german to check UI

### Additional Context

_No response_

ThibaultJanBeyer:
# Description

The current pseudolocalization doesn't allow for padding to catch UI bugs caused by longer languages.
Here is the related issue: https://github.com/lingui/js-lingui/issues/2582

## Types of changes

The signature of the `pseudoLocale` can now look either like this:

```ts
// before
export default defineConfig({
  locales: ['en', 'pseudo'],
  sourceLocale: 'en',
  pseudoLocale: 'pseudo',
  catalogs: [],
})
```

or like this:

```ts
// after
export default defineConfig({
  locales: ['en', 'pseudo'],
  sourceLocale: 'en',
  pseudoLocale: { locale: 'pseudo', prepend: '⟦ ', append: ' ⟧', extend: 0.4 },
  catalogs: [],
})
```

- [ ] Documentation update
- [ ] Examples update

Fixes #2582

## Checklist

(I'm not sure, added it to website/docs/ref/conf.md. Somewhere else to add?


packages/conf/src/index.ts
```diff
@@ -1,4 +1,5 @@
 export { makeConfig } from "./makeConfig"
 export { defineConfig } from "./defineConfig"
 export { getConfig } from "./getConfig"
+export { getPseudoLocale, getPseudoLocaleOptions } from "./getPseudoLocale"
```
timofei-iatsenko: i suggest not export that from the package, but instead apply and normalize config in the `makeConfig`. So all consumers would receive normalized (expanded) option

timofei-iatsenko (changes requested):

packages/conf/src/index.ts
ThibaultJanBeyer: Ok I normalized it in makeConfig, please have a look

website/docs/ref/conf.md
```diff
@@ -215,6 +215,36 @@ Default value: `""`
...
+
+| Option     | Type     | Default     | Description                                                                                          |
+| ---------- | -------- | ----------- | ---------------------------------------------------------------------------------------------------- |
+| `locale`   | `string` | —           | Locale used for pseudolocalization (required in the object form).                                    |
+| `prepend`  | `string` | `""`        | String prepended to the beginning of every pseudo-localized message.                                 |
+| `append`   | `string` | `""`        | String appended to the end of every pseudo-localized message.                                        |
+| `extend`   | `number` | `0`         | Extends the width of the string by the given percentage (e.g. `0.3` adds 30%).                       |
+| `override` | `string` | `undefined` | Replaces every (non-token) character with the given one. Handy to quickly spot untranslated strings. |
```
andrii-bodnar:
```suggestion
| Option     | Type     | Default     | Description                                                                                         |
| ---------- | -------- | ----------- | --------------------------------------------------------------------------------------------------- |
| `locale`   | `string` | —           | Locale used for pseudolocalization (required in the object form)                                    |
| `prepend`  | `string` | `""`        | String prepended to the beginning of every pseudo-localized message                                 |
| `append`   | `string` | `""`        | String appended to the end of every pseudo-localized message                                        |
| `extend`   | `number` | `0`         | Extends the width of the string by the given percentage (e.g. `0.3` adds 30%)                       |
| `override` | `string` | `undefined` | Replaces every (non-token) character with the given one. Handy to quickly spot untranslated strings |
```

andrii-bodnar (changes requested):

andrii-bodnar:
@ThibaultJanBeyer thanks for the contribution!

I was thinking of marking the old configuration (`pseudoLocale: string`) as deprecated and switching to the new, object-based configuration in the next major release.

What do you guys think?

ThibaultJanBeyer:
> I was thinking of marking the old configuration (pseudoLocale: string) as deprecated and switching to the new, object-based configuration in the next major release.

Up to you, I was keeping both to not introduce any breaking changes. @andrii-bodnar, let me know what you prefer

andrii-bodnar:
@ThibaultJanBeyer yes, that's the right way. I meant to keep the old way but mark it as deprecated, and switch completely to the new way in a future major release.

BTW, could you please take a look at the failed docs pipeline? There are some formatting issues.

timofei-iatsenko (approved):

ThibaultJanBeyer:
> @ThibaultJanBeyer yes, that's the right way. I meant to keep the old way but mark it as deprecated, and switch completely to the new way in a future major release.
>
> BTW, could you please take a look at the failed docs pipeline? There are some formatting issues.

@andrii-bodnar  I think I marked it as deprecated, please check if it's the correct way. Should also have fixed the pipeline issues

andrii-bodnar (approved):
@ThibaultJanBeyer thank you!

PR #2610, opened 2026-07-08
fix(react): fallback to Object.create when Proxy is unavailable

### LINKED ISSUE #2609: I18nProvider crashes on mount on engines without Proxy
### Describe the bug

Since v6, `I18nProvider` creates its context value with `new Proxy(i18n, {})` inside its `useState` initializer:

https://github.com/lingui/js-lingui/blob/main/packages/react/src/I18nProvider.tsx#L63

On JavaScript engines that don't implement `Proxy`, the provider throws `ReferenceError: Proxy is not defined` on first render and the app never mounts.

Unlike most missing built-ins, `Proxy` **cannot be polyfilled** — so there is no userland workaround short of patching the package.

### Affected environments

- Embedded WebViews (e.g. Trident-based `WebBrowser` control shells still used by installers and kiosks)
- Older smart-TV browsers (Tizen 2.x, webOS 1–3)
- Any pre-Chromium-49 runtime

These engines run React (and `@lingui/react` v5) fine when the app is transpiled/polyfilled for them — `Proxy` is the one thing that can't be provided.

### To Reproduce

Render `<I18nProvider i18n={i18n}>` in any environment without `Proxy`:

```ts
delete globalThis.Proxy // simulates the engines above
render(<I18nProvider i18n={i18n}>...</I18nProvider>)
// ReferenceError: Proxy is not defined
```

### Proposed fix

Per the code comment, the Proxy is an empty-handler wrapper used purely to give the context a fresh `i18n` reference on each update so that memoized values keyed on `i18n` invalidate. A guarded fallback preserves that contract at zero cost to engines with `Proxy`:

```ts
i18n: typeof Proxy === "function" ? new Proxy(i18n, {}) : Object.create(i18n),
```

Reads through `Object.create` delegate to the instance and the reference is fresh on every update, so the memo-invalidation behavior is preserved.

I have a PR ready with this change plus a regression test that simulates a Proxy-less engine.

nmoinvaz:
# Description

`I18nProvider` wraps `i18n` in an empty-handler `Proxy` solely to create a fresh reference on each context update, so that memoized values keyed on `i18n` invalidate correctly. On engines without `Proxy` (embedded WebViews, older smart-TV browsers) this crashes the provider on mount with `ReferenceError: Proxy is not defined` — and `Proxy` cannot be polyfilled, so affected apps have no userland workaround.

This PR uses a fallback to `Object.create(i18n)` when `Proxy` is missing: the reference is still fresh on every update and property reads still delegate to the i18n instance, so the memo-invalidation contract is preserved. Engines with `Proxy` support are byte-for-byte unaffected.

Includes a regression test that simulates a Proxy-less engine (deletes `globalThis.Proxy` around render/update) and verifies memoized values keyed on `i18n` still update on locale change. The test fails with `ReferenceError: Proxy is not defined` against the previous implementation.

Fixes #2609

## Types of changes

- [ ] Documentation update
- [ ] Examples update

## Checklist

### DIFF
--- a/packages/react/src/I18nProvider.test.tsx
+++ b/packages/react/src/I18nProvider.test.tsx
@@ -262,4 +262,54 @@ describe("I18nProvider", () => {
 
     expect(getByTestId("locale").textContent).toBe("cs")
   })
+
+  it("works on engines without Proxy support (fallback to Object.create)", () => {
+    // Simulate an engine without Proxy (embedded WebViews, older smart-TV
+    // browsers). Proxy cannot be polyfilled, so the provider must not
+    // depend on it unconditionally. Assertions observe values through a
+    // probe component rather than DOM queries, because the test DOM
+    // implementation itself needs Proxy for querySelectorAll.
+    const OriginalProxy = globalThis.Proxy
+    const observedLocales: string[] = []
+
+    delete (globalThis as any).Proxy
+    try {
+      const i18n = setupI18n({
+        locale: "en",
+        messages: {
+          en: {},
+          cs: {},
+        },
+      })
+
+      const ComponentWithMemoizedI18n = () => {
+        const { i18n } = useLingui()
+
+        const getLocale = useCallback(
+          (i18nInstance: I18n) => i18nInstance.locale,
+          [],
+        )
+        const currentLocale = useMemo(() => getLocale(i18n), [getLocale, i18n])
+
+        observedLocales.push(currentLocale)
+        return null
+      }
+
+      render(
+        <I18nProvider i18n={i18n}>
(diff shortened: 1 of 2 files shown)


nmoinvaz:
Or we can just use `Object.create` only.

timofei-iatsenko:
let's use `Object.create`, @yslpn any hidden caveats we haven't considered in this solution?

yslpn:
> let's use `Object.create`, @yslpn any hidden caveats we haven't considered in this solution? 

I will consider all possible fixing options later today.

nmoinvaz:
Ok let me know if you care for the unit test or if you want me to remove it too.

PR #2618, opened 2026-07-14
feat: runtime core macro

timofei-iatsenko:
### Motivation

Currently, Lingui requires a compilation step (babel/swc plugin) to transform macros like `t`, `plural`, `select`, and `selectOrdinal` into runtime code. This creates a barrier for use cases where a build pipeline is either unavailable or undesirable — most notably on the backend (Node.js scripts, serverless functions, CLI tools) where bundle size is not a concern but DX simplicity is.

This PR introduces a **runtime macro** — a set of functions with the same signature/contract as the compile-time macro, but executed entirely at runtime without any compiler plugin. This allows using Lingui's familiar API **without any compilation step**.

### Trade-offs

- **Bundle size penalty**: The compile-time macro strips the original `message` field in production, leaving only the `id` — this significantly reduces the i18n overhead in the bundle. The runtime macro cannot do this.
- **Labeled placeholders only**: Since variable names cannot be inferred at runtime, passing raw values like `${variable}` is not supported and will throw with a helpful error message. Users must always use the labeled placeholder syntax: `${{name: value}}`. This is intentionally strict — it ensures the runtime macro produces the same ICU messages regardless of variable naming, and keeps code portable between runtime and compile-time macros. We should add a configuration flag to the compile-time macro to enforce the same restriction, so that code written for one works identically with the other.
- **Option values cannot be template literals**: Template literals inside `plural`/`select` options (e.g., `` `She is ${gender}` ``) are already evaluated by the time they reach the function. However, the full functionality is preserved — instead of a plain template literal, users wrap it with `t` to produce a macro marker: `` t`She is ${{gender}}` `` instead of `` `She is ${gender}` ``. Nested macro markers (`plural`, `select`, `selectOrdinal`) work as option values as well.

### Compatibility with existing infrastructure

The runtime macro is a **subset** of the compile-time macro API. It produces the same `MessageDescriptor` shape (`{id, message, values}`) with the same ID generation algorithm. This means:

- The existing message extractor should work out of the box with runtime macro usage
- Catalogs produced from runtime macro code are identical in format
- The unsupported syntax (like labeled placeholders inferred from variable names) was already being phased out in prior iterations

### What's left

This is PoC, not fully working feature. I wanted to check what features could be implemented without a compile time macro and what's not. The current implementation is doing just macro expansion but not actually connected to i18n.

This is still should be in done. Also JSX macro is not supported (i have another branch with PoC for that as well)

Key questions: 

- How this should be published? Included into `@lingui/core` or distributed as a standalone package? 
- How it should be connected to i18n? Do everethying as is for compile time macro? (`t` bound to global i18n, or learn from experience and re-design this, to something like `` i18n.t`Message` ``

### How to try this in your app

Copy the file to your repository. Say you have setup an alias for this file `@mycomp/native-core-macro`, then add configuration to lingui config: 

```js

export default defineConfig({
  /* ... */
  macro: {
    corePackage: ['@mycomp/native-core-macro'],
  },
})
```

Then use in your code: 

```ts
// assume you have an i18n instance

import { msg } from '@mycomp/native-core-macro'
i18n.t(msg`Hello world!`)
```

### Usage example

```ts
import { msg, plural, select } from "@lingui/core/runtime-macro"

// Simple
msg`Hello World`
// → { id: "...", message: "Hello World" }

// Positional placeholders
msg`Hello ${name}`
// → { id: "...", message: "Hello {0}", values: { "0": name } }

// Named placeholders
msg`Hello ${{name}}`
// → { id: "...", message: "Hello {name}", values: { name: "Alice" } }

// Nested plural
msg`You have ${plural({count}, { one: "# message", other: "# messages" })}`
// → { id: "...", message: "You have {count, plural, one {# message} other {# messages}}", values: { count: 5 } }

// Call expression form
msg({ id: "welcome", message: "Welcome back", context: "homepage" })
```

## For myself

- Maybe invent a new, more ergonomical way to set context and other properties?


andrii-bodnar (commented):
Thanks for the PoC, the direction looks right to me. My main concern for taking this out of draft is a parity test suite against the compile-time macro: identical inputs -> identical id/message, to avoid silent drift between compile and runtime macro. Is it possible to tape it with unit tests?

I'd ship it as a subpath export `@lingui/core/runtime-macro` (is this a valid name? I recall something from our previous releases about the requirement for macro to end with the `/macro` suffix. So, not sure about this). BTW, is it really a macro now? A macro is by definition compile-time expansion. It could be a good idea for another terminology and eventually decoupling these things.

MurzNN:
@timofei-iatsenko thank you for the feature, let's move it forward! 

Some of my projects require a native declaration of the translation messages in pure JS without any macro magic, so I started to invent workarounds like this https://github.com/lingui/js-lingui/pull/2586 

But the current implementation works much better! As I can now declare Lingui messages in any script without depending on a JS transpiler, this makes possible to use defined strings not only with Lingui, but with any other parts of the application, where using Lingui is not possible!

For example, I use Payload CMS, which uses it's own [self-implemented translation system](https://payloadcms.com/docs/configuration/i18n) and no ability to integrate other translation systems. I submitted a feature request to fix this https://github.com/payloadcms/payload/discussions/17332 but afraid that it will take too long time.

So, for now, using the `runtime-macro` and help from @timofei-iatsenko (thank you a lot for help!) I am able to define translation strings only once and reuse them in both parts - Lingui-powered frontend, and Payload CMS backend!

Here is an example of defining a list of translatable labels for a request status:
```ts
import { msg } from "@lingui/runtime-macro";

export const RequestStatuses = {
  collecting: msg`Collecting`,
  choosing: msg`Choosing`,
  fulfilled: msg`Fulfilled`,
  canceled: msg`Canceled`,
}

const msgObjectToFallback = (obj: Record<string, MessageDescriptor>) => {
  return Object.fromEntries(Object.entries(obj).map(([key, value]) => [key, value.message!]))
}

export const RequestStatusesFallback = msgObjectToFallback(RequestStatuses)

export type RequestStatus = keyof typeof RequestStatuses
```

Using this way, I can pass the RequestStatusesFallback object as a list of labels for the "requestStatus" field and it works well, without any errors and dependencies on transpilers!

MurzNN:
And about the name of this feature `runtime-macro`: I believe  the `macro` term is wrong, and `runtime` sounds too complex and unclear. 

Actually, Timofei invented a purge JS function that declares a message for translation, that works on vanilla JS without any transpilers, that unlocks usage of Lingui with any pure JS project!

So, let's maybe name it just as `@lingui/js`?

MurzNN:
Also, this feature makes possible to use Lingui in Playwright tests! This is especially convenient with the pseudo locale, reusing in the tests the same source strings and expression as in the app code!

timofei-iatsenko:
> parity test suite against the compile-time macro: identical inputs -> identical id/message

i can add additional check validating id's to existing cases in runtime version. Since the runtimve version is a **subset** from a compile time using full test suite of compile time version is not feasible. 

> I'd ship it as a subpath export @lingui/core/runtime-macro (is this a valid name? I recall something from our previous releases about the requirement for macro to end with the /macro suffix. So, not sure about this). BTW, is it really a macro now? A macro is by definition compile-time expansion. It could be a good idea for another terminology and eventually decoupling these things.

I could not agree more that naming should be adjusted, actually naming is the hardest part. 

Let's break it all: 

1. Macro in the name

There are 2 things which are called macro used in the Lingui project.  

- Historically, you should write an ICU messages directly, like `You have {count, plural, =1 {1 message} other {# messaged}}`. Then macro **helpers** to write ICU messaged were added. So macro in this context means shortcuts to write an ICU messages, disregarding the implementation (compile time, runtime). (utility method, expanded to icu message)
- However, this macro was implemented using `babel-plugin-macros`, hence the macro name, so in that context it means macro as compile time thingy processed by this plugin (piece of code expanded to another piece of code). 

Over the years and iterations all this blurred a little bit. The main focus in the guides and APIs was shifted from writing ICU messages directly - to using helpers only, from using runtime methods - to using a compile time plugin.  Basically, ICU messages became an "implementation detail", the developer quite rarely interacting with them directly, instead uses  convenient API methods. 

So looking at the context - this runtime macro - is still a macro, because it expands to an icu message + variables. However i could agree that we can shift away from this terminology and invent a new one. 

We no longer enforced by any convention from the `babel-plugin-macros`, because it's deprecated and i'm not going to consider supporting it in a new features, so naming could be whatever suits best. 

Basically, this "runtime" version could just become a new default api. 

2. Runtime in the name

After adding this we would have a 3 api surfaces: 

1. low level runtime - message lookup and placeholders evaluation
2. compile time macro -  transpiling to (1)
3. runtime macro  - currently not using anything from (1)

And here is a question, should we keep message defintion api surface at all from (1)? eq: 

```ts
// this basically uses 2 functions: message definition + message lookup
i18n._("Hello world!")
```

Here functions are clearly divided, definition and lookup:
```ts
import { msg } from "@lingui/core/runtime-macro"

// defintion
const def = msg`Hello World`
// → { id: "...", message: "Hello World" }

// lookup
i18n._(def)
```

Because both (1) and (2) technically a runtime-something, it could bring a lot of confusion.  Maybe we should deprecate and drop all the ambigous methods signatures from i18n in favour of (3) ?

We should also consider that JSX macro would also follow the same pattern. So naming pattern should be scalable for jsx as well. 

---

Regarding moving forward - I would like to hide it behind a feature flag, so we can iterate on this without introducing a new major version every time. Feature flag would be on the lingui config, and basically would just toggle extractor support for this methods.

andrii-bodnar:
Thanks @timofei-iatsenko for the write-up and @MurzNN for the real-world use cases. Here's my thinking on a direction to get this out of PoC:

- **Scope**: additive sibling to the compile-time macro, behind a lingui-config feature flag, marked experimental. No changes to existing `i18n` APIs. The bigger idea - making this the canonical define/lookup API and deprecating `i18n._("literal")` / global `t` - is worth doing, but as a separate project.

- **Name**: `@lingui/core/runtime-macro`, kept next to `/macro` on purpose. The distinction users need is "import from `/macro` -> needs the plugin; from `/runtime-macro` -> just works", and the parallel names teach that. It's not a contradiction under the "ICU-authoring shorthand" sense of macro, and renaming only the runtime side would just increase asymmetry. It's flagged/experimental, so renamable later. (@MurzNN - holding off on `@lingui/js`: a standalone package is heavy here, over-commits to a product identity, and reads oddly inside js-lingui.)

- **i18n**: define-only. Export `msg`/`plural`/`select`/`selectOrdinal`, look up via `` i18n._(msg`…`) ``. No global-bound `t`.

- **JSX out of scope**: JSX requires transpilation, so JSX authors already have a build pipeline and can use the compile-time macro - the "no build step" motivation doesn't apply. Every use case here is non-JSX. Revisit if someone has JSX *without* a transpiler.

- **De-draft bar**: a bounded parity test. `id = generateMessageId(message, context)` is pure, so matching `message`+`context` -> matching `id`. So we only assert the runtime helpers emit the same `message` as the compile-time macro over the supported subset (reusing the existing babel-transform test setup) - not the full suite.

- **Deferred**: context/property sugar - `msg({ id, message, context })` already covers it.

Next steps: (1) move to the `/runtime-macro` subpath, (2) bounded parity test, (3) wire the flag to extractor support. What do you guys think?

timofei-iatsenko:
> JSX out of scope: JSX requires transpilation, so JSX authors already have a build pipeline and can use the compile-time macro - the "no build step" motivation doesn't apply. Every use case here is non-JSX. Revisit if someone has JSX without a transpiler.

JSX is out of scope by other reason - for incremental delivery. JSX without compilation macro still has a lot of sense:

1. JSX supported natively by almost all current transplers/bundlers at native speed, eq roldown, esbuild, rsbuild, swc, etc. Adding a plugin to them to just transpile lingui macro may slowdown transpilation a lot. Not all of them allowing executing native plugins, so adding plugins basically de-optimizing them significantly and also complicates the configuration.
2. Usage with react-compiler.  Looking at how things developed over time, it seems react compiler would be deeply integrated into tooling (mostly vite + rolldown) and it seems i't will not have a way to override an order (lingui first, than react compiler) so only one way to use lingui then - is a runtime jsx. 

> Name: @lingui/core/runtime-macro, kept next to /macro on purpose. The distinction users need is "import from /macro -> needs the plugin; from /runtime-macro -> just works", and the parallel names teach that. It's not a contradiction under the "ICU-authoring shorthand" sense of macro, and renaming only the runtime side would just increase asymmetry. It's flagged/experimental, so renamable later. (@MurzNN - holding off on @lingui/js: a standalone package is heavy here, over-commits to a product identity, and reads oddly inside js-lingui.)

BTW, adding a compile step will also optimize this runtime macro calls as well. So maybe some day we will just narrow api to that subset of runtime macro and offer compiler step just as optional improvement. (similarly to what `formatjs`  offers https://formatjs.github.io/docs/tooling/swc-plugin)

MurzNN:
Thank you folks for the details, yes - agree with most of them! But about "macro" term - it is still not clear to me, why it is relevant to the pure JS function `msg()` aka `defineMessage()`?

The "macro" term means a short expression, that will be exposed to a full expression on the compile time, here is an explanation of this term:

> In programming, a macro is a rule or pattern that tells the computer to map a short name or symbol to a larger sequence of code or instructions. Before the program runs or compiles, a tool replaces the macro name with its actual defined output, a process called macro expansion.

So, the "macro" term is actually relevant to only `t` and `msg` expressions that will be replaced to the actual code on the transpiling stage.

I am not against this term, I am good to go with the "runtime-macro", just worry about clear understanding for new users of the feature that this package provide. Seeing the "macro" word in the name I immediately associate the function with transpiling, babel, vite, or some other magic, and don't even expect the pure vanilla JS functionality there.

If we take a fresh look to the project, with introducing this feature we actually provide a "missing" core part of the Lingui:
before: 
- nothing native -> transpiling macro `t` -> React component `<Trans>`
and now:
- native JS `msg` -> transpiling macro `t` -> React component `<Trans>`

So, I believe the best place for this function `msg()` is in the "@lingui/core" package, actually.

And, by the way, can we somehow reuse the same native JS function `msg()` in the "upper" components like React component and transpiling macro, instead of keeping almost the same logic in several places?

So, from the fresh look, the function `msg()` is the main core concept of the whole Lingui project: it defines a translatable string as an unique identifier. This concrete main idea stands out the Lingui project from other approaches.

And all other functions are just additional features on top of this main concept.

And now, when we have the pure JS implementation of the `msg()`, we start to provide really a simple way of integrating this core translation approach to any JS project without any "macro" magic! And not only to JS, I want to implement the same Lingui approach for Drupal CMS on PHP too, because I have mixed projects and want to make them use the single `.po` translations file for the backend and the frontend.

I don't want to slow-down or sabotage the naming process, just sharing my thoughts as a kind-of newcomer to the Lingui world :) So, feel free to name it as you see it right.

MurzNN:
By the way, about tests - I see that we have a new separate test for the JS implementation of the `msg()` and friends. But actually they should behave the same as the functions for the macro and provide the feature parity, right?

So, maybe instead of writing a separate test, we can reuse the existing test to make it test both implementation with the same test data?

PR #2242, opened 2025-04-30
fix: use makePathRegexSafe with globSync

taku-hatano:
close #2241

# Description

[//]: # (Describe the big picture of your changes here to communicate to the maintainers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.)

When using catalog's input with file-base routing systems, `[]` is not escaped.
So `globSync` can't get path under the dynamic route, like `app/[lang]/page.tsx`

## Types of changes

[//]: # (What types of changes does your code introduce to Lingui?)
[//]: # (_Put an `x` in the boxes that apply_)

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

[//]: # (_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._)

- [x] I have read the [CONTRIBUTING](https://github.com/lingui/js-lingui/blob/main/CONTRIBUTING.md) and [CODE_OF_CONDUCT](https://github.com/lingui/js-lingui/blob/main/CODE_OF_CONDUCT.md) docs
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added the necessary documentation (if appropriate)

PR #2621, opened 2026-07-18
fix(cli): preserve bracketed names in catalog globs

### LINKED ISSUE #2183: When using Next.js dynamic routing, it is not possible to extract text into Po files.
### Description

In my Next.js project, I used dynamic routing [...params], and I customized the output directory for Lingui, finally I discovered that [...params] appeared in the directory. However, there was no content inside; it should have contained a "hello world" message.

The official demo can also reproduce it. 👉[repo](https://github.com/lingui/js-lingui/tree/main/examples/nextjs-babel)

### Verifications

### Reproduction Steps

**lingui.config.js**
Use {name} to custom the output
```javascript
const nextConfig = require("./next.config")

/** @type {import('@lingui/conf').LinguiConfig} */
module.exports = {
  locales: nextConfig.i18n.locales,
  pseudoLocale: "pseudo",
  sourceLocale: nextConfig.i18n.defaultLocale,
  fallbackLocales: {
    default: "en",
  },
  catalogs: [
    {
      path: "src/locales/{name}/{locale}/messages",
      include: ["src/pages/{name}"],
    },
  ],
}

```

 **pages/[...params]/index.tsx**
add 'Hello world'
```jsx
import { Trans } from "@lingui/react/macro"

export default function Home() {
  return <Trans>Hello World</Trans>
}

```

**locales/[...params]/cs/message.po**
But 'hello world' did not appear 
```po
msgid ""
msgstr ""
"POT-Creation-Date: 2025-02-13 20:31+0800\n"
"MIME-Version: 1.0\n"
"Content-Type: text/plain; charset=utf-8\n"
"Content-Transfer-Encoding: 8bit\n"
"X-Generator: @lingui/cli\n"
"Language: cs\n"

```

### Expected Behavior

"hello word" should appear in the po file

### Macro Support

Babel with babel-macro-plugin

### Lingui Version

5.0.0-next.2

### Babel Version

5.0.0-next.2

### Framework

Next.js

ahmojo:
# Description

Fixes #2183.

## Reproduction

Given this `lingui.config.js`:

```js
module.exports = {
  locales: ["en"],
  catalogs: [
    {
      path: "src/locales/{name}/{locale}/messages",
      include: ["src/pages/{name}"],
    },
  ],
}
```

and this file structure:

```text
src/
  pages/
    [...params]/
      index.tsx
```

where `index.tsx` contains:

```tsx
import { Trans } from "@lingui/react/macro"

export default function Page() {
  return <Trans>Hello World</Trans>
}
```

run:

```sh
yarn lingui extract
```

Before this change, Lingui creates:

```text
src/locales/[...params]/en/messages.po
```

but the catalog does not contain `Hello World`. The expected behavior is for the message from `src/pages/[...params]/index.tsx` to be extracted into that catalog.

## Root cause

Named catalog discovery replaces `{name}` with `*`, so it correctly discovers the directory `src/pages/[...params]`.

The discovered name is then substituted back into the configured source path:

```text
src/pages/[...params]
```

This value is later passed to Node's glob implementation. In a glob, `[...]` is interpreted as a character class instead of literal path data. The resulting pattern therefore does not match the actual `[...params]` directory, and its source files are omitted from extraction.

## Solution

This change keeps two representations of named catalog paths:

- Literal values for catalog paths and the public `include` and `exclude` properties.
- Glob-safe values for matching source files.

Only bracket and parenthesis characters originating from the `{name}` replacement are escaped:

```text
[...params] -> [[]...params[]]
(group) -> [(]group[)]
```

Configured glob syntax in `include` and `exclude`, including patterns such as `**/*.{ts,tsx}`, remains unchanged. The glob-safe patterns are used by both regular extraction and watch mode.

This fix intentionally targets bracket and parenthesis characters used by Next.js dynamic route and route group directory names. It does not introduce general escaping for every possible glob metacharacter.

## Validation

- `yarn vitest --run packages/cli/src/api/catalog/getCatalogs.test.ts packages/cli/src/api/getPathsForExtractWatcher.test.ts packages/cli/src/api/catalog.test.ts` on Node 22.19 and Node 24
- `yarn exec tsc --noEmit -p packages/cli/tsconfig.json`
- Targeted ESLint and Prettier checks
- `git diff --check`

Regression tests cover ordinary named directories, `(group)`, `[slug]`, `[...params]`, `[[...params]]`, exclusion patterns, watch-mode matching, and updated public `include` and `exclude` values.

No new dependencies were added.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] Documentation update
- [ ] Examples update

## Checklist

### DIFF
--- a/packages/cli/src/api/catalog/getCatalogs.test.ts
+++ b/packages/cli/src/api/catalog/getCatalogs.test.ts
@@ -6,10 +6,10 @@ import {
 } from "./getCatalogs.js"
 import { Catalog } from "../catalog.js"
 import { LinguiConfig, makeConfig } from "@lingui/conf"
-import fs from "fs"
-import os from "os"
 import path from "path"
 import { FormatterWrapper, getFormat } from "../formats/index.js"
+import { createFixtures } from "../../tests.js"
+import normalizePath from "normalize-path"
 
 function mockConfig(config: Partial<LinguiConfig> = {}) {
   return makeConfig(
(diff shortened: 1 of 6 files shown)


timofei-iatsenko:
Please describe what was the issue, what is steps to reproduce (configuration fro catalogs, file structure, etc) and how this is solving it. From the PR description it is not clear.

ahmojo:
> Please describe what was the issue, what is steps to reproduce (configuration fro catalogs, file structure, etc) and how this is solving it. From the PR description it is not clear.

Thanks, I updated the PR description with the reproduction steps and a clearer explanation of the fix.

packages/cli/src/api/catalog/getCatalogs.test.ts
```diff
@@ -133,6 +135,71 @@ describe("getCatalogs", () => {
...
+        )
+
+        const messages = await catalog!.collect()
+        expect(Object.values(messages!)).toEqual([
+          expect.objectContaining({ message: "Hello World" }),
+        ])
+      } finally {
+        process.chdir(previousCwd)
+        fs.rmSync(fixtureRoot, { recursive: true, force: true })
+      }
```
timofei-iatsenko: Please use `createFixtures` helper for this and other added tests:

```ts
    const rootDir = await createFixtures({
        "/components/foo.tsx": "",
        "/pages/bar.tsx": "",
        "/en/foo.tsx.po": `
msgid "Foo Hello World"
msgstr "Foo Hello World"
        `,
        "/en/bar.tsx.po": `
msgid "Bar Hello World"
msgstr "Bar Hello World"
        `,
        "/pl/foo.tsx.po": `
msgid "Foo Hello World"
msgstr "[PL] Foo Hello World"
        `,
        "/pl/bar.tsx.po": `
msgid "Bar Hello World"
msgstr "[PL] Bar Hello World"
        `,
      })

      const config = getConfig(rootDir)
```

You also don't need to change the working dir in tests, the implementation should not rely on working dir.

timofei-iatsenko (changes requested):

packages/cli/src/api/catalog/getCatalogs.test.ts
ahmojo: Thanks, I updated the added tests to use createFixtures and removed the working directory changes.

ahmojo:
The Windows Node 22 check timed out in an unrelated existing compile test. The tests changed in this PR passed in the same job. Could you please rerun the failed job?

andrii-bodnar (approved):
@ahmojo thank you for the contribution!

We had a similar one in the past (#2242) which was later reverted in #2396. But the current approach is narrower.

One thing: you only escape [/] - do Next.js route groups like (group) need the same handling, or does minimatch leave bare parens alone? Worth a test to lock it in. LGTM otherwise.

ahmojo:
> @ahmojo thank you for the contribution!
> 
> We had a similar one in the past (#2242) which was later reverted in #2396. But the current approach is narrower.
> 
> One thing: you only escape [/] - do Next.js route groups like (group) need the same handling, or does minimatch leave bare parens alone? Worth a test to lock it in. LGTM otherwise.

Good catch. Next.js route groups also need escaping in watch mode. I updated the fix and added tests for `(group)`.

andrii-bodnar (approved):

PR #2622, opened 2026-07-20
fix(babel-plugin-lingui-macro): unwrap non-null assertion (`x!`) / `satisfies` in placeholders to match SWC plugin

pkasarda:
## Description

`tokenizeExpression` in the babel macro only unwraps `TSAsExpression`, so a placeholder that wraps its expression in a **non-null assertion** (`x!`) or `satisfies` is treated as a *complex* expression and assigned an **indexed** placeholder (`{0}`). `@lingui/swc-plugin`, however, unwraps those TS-only wrappers and produces a **named** placeholder (`{x}`).

Because the message id is derived from the normalized message, the two toolchains generate **different ids for identical source**, which breaks catalog lookups in projects that **extract with the JS macro/CLI but run with the SWC plugin** — the runtime asks for an id the compiled (babel-extracted) catalog doesn't contain, so the raw id renders. This is the same class of JS↔SWC divergence addressed for numeric placeholders in lingui/swc-plugin#239.

## Reproduction

```ts
t`Value: ${x!}`
```

| toolchain | message | placeholder |
|---|---|---|
| `@lingui/babel-plugin-lingui-macro` (extract) — **before** | `Value: {0}` | indexed |
| `@lingui/swc-plugin` (runtime) | `Value: {x}` | named |

`${x as T}` already agrees (`{x}` on both) because babel unwraps `as`; only `!` (and `satisfies`) diverged.

## Fix

Unwrap `TSNonNullExpression` and `TSSatisfiesExpression` alongside `TSAsExpression` in `tokenizeExpression`, so all TS-only expression wrappers let the inner expression drive placeholder naming — matching the SWC plugin and the existing `as` behavior. After the fix, `t`Value: ${x!}`` extracts as `Value: {x}` on both toolchains.

## Notes

- Adds a regression test in `js-t.test.ts` (`useTypescriptPreset`).
- This changes the generated id for existing `${x!}` usages (`{0}` → named). That's the point (aligning with the SWC plugin), but worth a note in release notes for anyone relying on the old indexed id.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/babel-plugin-lingui-macro/src/macroJsAst.ts
+++ b/packages/babel-plugin-lingui-macro/src/macroJsAst.ts
@@ -262,7 +262,13 @@ export function tokenizeExpression(
   node: Node | Expression,
   ctx: MacroJsContext,
 ): ArgToken {
-  if (t.isTSAsExpression(node)) {
+  if (
+    t.isTSAsExpression(node) ||
+    t.isTSNonNullExpression(node) ||
+    t.isTSSatisfiesExpression(node)
+  ) {
+    // Unwrap TS-only expression wrappers (`x as T`, `x!`, `x satisfies T`) so the
+    // inner expression drives placeholder naming (e.g. `${x!}` → `{x}`, not `{0}`).
     return tokenizeExpression(node.expression, ctx)
   }
   if (t.isObjectExpression(node)) {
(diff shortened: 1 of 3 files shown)


timofei-iatsenko (approved):

timofei-iatsenko:
@pkasarda could you add same exact test to the swc plugin repo?

pkasarda:
done → lingui/swc-plugin#247

heads up: ended up being fix + test, not just a test — the plugin had the same gap (only `TsAs` was unwrapped). in practice @swc/core strips the `!` before the plugin runs so real pipelines already emit `{x}`, the fix just makes the plugin itself consistent + lets the test assert `{name}` instead of `{0}`.

PR #2623, opened 2026-07-21
feat(babel-plugin): support Babel 8 via peer dependencies

### LINKED ISSUE #2592: Support for Babel 8
### Problem Description

As [Babel 8 has been released](https://babeljs.io/blog/2026/06/16/8.0.0/), it would be great if the macro plugin and the CLI could be updated to support Babel 8 to avoid installing a mix of 2 Babel versions in projects.

### Proposed Solution

Upgrade lingui to use babel 8

### Alternatives Considered

Alternative would be to use peer dependencies to let the project choose between babel 7 and 8, but it might be hard to support both versions in the same codebase.

### Additional Context

Babel 7 is now receiving only security fixes. It won't get any new features or bug fixes.

youdie006:
## What

Adds Babel 8 support to the macro/babel plugins via peer dependencies, per #2592 and the direction @timofei-iatsenko and @andrii-bodnar set (support both Babel 7 and 8 via peer deps, not a hard upgrade).

## Changes

- **`@lingui/babel-plugin-lingui-macro`**: move `@babel/core` / `@babel/types` from direct `dependencies` to optional `peerDependencies` (`^7.20.12 || ^8.0.0` / `^7.20.7 || ^8.0.0`), mirroring `@lingui/vite-plugin`'s existing peer pattern. Added matching devDeps so the package still builds and tests locally.
- **`@lingui/vite-plugin`**: extend the existing peer from `^8.0.0-rc.1` to stable `^8.0.0`.
- **`@lingui/cli`** (extractor): unchanged — keeps its fixed Babel 7 direct deps, per the architectural line.

## Why peer, not direct-widening

I first tried simply widening the macro's direct deps to `^7 || ^8`. On the repo's own install the resolver then pulled **two** `@babel/core` majors (7.x for the CLI/others, 8.x for the macro) — exactly the "two Babel majors installed" outcome the issue wants to avoid, and it broke macro tests. Expressing them as peers keeps the repo (and consumers) on a single coherent major while advertising `^7 || ^8` downstream. The macro doesn't import `@babel/*` at consumer runtime (Babel is only touched inside a Babel transform), so the peers are optional — install completes with only the non-fatal optional-peer notices.

## Babel 8 fixes (small, mechanical)

- `messageDescriptorUtils.ts`: numeric placeholder keys (plural/select indices) were built as `types.identifier("0")`; Babel 8's `@babel/types` validates identifier names and rejects that, so numeric keys are now emitted as numeric literals (identical `{ 0: ... }` output on both majors).
- `babel-plugin-extract-messages` test: use `transformSync` (Babel 8 made `transform` callback-only; `transformSync` exists on both).

## Verification

Tests pass on **both** Babel 7 (shipping default) and Babel 8 (validated with a temporary root `resolutions` pin): the macro + extract-messages + cli suites are green, and `tsc` is clean on the changed packages.

Note: the CLI extractor intentionally stays on Babel 7 (per the architectural line above), so this PR does not migrate it. Moving the CLI itself onto Babel 8 would be a separate, larger change (a few type-only frictions around `@babel/core`/`@babel/parser` re-exports and Babel 8's tightened plugin/parser types).

Closes #2592.

---

*AI disclosure: this PR was drafted with Claude Code (AI-assisted). I verified it by running the macro + extract-messages suites on both Babel 7 and Babel 8 (green) plus `tsc` on the changed packages.*

### DIFF
--- a/packages/babel-plugin-extract-messages/test/extract-messages.test.ts
+++ b/packages/babel-plugin-extract-messages/test/extract-messages.test.ts
@@ -1,6 +1,6 @@
 import path from "path"
 import fs from "fs"
-import { transform as babelTransform } from "@babel/core"
+import { transformSync as babelTransform } from "@babel/core"
 import plugin, { ExtractedMessage, ExtractPluginOpts } from "../src/index"
 import { mockConsole } from "@lingui/test-utils"
 import linguiMacroPlugin, {
--- a/packages/babel-plugin-lingui-macro/package.json
+++ b/packages/babel-plugin-lingui-macro/package.json
@@ -55,16 +55,28 @@
     "node": ">=22.19.0"
   },
   "dependencies": {
-    "@babel/core": "^7.20.12",
-    "@babel/types": "^7.20.7",
     "@lingui/conf": "workspace:*",
     "@lingui/message-utils": "workspace:*"
   },
+  "peerDependencies": {
+    "@babel/core": "^7.20.12 || ^8.0.0",
+    "@babel/types": "^7.20.7 || ^8.0.0"
+  },
+  "peerDependenciesMeta": {
+    "@babel/core": {
+      "optional": true
+    },
+    "@babel/types": {
+      "optional": true
+    }
+  },
   "devDependencies": {
+    "@babel/core": "^7.29.0",
     "@babel/parser": "^7.29.2",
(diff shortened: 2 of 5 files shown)


timofei-iatsenko (approved):

andrii-bodnar (approved):
@youdie006 thank you!

PR #2644, opened 2026-08-06
feat(cli): move the extractor toolchain to Babel 8

kristoferma:
# Pull Request

## Description

Follow-up to #2623, which added Babel 8 support to the macro plugin via peer dependencies but deliberately left `@lingui/cli` on fixed Babel 7 direct dependencies — noted there as "a separate, larger change".

That leaves the original goal of #2592 unmet for consumers: because the CLI pins `@babel/core: ^7.21.0` as a real dependency, a project on Babel 8 still ends up with **two Babel majors installed**. This PR moves the CLI's pinned Babel to 8 and makes the surrounding packages type-check against it, so the tree resolves to exactly one `@babel/core`.

This respects the architectural line from #2592 — Babel stays a *fixed direct dependency* of the CLI (an internal implementation detail), not a peer. Only the pinned major moves. The plugins' peer ranges are untouched and still advertise `^7 || ^8`.

One thing worth flagging: the macro plugin's `^7 || ^8` peer range shipped in 6.6.0 without ever being type-checked against Babel 8, since its devDependencies stayed on `^7`. Several of the fixes below are in that package — they are latent breakages the peer range already implies, surfaced now that the repo actually resolves Babel 8.

## Babel 8 changes handled

| Change | Fix |
|---|---|
| `@babel/core` no longer re-exports `ParserOptions` | Import from `@babel/parser`, next to the existing `ParserPlugin` import |
| `importAttributes` and `explicitResourceManagement` are enabled by default and rejected as parser plugin names | Removed from `getBabelParserOptions` |
| `t.jsxElement` dropped its 4th `selfClosing` argument | Removed at both call sites |
| `PluginObj` renamed to `PluginObject` | Plugin shape declared structurally from `Visitor`/`PluginPass`, which both majors export |
| `PluginTarget` / `PresetTarget` no longer admit bare objects | Two inline test plugins become factories; the rolldown preset is cast, as Babel still accepts an object preset at runtime |
| `@babel/traverse` stopped exporting `VisitNodeObject` | Declared locally |
| `transformAsync`'s source map has readonly array fields, which rolldown's `ExistingRawSourceMap` rejects | Copied before returning |
| `file.hub` is `HubInterface`, not the `Hub` class | Parameter types widened to `HubInterface` (exported by both majors) |

Where a Babel 8 type name would have locked the source to one major, I declared the shape locally instead, so the plugins keep type-checking under either — consistent with their peer ranges.

The parser-plugin removal is the one change that could regress silently, so I verified it rather than trusting the release notes. On `@babel/core` 8.0.1, both syntaxes parse **without** the plugins listed:

```js
parseSync(`import x from "./a.json" with { type: "json" };`)   // OK
parseSync(`async function f(){ await using r = getR(); }`)      // OK
```

`decoratorAutoAccessors` and `deferredImportEvaluation` are still required and are kept.

## Verification

On this branch, with a single `@babel/core@8.0.1` resolved across the workspace (confirmed — no nested copies remain):

- `yarn release:build` — clean
- `yarn lint:types` — clean
- `yarn lint:eslint` — clean
- `yarn test` — **1414 passed, 2 skipped, 88 files**
- `yarn prettier:check` — clean

One note on running the suite locally: `extract-messages`' `expectNoConsole` helper fails on a fresh clone because Browserslist warns that its bundled `caniuse-lite` data is stale, and that warning trips `expect(console.warn).not.toBeCalled()`. It is unrelated to Babel; `BROWSERSLIST_IGNORE_OLD_DATA=1` (or refreshing the DB) makes the suite green. Happy to send that as a separate hardening PR if it is worth insulating the helper from Browserslist.

`packages/solid` and `packages/vite-plugin` had exact `@babel/core: 7.29.7` devDependencies; those move to `8.0.1` too, otherwise the workspace keeps three Babel copies and type resolution mixes majors between the hoisted and nested ones.

## Scope note

I kept this to the dependency move plus the minimum needed to make it type-check and pass. If you would rather see the plugin-side type fixes split out from the CLI dependency bump, I am happy to break it into two PRs.

Refs #2592, #2623.

---

*AI disclosure: this PR was drafted with Claude Code (AI-assisted). I verified it by running build, types, eslint, prettier and the full test suite on a single-major Babel 8 install, and by parse-testing the removed parser plugins against @babel/core 8.0.1.*

### DIFF
--- a/packages/babel-plugin-extract-messages/package.json
+++ b/packages/babel-plugin-extract-messages/package.json
@@ -46,9 +46,9 @@
     "@lingui/conf": "workspace:*"
   },
   "devDependencies": {
-    "@babel/core": "^7.29.0",
-    "@babel/traverse": "^7.29.0",
-    "@babel/types": "^7.29.0",
+    "@babel/core": "^8.0.0",
+    "@babel/traverse": "^8.0.0",
+    "@babel/types": "^8.0.0",
     "@lingui/babel-plugin-lingui-macro": "workspace:*",
     "@lingui/test-utils": "workspace:*",
     "typescript": "catalog:",
--- a/packages/babel-plugin-extract-messages/src/index.ts
+++ b/packages/babel-plugin-extract-messages/src/index.ts
@@ -8,14 +8,24 @@ import type {
   ObjectProperty,
   // eslint-disable-next-line import/no-duplicates
 } from "@babel/types"
-import type { PluginObj, PluginPass, NodePath } from "@babel/core"
-import type { Hub } from "@babel/traverse"
+import type { PluginPass, NodePath, Visitor } from "@babel/core"
+import type { HubInterface } from "@babel/traverse"
 import {
   getConfig as loadConfig,
   type LinguiConfigNormalized,
 } from "@lingui/conf"
 
 type BabelTypes = typeof BabelTypesNamespace
+
+/**
+ * `@babel/core` exports this shape as `PluginObj` in Babel 7 and `PluginObject`
+ * in Babel 8. Declaring it structurally keeps the plugin type-checking against
+ * either major, matching the `^7 || ^8` peer range.
+ */
+type PluginObj = {
+  name?: string
+  visitor: Visitor<PluginPass>
+}
 
 export type ExtractedMessage = {
   id: string
@@ -71,7 +81,7 @@ function collectMessage(
 function getTextFromExpression(
   t: BabelTypes,
   exp: Expression,
-  hub: Hub,
+  hub: HubInterface,
   emitErrorOnVariable = true,
 ): string {
   if (t.isStringLiteral(exp)) {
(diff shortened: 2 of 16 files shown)


andrii-bodnar:
@kristoferma thanks for the contribution! Could you please take a look at the failed pipeline?

andrii-bodnar:
Thanks for working on this! The Babel 8 fixes look good, but I'd suggest narrowing the scope.

Since #2634 landed, `@babel/types` is a hard dependency of the macro plugin again, so after a rebase the CLI would run Babel 8 with a plugin pulling in types 7. The proper fix is for the plugin to use `types` from the Babel plugin API instead of importing `@babel/types` directly, so it always follows the host's Babel version. With that in place, `@babel/types` can be dropped from the plugin's dependencies, and a Babel 8 CI job next to the existing Babel 7 one would cover the peer range.

Moving the CLI and the other packages to Babel 8 has Node engine and parser implications for users, so let's keep that for the next major. The CLI-side changes from this branch will carry over.

PR #2649, opened 2026-08-22
docs(conf): clarify JSX placeholder default collisions

### LINKED ISSUE #2640: Docs - Clarify limitations and gotchas for jsxPlaceholderDefaults
### Page

https://lingui.dev/ref/conf#macrojsxplaceholderdefaults

### Type of Issue

Missing information

### Description

Context: https://github.com/lingui/js-lingui/issues/2631#issuecomment-5131146335

The `jsxPlaceholderDefaults` configuration can define a placeholder for a tag so that `<a>` becomes `<link>` instead of `<0>` for example. This is can be interpreted as multiple `<a>` generating `<link1>...<linkN>` etc.  However, this is not the case and instead we get an error stating there is a placeholder collision because 2 tags tried to take the name `<link>`.

### Suggested Improvement

Although I would wish the tags incrementally use a counter for `<link1>`...`<linkN>`. The library owners preferred not to. Meaning that if I want to have incremental placeholders or whatever, then I should manually tag them.

Fair. I think that's valid, but I think it should be explicitly documented as such rather than being only found out during extraction 😅 

I can open a PR, likely later this week for this.

### Contribution

chm-labs:
Closes #2640

## Summary

- document that `jsxPlaceholderDefaults` does not generate numbered names for repeated tags
- explain that duplicate default names cause a placeholder collision
- show the supported fix: unique explicit names via `jsxPlaceholderAttribute`

## Validation

- `yarn lint` (website)
- `yarn checkFormat` (website)
- `yarn build` (website)

### DIFF
--- a/website/docs/ref/conf.md
+++ b/website/docs/ref/conf.md
@@ -664,6 +664,24 @@ A mapping of JSX element tag names to default placeholder names. When a JSX elem
 
 Explicit attributes (via `jsxPlaceholderAttribute`) take priority over defaults.
 
+Default placeholder names are not automatically numbered. If more than one JSX
+element resolves to the same default name, extraction reports a placeholder
+collision. Give each element a unique explicit name instead:
+
+```jsx
+<Trans>
+  Read the{" "}
+  <a _t="termsLink" href="/terms">
+    terms
+  </a>{" "}
+  and
+  <a _t="privacyLink" href="/privacy">
+    privacy policy
+  </a>
+  .
+</Trans>
+```
+
 ## macro.idPrefixLeader
 
 Default value: `undefined`


website/docs/ref/conf.md
```diff
@@ -664,6 +664,24 @@ A mapping of JSX element tag names to default placeholder names. When a JSX elem
 
 Explicit attributes (via `jsxPlaceholderAttribute`) take priority over defaults.
 
+Default placeholder names are not automatically numbered. If more than one JSX
+element resolves to the same default name, extraction reports a placeholder
+collision. Give each element a unique explicit name instead:
```
andrii-bodnar, line 669:
```suggestion
Default placeholder names are not automatically numbered. If multiple distinct JSX elements (different tag or attributes) resolve to the same placeholder name, the macro throws an error. Identical elements are deduplicated and can share a name. Give distinct elements unique explicit names instead:
```

website/docs/ref/conf.md
```diff
@@ -664,6 +664,24 @@ A mapping of JSX element tag names to default placeholder names. When a JSX elem
...
+
+```jsx
+<Trans>
+  Read the{" "}
+  <a _t="termsLink" href="/terms">
+    terms
+  </a>{" "}
+  and
+  <a _t="privacyLink" href="/privacy">
```
andrii-bodnar, line 678: Missing space - this will render as "andprivacy policy"

andrii-bodnar (changes requested):

PR #2650, opened 2026-08-22
 docs: clarify jsxPlaceholderDefaults collision behavior

### LINKED ISSUE #2640: Docs - Clarify limitations and gotchas for jsxPlaceholderDefaults

Jyatin:
## Description

Clarifies the documented behavior of `macro.jsxPlaceholderDefaults`.

Default placeholder names are used as-is and are not automatically made unique. Reusing the same default placeholder name can result in a placeholder collision, rather than automatically generating names such as `<link1>`, `<link2>`, etc.

Users who need distinct placeholder names should assign them explicitly with `jsxPlaceholderAttribute`.

Fixes #2640

## Types of changes

- [x] Documentation update

## Checklist

- [x] I have read the CONTRIBUTING and CODE_OF_CONDUCT docs

### DIFF
--- a/website/docs/ref/conf.md
+++ b/website/docs/ref/conf.md
@@ -663,6 +663,11 @@ A mapping of JSX element tag names to default placeholder names. When a JSX elem
 ```
 
 Explicit attributes (via `jsxPlaceholderAttribute`) take priority over defaults.
+> **Important:** Default placeholder names are used as-is and are not automatically made unique.
+>
+> If multiple JSX elements use the same default placeholder name, Lingui reports a placeholder collision instead of generating unique names such as `<link1>`, `<link2>`, etc.
+>
+> To use different placeholder names for multiple elements, assign them explicitly with `jsxPlaceholderAttribute`.
 
 ## macro.idPrefixLeader
 


andrii-bodnar:
@Jyatin thanks for the contribution! Closing this PR since we already have #2649

PR #2651, opened 2026-08-31
feat(conf): support pseudolocale extend character

### LINKED ISSUE #2642: Docs - Missing Pseudolocalization options
### Page

https://lingui.dev/ref/conf#pseudolocale

### Type of Issue

Missing information

### Description

I was looking through pseudolocales, and as I was testing the output with the `extend` option, and I was getting something akin to:

```
[[       Ĥèłłó       ]]
```

Turns out that `expand` simply attaches a prefix before and after the message. It does not use more characters... For example, I was expecting something like:

```
[[¶øœĤèèłłóïøäß]]
```

But, I guess this is not how it works. It turns out there is a hidden option: `extendCharacter`

https://github.com/MartinCerny-awin/pseudolocale/blob/25f56961a216c873c167afaa7369632557229a73/src/str.ts#L5-L14

### Suggested Improvement

Document all options that are defined here but in the linguijs repo: https://github.com/MartinCerny-awin/pseudolocale/blob/25f56961a216c873c167afaa7369632557229a73/src/str.ts#L5-L14

Or define your own set of options, agnostic to the pseudolocalization library, and you can pass through through adhoc according to the pseudolocalization library needs if it supports it at all. This may even enable swapping pseudolocalization libraries.. But that's a bit too ambitious. For now I think documenting properly the options would be nice.

### Contribution

Soumoditya:
Fixes #2642.

Expose the underlying pseudolocale library's xtendCharacter option through Lingui's public configuration type. The option already passes through at runtime; this adds type support, reference documentation, and regression coverage for custom padding.

Validation: git diff --check. Full Yarn tests could not run locally because dependencies are not installed; CI will validate the TypeScript and documentation changes.

### DIFF
--- a/website/docs/ref/conf.md
+++ b/website/docs/ref/conf.md
@@ -221,13 +221,13 @@ It can be provided either as a string, or as an object to additionally configure
 The string form (`pseudoLocale: "pseudo"`) is deprecated and will be removed in a future major release. Use the object form (`pseudoLocale: { locale: "pseudo" }`) instead.
 :::
 
-| Option             | Type     | Default     | Description                                                                                         |
-| ------------------ | -------- | ----------- | --------------------------------------------------------------------------------------------------- |
-| `locale`           | `string` | —           | Locale used for pseudolocalization (required in the object form)                                    |
+| Option            | Type     | Default     | Description                                                                                         |
+| ----------------- | -------- | ----------- | --------------------------------------------------------------------------------------------------- |
+| `locale`          | `string` | —           | Locale used for pseudolocalization (required in the object form)                                    |
 | `prepend`         | `string` | `""`        | String prepended to the beginning of every pseudo-localized message                                 |
 | `append`          | `string` | `""`        | String appended to the end of every pseudo-localized message                                        |
 | `extend`          | `number` | `0`         | Extends the width of the string by the given percentage (e.g. `0.3` adds 30%)                       |
-| `extendCharacter` | `string` | `" "`       | Character used to pad messages when `extend` is set                                                  |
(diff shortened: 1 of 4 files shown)


website/docs/ref/conf.md
```diff
@@ -227,6 +227,7 @@ The string form (`pseudoLocale: "pseudo"`) is deprecated and will be removed in
 | `prepend`  | `string` | `""`        | String prepended to the beginning of every pseudo-localized message                                 |
 | `append`   | `string` | `""`        | String appended to the end of every pseudo-localized message                                        |
 | `extend`   | `number` | `0`         | Extends the width of the string by the given percentage (e.g. `0.3` adds 30%)                       |
+| `extendCharacter` | `string` | `" "` | Character used to pad messages when `extend` is set                                                  |
```
Goodwine: Nit: I think you should format this block

Goodwine (approved):
I'm not a maintainer, but I am implementing something similar as I just added support for RTL on the `pseudolocale` dependency, and this change looks good to me.

website/docs/ref/conf.md
Soumoditya: Formatted in 3bb3f13.

andrii-bodnar: @Soumoditya the CI/CD pipeline still fails on the docs-suite job. Please take a look

Soumoditya:
I addressed the documented table-formatting failure in the latest commit. The new checks are currently awaiting maintainer authorization to run workflows from my fork; once authorized, I will address any remaining CI result promptly.

andrii-bodnar (approved):
@Soumoditya thank you!

PR #2660, opened 2026-09-02
fix(cli): ignore writing to the locale file when the extractor fails

danielbyomujuni:
# Description
This fixes an Issue, we experienced that is the extractor crashed for one reason or another it would take out our translations along with it.

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)
- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)
  There is no issue, It was a bug we experienced using the tool, that we need fix, and wanted to upstream it

## Checklist
- [X] I have added the necessary documentation (if appropriate)

### DIFF
--- a/packages/cli/src/api/catalog.test.ts
+++ b/packages/cli/src/api/catalog.test.ts
@@ -9,6 +9,7 @@ import { createCompiledCatalog } from "./compile.js"
 
 import {
   copyFixture,
+  createFixtures,
   defaultMakeOptions,
   defaultMakeTemplateOptions,
   makeNextMessage,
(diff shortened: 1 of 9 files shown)


timofei-iatsenko:
@danielbyomujuni thanks for contribution. Could you specify what exactly crashing the extractor? Maybe we should start from here and eliminates crashes?

danielbyomujuni:
I haven't been able to pin point the exact cause of the crash, which is why my PR was about preventing damage.

What I currently think is the cause is how our build set up with elixir and js, cause lingui to crash randomly

PR #2665, opened 2026-09-04
fix: make orderBy handling consistent

### LINKED ISSUE #2662: `orderByMessageId` never got the #1808 fix, and custom `orderBy` functions throw
### Verifications

- [x] I've searched [existing issues](https://github.com/lingui/js-lingui/issues) and this hasn't been reported yet.
- [x] I've checked [the docs](https://lingui.dev) and this isn't expected behavior.
- [x] I'm using the latest version of Lingui, and the problem still occurs.

### Description

Hello!

I found two related gaps in the catalog sorting code, both still present on `main`.
1. `orderByMessage` was fixed to use a hardcoded collator in #1808 to stop catalogs from sorting differently depending on the machine's locale, but that fix was never applied to `orderByMessageId`.
2. #2394 added support for passing a custom `orderBy` function, but seems like `makeConfig.ts`'s `jest-validate` schema was never updated to allow it. `exampleConfig.orderBy` is still just `"message"` (a bare string), so `validate(config, configValidation)` rejects a function. 

The only way around these right now is by patching. Patches are small, so I am happy to open a PR if that lightens the load for the maintainers. 😄

### Reproduction Link

_No response_

### Reproduction Steps

```ts
// lingui.config.ts
import { defineConfig } from "@lingui/cli"
import { formatter } from "@lingui/format-json"

export default defineConfig({
  locales: ["en", "es"],
  sourceLocale: "en",
  format: formatter({ style: "minimal" }),
  orderBy: "messageId",
  catalogs: [{ path: "src/locales/{locale}/messages", include: ["src"] }],
})
```

- Run `lingui extract` with different locales. The same source produces catalogs with keys in different order.
- Try to work around it by pinning your own collator: change `orderBy` to `(a, b) => new Intl.Collator("en-US").compare(a.messageId, b.messageId)`. `lingui extract` throws:
```
Option "orderBy" must be of type:
  string
but instead received:
  function
```

### Expected Behavior

- `orderByMessageId` should behave the same way as `orderByMessage`
- A custom `orderBy` function should be accepted by config validation, matching what's already documented at https://lingui.dev/ref/conf#custom-function

### Macro Support

Babel with @lingui/babel-plugin-lingui-macro

### Lingui Version

6.6.0

### Babel / SWC Version

_No response_

### Node Version

22

### Framework

Vite 8

cqxswbc2:
## Summary
- reuse the existing `en-US` collator when ordering catalogs by message ID
- allow custom `orderBy` functions through configuration validation
- add focused regression tests for both behaviors

## Testing
- `yarn test` (1423 passed, 2 skipped)
- `yarn vitest --run packages/cli/src/api/catalog.test.ts`
- `yarn vitest --run packages/conf/src/index.test.ts`
- `yarn tsc --noEmit -p packages/cli/tsconfig.json`
- `yarn tsc --noEmit -p packages/conf/tsconfig.json`
- `yarn exec eslint packages/cli/src/api/catalog.ts packages/cli/src/api/catalog.test.ts packages/conf/src/makeConfig.ts packages/conf/src/index.test.ts --flag v10_config_lookup_from_file`
- `yarn exec prettier --check packages/cli/src/api/catalog.ts packages/cli/src/api/catalog.test.ts packages/conf/src/makeConfig.ts packages/conf/src/index.test.ts`

Closes #2662

### DIFF
--- a/packages/cli/src/api/catalog.ts
+++ b/packages/cli/src/api/catalog.ts
@@ -462,8 +462,12 @@ export function order<T extends CatalogType>(by: OrderBy, catalog: T): T {
  * Object keys are in the same order as they were created
  * https://stackoverflow.com/a/31102605/1535540
  */
+// hardcoded en-US locale to have consistent sorting
+// @see https://github.com/lingui/js-lingui/pull/1808
+const collator = new Intl.Collator("en-US")
+
 const orderByMessageId: OrderByFn = (a, b) => {
-  return a.messageId.localeCompare(b.messageId)
+  return collator.compare(a.messageId, b.messageId)
 }
 
 const orderByOrigin: OrderByFn = (a, b) => {
@@ -515,10 +519,6 @@ export async function writeCompiled(
   await writeFile(filename, compiledCatalog)
   return filename
 }
-
-// hardcoded en-US locale to have consistent sorting
-// @see https://github.com/lingui/js-lingui/pull/1808
-const collator = new Intl.Collator("en-US")
 
 export const orderByMessage: OrderByFn = (a, b) => {
   const aMsg = a.entry.message || ""
--- a/packages/cli/src/api/catalog.ts
+++ b/packages/cli/src/api/catalog.ts
@@ -458,14 +458,14 @@ export function order<T extends CatalogType>(by: OrderBy, catalog: T): T {
       return acc
     }, {} as T)
 }
-/**
- * Object keys are in the same order as they were created
- * https://stackoverflow.com/a/31102605/1535540
- */
 // hardcoded en-US locale to have consistent sorting
 // @see https://github.com/lingui/js-lingui/pull/1808
 const collator = new Intl.Collator("en-US")
 
+/**
+ * Object keys are in the same order as they were created
+ * https://stackoverflow.com/a/31102605/1535540
+ */
 const orderByMessageId: OrderByFn = (a, b) => {
   return collator.compare(a.messageId, b.messageId)
 }
--- a/packages/conf/src/index.test.ts
+++ b/packages/conf/src/index.test.ts
@@ -60,6 +60,13 @@ describe("@lingui/conf", () => {
     })
   })
 
+  it("should accept a custom `orderBy` function", () => {
+    const orderBy = () => 0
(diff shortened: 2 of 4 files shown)


packages/conf/src/index.test.ts
```diff
@@ -60,6 +60,13 @@ describe("@lingui/conf", () => {
     })
   })
 
+  it("should accept a custom `orderBy` function", () => {
+    const orderBy = () => 0
+    const config = makeConfig({ locales: ["en"], orderBy })
+
+    expect(config.orderBy).toBe(orderBy)
+  })
```
timofei-iatsenko, line 72: This should be wrapped in `mockConsole((console) => {` and validated that no console.warn or error was called. Check the previous tests in this suite. If i', not mistaken validation is not throwing errors it's just printing the message to console.

timofei-iatsenko (changes requested):

packages/cli/src/api/catalog.ts
```diff
@@ -462,8 +462,12 @@ export function order<T extends CatalogType>(by: OrderBy, catalog: T): T {
  * Object keys are in the same order as they were created
  * https://stackoverflow.com/a/31102605/1535540
  */
+// hardcoded en-US locale to have consistent sorting
+// @see https://github.com/lingui/js-lingui/pull/1808
+const collator = new Intl.Collator("en-US")
+
```
andrii-bodnar, line 464: seems like this placement breaks the relationship between the comment above and the function below

packages/conf/src/index.test.ts
cqxswbc2: Updated in ba6eef6f: the test now runs inside mockConsole and asserts that neither console.warn nor console.error is called.

packages/cli/src/api/catalog.ts
cqxswbc2: Updated in ba6eef6f: moved the collator declaration before the existing block comment so the comment remains directly attached to orderByMessageId.

andrii-bodnar (approved):
@cqxswbc2 thank you!

andrii-bodnar:
@timofei-iatsenko does this look good to you?

Issue #1801, opened 2023-11-15
Extract is not working with pnpm workspace with hoist disabled


MPeloquin:
**Describe the bug**
Using `pnpm` [workspaces](https://pnpm.io/workspaces) and [hoist](https://pnpm.io/npmrc#hoist) disabled, `lingui extract` on v4 breaks with `Cannot find module '@lingui/macro'`

**To Reproduce**
Have a pnpm workspace
Disable hoist in `.npmrc`
Have a lingui v4 project
Run `extract`

See repro: https://github.com/MPeloquin/LinguiV4Extractpnpmissue

**Expected behavior**
Should be able to run extract without enabling hoist

A workaround is to add `hoist-pattern[]=*@lingui/macro*` in `.npmrc`

**Additional context**
pnpm version: 8.7.4

- jsLingui version ``lingui --version`` 4.5.5
- Babel version ``npm list @babel/core`` 7.23.3
- Macro support:
 - [ ] I'm using SWC with `@lingui/swc-plugin`
 - [x] I'm using Babel with `babel-macro-plugin`
 - [ ] I'm not using macro
- Your Babel config (e.g. `.babelrc`) or framework you use (Create React App, NextJs, Vite) Vite

thekip:
Well, it seems there is nothing I can do about it. `@lingui/cli` doesn't directly depend on `@lingui/macro`, instead it depends on `babel-plugin-macros`. 
The `babel-plugin-macros` is designed to actually implicitly depends (require) on "macros" packages discovered in your source code, but obviously doesn't state this packages in its own `package.json`, so your package manager doesn't know how to handle them correctly. 

I think your solution is ok for this macro, and any other you might use in your code. 

For the future plans, i have been thinking for long time that it's better to switch from `babel-plugin-macros` to just simple babel plugin. Because `babel-plugin-macros` just becoming more and more "out" from modern ecosystem.

Issue #1571, opened 2023-03-29
useLingui hook issue when importing a React Component from one application into another


nagamohanm:
I am trying to export a few react components (**ComponentA**, **ComponentB**) that include text surrounded by `<Trans />`  from let's say **Project A**. In **Project B**, I am importing these components as 

```
import { ComponentA, Component B } from '@projectA'
```

What is happening is as long as the code with `<Trans />` surrounding the text is not loaded (say due to some conditions not satisfied), the remaining Component is getting displayed in the UI. But once the conditions are met and the text has to be loaded, I am getting the below error:

```
useLingui hook was used without I18nProvider.
    at useLingui (webpack-internal:///./node_modules/@lingui/react/build/cjs/index.js:11:13)
    at Trans (webpack-internal:///./node_modules/@lingui/react/build/cjs/index.js:187:7)
    at renderWithHooks (webpack-internal:///./node_modules/@hot-loader/react-dom/cjs/react-dom.development.js:14804:18)
    at mountIndeterminateComponent (webpack-internal:///./node_modules/@hot-loader/react-dom/cjs/react-dom.development.js:17483:13)
    at beginWork (webpack-internal:///./node_modules/@hot-loader/react-dom/cjs/react-dom.development.js:18597:16)
    at HTMLUnknownElement.callCallback (webpack-internal:///./node_modules/@hot-loader/react-dom/cjs/react-dom.development.js:189:14)
    at Object.invokeGuardedCallbackDev (webpack-internal:///./node_modules/@hot-loader/react-dom/cjs/react-dom.development.js:238:16)
    at invokeGuardedCallback (webpack-internal:///./node_modules/@hot-loader/react-dom/cjs/react-dom.development.js:293:31)
    at beginWork$1 (webpack-internal:///./node_modules/@hot-loader/react-dom/cjs/react-dom.development.js:23204:7)
    at performUnitOfWork (webpack-internal:///./node_modules/@hot-loader/react-dom/cjs/react-dom.development.js:22155:12)
```

Both **ProjectA** and **ProjectB** have the `<I18nProvider i18n={i18n} />` surrounding the root component in their respective root files.

**Expected behavior**

**Additional context**
This started happening once I upgraded the version from 3.12.0 to 3.15.0

- jsLingui version ``^3.15.0``
- Babel version ``^7.7.7``

.babelrc
```
{
  "presets": [
    "@babel/preset-env",
    "@babel/preset-react"
  ],
  "plugins": [
[...]

Issue #1500, opened 2023-03-10
I18nContext not available in child component in component library with webpack5


skumar5-bech:
**Describe the bug**

I18nContext not available in child component in component library with webpack5 and throwing below error :
**Uncaught Error: useLingui hook was used without I18nProvider.**

**To Reproduce**
Code sample is available at this path where this issue is replicable
https://github.com/skumar5-bech/lingui-webpack5
To run the code, follow READ ME.

The sample is created using yarn workspace and has following:

1) React Application (lingui-sample) is created with create-react-app template
2) A reusable component library (common) is created with webpack5

Application & Components Illustration is given below to understand the issue:

React App ( (lingui-sample)
App Component -- App.tsx
----&lt;i18nProvider i18n={i18n}&gt;
---------&lt;Text&gt;&lt;/Text&gt; Referenced from reusable component library (common) 
-----&lt;18nProvider /&gt;

Reusable component library (common)
Text Component -- text/index.tsx
when trying to access i18nContext in this component using below code snippet then this issue will occur.
 const { i18n } = useLingui();

**Expected behavior**
I18nContext should be available to child components if parent is wrapped with i18nProvider.

**Additional context**

Other context providers (ThemeProvider from styled-components and custom) are working with webpack5 and also implemented in code sample provided in reproducible steps.

andrii-bodnar:
Hi @skumar5-bech,

currently, we have a v4 pre-release that contains a lot of improvements, fixes, and other changes. Could you, please check if the issue is reproducible on the Lingui `v4.0.0-next.1`?

refs: [v4.0.0-next.0](https://github.com/lingui/js-lingui/releases/tag/v4.0.0-next.0), [v4.0.0-next.1](https://github.com/lingui/js-lingui/releases/tag/v4.0.0-next.1)

thekip:
@andrii-bodnar this will not help for the issue. @skumar5-bech but you still give a try because there are many improvements and it quite stable 

@skumar5-bech, the problem is that your application and library consume different versions of `@lingui` dependencies. Application consumes esm version and library consumes cjs. Hence, you have two copies of the lib in the bundle, and 
 therefore react's context doesn't work. 

That happens because you library is built with cjs modules and have externalized imports like this: 

```ts
/***/ "@lingui/react":
/*!********************************!*\
  !*** external "@lingui/react" ***!
  \********************************/
/***/ ((module) => {

module.exports = require("@lingui/react");

/***/ }),
``` 

[...]

Issue #1458, opened 2023-02-22
RFC: deps tree Extractor. Support MPA apps (NextJS) and monorepos. 


thekip:
## Overview

Current extraction process operating on glob patterns to discover files: 

<img src="https://user-images.githubusercontent.com/1586852/220613463-9d710a1e-2ee2-4af1-970e-b7b916be3263.jpg" width="600">

That works, however this lead to a problem with multipage (MPA) frameworks such as NextJS. 

Consider we have following nextjs project structure: 

```
pages
  - index.ts
    - index.message1
    - index.message2
  - about.ts
    - about.message1
    - about.message2
  - faq.ts
    - faq.message1
    - faq.message2
```

Using file discovering approach, we will end up with a catalog consisting from all messages from all pages: 

```
index.message1
index.message2
about.message1
about.message2
faq.message1
faq.message2
``` 

This quickly become a problem, because we have to load full catalog for each page / navigation, and we would load messages which are completely unused on that page. 

We also couldn't setup extractor to extract each page directory separately because usually you have some common components between them:

```
pages
  - index
    - index.message1
    - index.message2
  - about
    - about.message1
    - about.message2

components
   header
    - header.message1
    - header.message2
```

So ideally we want to split somehow catalogs by routes (or other multi-entries) but still benefit from using extractor and avoid manual work. 

There are different solution out there. For example, i18next propose of using `ns` for each message, and load catalogs by this `ns`. But this quickly become a headache because you need to maintain all this dedendency manually. 

Here is also another more advanced proposal https://github.com/lingui/js-lingui/issues/1384 which might effectively solve this issue, on even more granular level. But i don't see how it could be easily integrated in nextjs ecosystem. I think having few bundles and running few instances of nodejs application is not a solution. 

### A monorepo problem
Monorepo share similar problems with multipage frameworks, consider you have a following structure

<img src="https://user-images.githubusercontent.com/1586852/220619055-71e6f576-98bd-4c5f-9c69-aa6cc24e729c.jpg" width="600">

Each application defines own messages and each library defines own messages as well. To correctly extract this we should keep track of all libraries used in each application in extractor config. 

```js
// Application1
  catalogs: [{
    path: '<rootDir>/translates/{locale}',
    include: [
         '<rootDir>/src/**',
         '<rootDir>/../libs/lib1/src/**',
         '<rootDir>/../libs/lib2/src/**',
    ],
  }],

// Application2
  catalogs: [{
[...]

PR #2667, opened 2026-09-07
docs: rewrite the monorepo setup guide

### LINKED ISSUE #2537: Rewrite Monorepo docs
I have this in the back of my mind for a looong time, need to put it as a task and do some day, here are topics i would like to cover: 

- Without Publishable Packages (apps/app-a, apps/app-b, packages/button, paclages/dialog, etc)
  Possible strategis: 
  - One catalog per app
  - Catalog per package + merging
- With publishable Packages
  - Avoid having translations in packages if possible, pass copies as Props
  - If not possible:
    - Catalog per package + merging
    - One catalog per whole UI Components library

The thing is that "Monorepo" could be very diffrent from project to project. We need to brainstorm how to structure this information, so the user would find a suitable information without reading unrelated parts. 

So something like flow diagram, or wizard:

"Do you publish your packages?" - Yes/No -> go to paragraph 2 or 3

> all above is a raw dump of my thoughts, to be improved later

andrii-bodnar:
# Description

Replaces the three-bullet monorepo stub with a guide covering the setups people actually ask about: one catalog per app, catalog per package merged per app, dependency-based extraction (experimental), and published packages, plus runtime sharing, Babel/SWC/Turborepo notes and troubleshooting.

Source material: issues #810, #1157, #894, #1458, #1801, #1571, #1500; discussions #2091, #2275, #1857, #1147, #908; and the Discord threads on Turborepo/pnpm setup, Nx multi-app extraction, and Lingui in published libraries.

Every config snippet was run with the built CLI against a scratch monorepo (two apps, one shared package), including `extract`, `compile` with `catalogsMergePath`, and `extract-experimental` with `includeDeps`.

Found while testing, left for follow-ups: relative `exclude` globs (including the default) do not apply to included paths outside the cwd; `include` wildcards do not descend into symlinked `node_modules` packages; the config reference documents the default exclude as `**/node_modules/**` while the code uses `*/node_modules/*`. An example monorepo project is planned separately.

## Types of changes

- [x] Documentation update
- [ ] Examples update

Closes #2537.

## Checklist


timofei-iatsenko:
The doc saying right things, but i'm afraid it too concise and could produce additional questions. I think it might need more examples/snippets for each strategy and more explanation.

andrii-bodnar:
I was thinking about creating a few example projects to follow up on this topic.

andrii-bodnar:
I added code where there was none: the app's `package.json` scripts, an `activate.ts` with two `i18n.load` calls for the Vite/webpack case, and a `packages/ui` component using `t` from `useLingui()`. All three are checked against a scratch workspace. The longer walkthroughs will come as example projects in a follow-up, where readers can clone and run them.

andrii-bodnar:
I'd suggest merging as is. It's a big improvement over the current stub, and the remaining bits (a two-apps example, the example projects) can go in follow-up PRs as questions come in.

PR #2639, opened 2026-08-03
fix(cli): export types needed by TypeScript

matthieusieben:
# Description

Fixes the following TypeScript error:

> The inferred type of 'default' cannot be named without a reference to 'LinguiConfig' from '.pnpm/@lingui+conf@6.6.0/node_modules/@lingui/conf'. This is likely not portable. A type annotation is necessary.

Capture d’écran 2026-08-03 à 13 18 04

Capture d’écran 2026-08-03 à 13 20 32

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/lingui/js-lingui/blob/main/CONTRIBUTING.md) and [CODE_OF_CONDUCT](https://github.com/lingui/js-lingui/blob/main/CODE_OF_CONDUCT.md) docs
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added the necessary documentation (if appropriate)

PR #2668, opened 2026-09-08
fix(cli): export all types that are part of the public API

matthieusieben:
# Description

Fixes the following TypeScript error:

> The inferred type of 'default' cannot be named without a reference to 'LinguiConfig' from '.pnpm/@lingui+conf@6.6.0/node_modules/@lingui/conf'. This is likely not portable. A type annotation is necessary.

Capture d’écran 2026-08-03 à 13 18 04

Capture d’écran 2026-08-03 à 13 20 32

This is a *real missuse of TypeScript*. A module should always explicitly re-export every type that are part of its public API. https://github.com/lingui/js-lingui/pull/2639 was closed without giving the opportunity to discuss this properly.

Note that this error happens with package managers that don't flatten the dependencies, like `pnpm` (and unlike `yarn`).

## Types of changes

- [ ] Documentation update
- [ ] Examples update

Fixes # (issue)

## Checklist


matthieusieben:
Here is how to reproduce (install using `pnpm`):


```jsonc
// package.json

{
  "name": "app",
  "private": true,
  "type": "module",
  "devDependencies": {
    "@lingui/cli": "*",
    "typescript": "*"
  },
  "scripts": {
    "build": "tsc"
  }
}
```

```jsonc
// tsconfig.json

{
  "compilerOptions": {
    "noEmit": true,
    "declaration": true // <== causes the error despite "noEmit" being true
  },
  "files": ["lingui.config.ts"]
}
```

```ts
// lingui.config.ts

export default defineConfig({ locales: ['en'] })
```

andrii-bodnar (commented):
@matthieusieben thank you!

Makes sense to me. It's type-only and the docs point everyone at `defineConfig` from `@lingui/cli`, so the type behind it should be reachable from there too.

@timofei-iatsenko, I know you closed #2639 with the tsconfig workaround, but this feels like the cheaper fix on our side. What do you think?

matthieusieben:
fwiw @timofei-iatsenko was right, the `lingui.config.ts` file was covered by a tsconfig that had `declarations: true`. This causes a TS error despite `noEmit: true`. The proper fix was indeed to create a dedicated tsconfig file, but I believe that the change suggested here makes things smoother for users of `@lingui/cli` which may have weird tsconfig setup like I had.

timofei-iatsenko (approved):

PR #2677, opened 2026-09-18
fix(cli): add translation as the last key of a new message

### LINKED ISSUE #2671: "Translation" moved from start to end of json object in 2nd extract
### Verifications

- [x] I've searched [existing issues](https://github.com/lingui/js-lingui/issues) and this hasn't been reported yet.
- [x] I've checked [the docs](https://lingui.dev) and this isn't expected behavior.
- [x] I'm using the latest version of Lingui, and the problem still occurs.

### Description

Image

Image

Since the first extract adds the translation to the front then it gets moved on merge there are PRs that show a json diff of this line being moved. Just add it to the end from the start

### Reproduction Link

_No response_

### Reproduction Steps

1. Create a new string
2. Run extract and merge PR
3. Pull latest and add different strings
4. Notice the json diffs will include the "translation" line being moved from the top to the bottom

### Expected Behavior

"Translation" line is added to the bottom on every extract run

### Macro Support

Babel with @lingui/babel-plugin-lingui-macro

### Lingui Version

6.4.0

### Babel / SWC Version

_No response_

### Node Version

_No response_

### Framework

_No response_

askalf:
Fixes #2671

## Summary

- `mergeCatalog` builds a **new** message as `{ translation, ...extracted }` but rebuilds an
  **existing** one as `{ ...extracted, extra, translation }`. The two paths disagree on key order.
- The JSON formatter serializes with `JSON.stringify`, which preserves insertion order, so a
  message written by the first extract has `"translation"` first and the *same* message written by
  the second extract has it last — a diff that moves the line and changes nothing.
- Fix: build new messages with `translation` last, matching the merge path. One statement reordered.
- 4 new regression tests (2 assertions × 2 locale modes) pin both the key position and the
  run-1 ≡ run-2 byte equality. All 4 fail on base, all 11 pass with the fix.
- Reported as #2671 and confirmed by maintainer `andrii-bodnar`, who prescribed this exact fix.

```console
$ cd packages/cli && npx vitest --run src/api/catalog/mergeCatalog.test.ts --reporter=verbose
# --- BASE (origin/main @ 4cc6cddc) ---
 ✓ src/api/catalog/mergeCatalog.test.ts > mergeCatalog > should initialize language catalog 7ms
 ✓ src/api/catalog/mergeCatalog.test.ts > mergeCatalog > should prefill translations for source language 0ms
 ✓ src/api/catalog/mergeCatalog.test.ts > mergeCatalog > should merge translations from existing catalogs 1ms
 ✓ src/api/catalog/mergeCatalog.test.ts > mergeCatalog > should overwrite translations for source locale when overwrite: true 1ms
 ✓ src/api/catalog/mergeCatalog.test.ts > mergeCatalog > should mark obsolete messages 1ms
 ✓ src/api/catalog/mergeCatalog.test.ts > mergeCatalog > should not mark as obsolete when files option passed 0ms
 ✓ src/api/catalog/mergeCatalog.test.ts > mergeCatalog > should keep message extra from the previous catalog 14ms
 × src/api/catalog/mergeCatalog.test.ts > mergeCatalog > key order > should add translation as the last key of a new message for a translated locale 7ms
   → expected 'message' to be 'translation' // Object.is equality
 × src/api/catalog/mergeCatalog.test.ts > mergeCatalog > key order > should add translation as the last key of a new message for the source locale 1ms
   → expected 'message' to be 'translation' // Object.is equality
 × src/api/catalog/mergeCatalog.test.ts > mergeCatalog > key order > should keep the key order of a message stable between extract runs for a translated locale 8ms
   → expected '{\n  "custom.id": {\n    "origin": [\…' to deeply equal '{\n  "custom.id": {\n    "translation…'
 × src/api/catalog/mergeCatalog.test.ts > mergeCatalog > key order > should keep the key order of a message stable between extract runs for the source locale 3ms
   → expected '{\n  "custom.id": {\n    "origin": [\…' to deeply equal '{\n  "custom.id": {\n    "translation…'

 Test Files  1 failed (1)
      Tests  4 failed | 7 passed (11)
   Duration  1.29s (transform 466ms, setup 15ms, import 831ms, tests 50ms, environment 0ms)

# --- HEAD (this PR) ---
 ✓ src/api/catalog/mergeCatalog.test.ts > mergeCatalog > should initialize language catalog 7ms
 ✓ src/api/catalog/mergeCatalog.test.ts > mergeCatalog > should prefill translations for source language 1ms
 ✓ src/api/catalog/mergeCatalog.test.ts > mergeCatalog > should merge translations from existing catalogs 1ms
 ✓ src/api/catalog/mergeCatalog.test.ts > mergeCatalog > should overwrite translations for source locale when overwrite: true 1ms
 ✓ src/api/catalog/mergeCatalog.test.ts > mergeCatalog > should mark obsolete messages 1ms
 ✓ src/api/catalog/mergeCatalog.test.ts > mergeCatalog > should not mark as obsolete when files option passed 0ms
 ✓ src/api/catalog/mergeCatalog.test.ts > mergeCatalog > should keep message extra from the previous catalog 13ms
 ✓ src/api/catalog/mergeCatalog.test.ts > mergeCatalog > key order > should add translation as the last key of a new message for a translated locale 1ms
 ✓ src/api/catalog/mergeCatalog.test.ts > mergeCatalog > key order > should add translation as the last key of a new message for the source locale 0ms
 ✓ src/api/catalog/mergeCatalog.test.ts > mergeCatalog > key order > should keep the key order of a message stable between extract runs for a translated locale 1ms
 ✓ src/api/catalog/mergeCatalog.test.ts > mergeCatalog > key order > should keep the key order of a message stable between extract runs for the source locale 0ms

 Test Files  1 passed (1)
      Tests  11 passed (11)
   Duration  1.19s (transform 428ms, setup 13ms, import 828ms, tests 28ms, environment 0ms)
```

## Decisions

The `translation` property is moved after the spread so the new-message branch produces the same
key order as the merge branch twelve lines below. This is minimal and behaviour-preserving beyond
the order: `ExtractedMessageType` has no `translation` field, and `mergeCatalog.ts` is the only
place in `packages/cli/src/api` that assigns `translation` to an extracted message, so the spread
cannot overwrite the computed value — only its position moves.

Rejected alternatives:
- Sort keys in the JSON formatter's `serialize` — bigger blast radius (rewrites every existing
  catalog on the next extract), doesn't fix the underlying disagreement between the two merge
  branches, and wouldn't help other order-preserving formatters.
- Move `translation` to the front in the merge branch instead — also makes the two paths agree, but
  rewrites every message in every existing catalog rather than only the ones the previous extract
  added; the reporter and maintainer both asked for translation-last.
- Normalize key order in a post-processing pass in `Catalog.make` — a new traversal of every
  message to fix what is a two-token ordering bug in the constructor.

Not run: the full test suite (`yarn test`), scoped to the changed test file plus
`src/api/catalog.test.ts` and `test/index.test.ts` instead, per the PR's own testing constraints.

AI assistance: this bug was found and the fix and tests were drafted with AI tooling in my workflow; the tests and checks above were executed as pasted. I'm responsible for the change and will handle review feedback.


packages/cli/src/api/catalog/mergeCatalog.test.ts
```diff
@@ -188,4 +188,59 @@ describe("mergeCatalog", () => {
       }
     `)
   })
+
+  describe("key order", () => {
+    // key order is what a formatter writes out, so it is what shows up in the
+    // catalog diff, @see https://github.com/lingui/js-lingui/issues/2671
```
andrii-bodnar: Let's avoid mentioning issues in the comments. If necessary, they can be traced back, as the commit name will include the PR number.

askalf:
Done — the issue references are gone from the comments. The same citation appeared twice, so I removed both: the flagged one in `mergeCatalog.test.ts` and its twin in `mergeCatalog.ts`. Both comments retain their explanation of stable key ordering.

`packages/cli/src/api/catalog/mergeCatalog.test.ts`: 11/11 passing; Prettier check and `tsc --noEmit` in `packages/cli` are clean. The diff from the reviewed head is comment-only (two removed issue-reference lines); production behavior is unchanged.

packages/cli/src/api/catalog/mergeCatalog.test.ts
```diff
@@ -188,4 +188,59 @@ describe("mergeCatalog", () => {
...
+      Object.keys(JSON.parse(JSON.stringify(entry)))
+
+    const locales: [label: string, forSourceLocale: boolean][] = [
+      ["for a translated locale", false],
+      ["for the source locale", true],
+    ]
+
+    it.each(locales)(
+      "should add translation as the last key of a new message %s",
+      (_, forSourceLocale) => {
```
andrii-bodnar: `forSourceLocale` only changes the value of translation, never its position - both literals put it last unconditionally, so this axis doubles both tests without covering anything new. I'd drop it and keep one key-position test and one stability test.

askalf:
Agreed, `forSourceLocale` only picks the value of `translation`, while the literal places it last either way, so the locale axis added no coverage. I removed the `locales` table and both `it.each` wrappers; the key-order block now has one position test and one stability test, both with `forSourceLocale: false`. `mergeCatalog.ts` is untouched. `mergeCatalog.test.ts` is 9/9 passing; with `mergeCatalog.ts` at the merge-base, exactly the two key-order tests fail. Prettier, ESLint, and `tsc --noEmit` in `packages/cli` are clean.

andrii-bodnar (approved):
@askalf thank you!

